PR #4, opened 2019-01-02
sql scheme from UML in google drive

FilipePintoReis:
Está a faltar aí um check na table do Ad.
Pedia que revissem as linhas 20 e 60 do ficheiro sql


imnotteixeira:
@FilipePintoReis Please re-create this PR from `feat/database` -> `develop` instead of `master`. 

Also, on lines 20 and 60, remove the comments from the code, and add a line review for them on github instead, so that when merging there are no comments behind

PR #5, opened 2019-01-06
Feat/database

FilipePintoReis:
Wouldn't it be better if there was a telephone number and an email and one of them had to exist?
line 20

endDate - publishDate >= 1 and <= 6, there's a check missing
line 58

What format should this be in?
line 60


src/main/database/database.sql
```diff
@@ -0,0 +1,105 @@
...
+	companyId INTEGER REFERENCES Company(id),
+	timestamp	DATETIME,
+	extrainfo	TEXT
+);
+
+DROP TABLE IF EXISTS Company;
+CREATE TABLE Company (
+	id	 	INTEGER PRIMARY KEY AUTOINCREMENT,
+	name	TEXT,
+	contacts	TEXT NOT NULL,
```
FilipePintoReis: Wouldn't it be better if there was a telephone number and an email and one of them had to exist?

src/main/database/database.sql
```diff
@@ -0,0 +1,105 @@
...
+	technologyId3 INTEGER REFERENCES Technology(id),
+	technologyId4 INTEGER REFERENCES Technology(id),
+	technologyId5 INTEGER REFERENCES Technology(id),
+	technologyId6 INTEGER REFERENCES Technology(id),
+	technologyId7 INTEGER REFERENCES Technology(id),
+	cityId INTEGER REFERENCES City(id),
+	title	TEXT,
+	publishDate	DATETIME, 
+	endDate	DATETIME,
+	jobDuration INTEGER,
```
FilipePintoReis: What format should this be in?

miguelpduarte:
Heya!

Converting the database model into code is a good idea, but this is SQL!

We are going to be using MongoDB which is another database language and does not work with SQL :sweat_smile:

Issue #19, opened 2019-08-06
Improve code structure


miguelpduarte:
`Express` is a decent library/framework (who cares, really?), but it does not enforce any structure at all, which can result in getting stuff all over the place.

So, we should follow something like [this](https://softwareontheroad.com/ideal-nodejs-project-structure/) - I'm reading it now and it all makes a lot of sense. We can iterate on it, but I feel that at least a basic level of what's described there should be implemented :smile:

miguelpduarte:
Would like to have some insight/discussion on this, to get some opinions before enforcing too much stuff

imnotteixeira:
Didn't have the chance to dive deep into that solution, but I liked the general ideas in there, which promote code reliability and improve structure, making this project maintainable, which should be one of the goals. That being said, I approve.

xRuiAlves:
+1

miguelpduarte:
Cool beans, will bother you both when the PR eventually drops 😄 

(Probably should finish up #7 first, long time coming and almost done - it's on `feature/database`, should open a draft PR today and finish up soonTM - will need to fix my old code surely 🙃 )

PR #20, opened 2019-08-06
Feature/database

### LINKED ISSUE #7: Implement base MongoDB Schema

miguelpduarte:
Creating a PR already because draft PRs are a thing now, yay! 🎉 

Closes #7 finally, will try to finish this up soon TM


miguelpduarte:
Oh god, so many useless commits 😢 

This will need some aggressive interactive rebasing for squasherooo

imnotteixeira:
Can't we simply Squach and Merge this PR in the end? @miguelpduarte

miguelpduarte:
> Can't we simply Squach and Merge this PR in the end? @miguelpduarte

Yes but no, would rather just Merge without squashing to maintain a representative git history - actually want to enforce this as much as possible, and try to make people less afraid of git's cli and interactive rebasing, force pushing, etc

miguelpduarte:
Also something I noticed, our tests should not rely on the database running.
Should change the travis CI configs then, to ensure that we do not rely on this. It probably increases build time and encourages bad practices unnecessarily.

#19 will hopefully result in better and more easily testable code so it should also not be a problem moving forward 😄 

Will track it in an issue and add the link here.

imnotteixeira:
Well, they should if they are E2E Tests. For simple unit tests, no external service should be needed

miguelpduarte:
> 
> 
> Well, they should if they are E2E Tests. For simple unit tests, no external service should be needed

Not really... Depending on a database service running to test your code sounds like something I'd like to pointedly avoid... Dependency Injection and simply splitting the code correctly should make for testable code that does not require any DB (or external service, for that matter, if we end up using any) running.
It sounds simply like good practices to me.

imnotteixeira:
Real E2E tests should mimic the environment fully, and so they should have a DB service running. However, tests with mocked DBs should be enough in our case I believe

miguelpduarte:
> 
> 
> Real E2E tests should mimic the environment fully, and so they should have a DB service running. However, tests with mocked DBs should be enough in our case I believe

In what way does communicating with external anythings improve the tests? It just introduces an uncertainty factor relative to the communication itself and to the external service's uptime and behavior.

It is much better to always mock everything - in that way, as I've said, we are not testing any configuration or external service. What you mention makes sense, but only in the concept of having a `staging` environment, and not in the context of testing at all.

imnotteixeira:
Imagine the way you make the calls to the DB service makes it hang or abruptly fail die to a misconfiguration of some sort. That's when you want a real service to test, because the mocks would not fail in those cases.

As I've said, this is probably overengineering to have a full service running on our part, but to say that it is much better to always mock everything is not fully true.

miguelpduarte:
> Imagine the way you make the calls to the DB service makes it hang or abruptly fail die to a misconfiguration of some sort. That's when you want a real service to test, because the mocks would not fail in those cases.

I understand your point. However, the place for verifying that is not in tests that will run in the dev's PCs and in CI suites. That is a concern for staging/production environments configuration. And testing that in tests is entirely out of scope.

> As I've said, this is probably overengineering to have a full service running on our part, but to say that it is much better to always mock everything is not fully true.

That is my point exactly, it's overengineering and expensive one - we could use a node alpine image, which would certainly reduce the amount of time for CI to spin up drastically, and certainly the test suites in the dev's PCs as well.
I stand my point on mocking everything. Testing everything at once is being lazy imo, and just calling it E2E or integration testing is just an easy way out - my _personal_ opinion.
Feel free to disagree, but the point about CI spin up time stands, see what @xRuiAlves said in #22 for example.

(By the way, this discussion should totally be in #21 and not here 😅)

miguelpduarte:
I think this is good to go, undrafting

miguelpduarte:
Derp, need to rebase master into this
Will do it ASAP

miguelpduarte:
Well, that resulted in a lot of things not making sense anymore... :sweat_smile: 

Guess I'm gonna go for a merge to update upstream changes next time, despite the superfluous commit.

src/models/Ad.js
```diff
@@ -0,0 +1,117 @@
...
+const uniqueArrayPlugin = require("mongoose-unique-array");
+const JobTypes = require("./JobTypes");
+const {FieldTypes, MIN_FIELDS, MAX_FIELDS} = require("./FieldTypes");
+const {TechnologyTypes, MIN_TECHNOLOGIES, MAX_TECHNOLOGIES} = require("./TechnologyTypes");
+const PointSchema = require("./Point");
+
+// Defining relevant constants
+const {MONTH_IN_MS, AD_MAX_LIFETIME_MONTHS} = require("./TimeConstants");
+
+//TODO: Define indexes? (Location is indexed in Point subdocument supposedly already) Probably should index the search fields (all?)
```
imnotteixeira: Please do not place `TODO`s in code, create issues instead.

src/models/FieldTypes.js
```diff
@@ -0,0 +1,17 @@
+const FieldTypes = Object.freeze([
```
imnotteixeira: We should use a ENUM notation for the variables : `BACK_END` instead of `Back-End`, either here or the DB, assuming this will be there in the future, so that we have the freedom to work with translations without much changes to the code, as the variables will be the same, and only the display changes depending on language

src/models/JobTypes.js
```diff
@@ -0,0 +1,9 @@
+const JobTypes = Object.freeze([
```
imnotteixeira: Same comment as `FieldTypes`: ENUM notation.

src/models/Ad.js
miguelpduarte: Tracked here https://github.com/NIAEFEUP/nijobs-be/issues/27

miguelpduarte: (Probably a carry-over from the old development in which the communication was being done via commit comments and not PRs since there were no draft PRs 😅 )

src/models/FieldTypes.js
miguelpduarte: I understand the idea, but am not really sure if this is all that relevant: it's a string all the same and the front-end will have to parse things back and forth if this is changed to not be the display value.

But I see the point and will make no resistance to it (this would be more relevant especially if strings were to contain spaces)

miguelpduarte: What do you think? I can see this being annoying because of having to convert back and forth

miguelpduarte: Also, what about `TechnologyTypes`? Not considering translation there I guess?

imnotteixeira: `TechnologyTypes` is different, because they don't have translations. In this case we would store the enum values in the schema, and the frontend would fetch a translation document in the db for the selected value, hence the enum-style, so that this is decoupled from the display as much as possible. It is easier to do from the start rather than later.

miguelpduarte: Hmm I see 🤔 

However, I think that fetching the translation document sounds weird - we could just have the translation data in the Front-End using something like [LinguiJS](https://lingui.js.org/) or [react-intl](https://github.com/formatjs/react-intl).
What really matters is that the Front-End adheres to the Schema when making requests, and everything else (display related) should be handled by the Front-End imo

Anyway, this is a bit of a premature discussion anyway, as the time to worry about that is still a long time away haha

miguelpduarte:
Implemented the requested fixes (`s/.*/\U&` and `s/-/ _` made that a breeze, 😀)

Had some issues related to needing the DB to be running to be able to run the tests, but that should be addressed soon(TM) (#21) 😄

src/models/FieldTypes.js
imnotteixeira: Translations should not be hardcoded in the front-end, that makes changes harder, as to fix a simple typo in a translation, for example, one would need to create a PR, pass the tests, etc.. versus changing the DB record directly. But yeah, this is another discussion.

miguelpduarte: I don't really agree but yeah, that should be another discussion.

I'm going from what I've seen in more "corporate" environments, in which translation database files were coupled to the front-end, as they are display-related.

Furthermore, something that is displayed relying on a database means that changing it is not as easy, especially if an external database service is used. I believe that keeping this in source-code allows it to be version controlled and thus, more easily acted upon.

With NIAEFEUP/niployments#1 the deploy cycle should become faster and thus this would also be a reduced problem.

miguelpduarte:
🚢 'ing it!

Issue #21, opened 2019-08-06
Make tests not rely on having a running database


miguelpduarte:
Mostly explained here: https://github.com/NIAEFEUP/nijobs-be/pull/20#issuecomment-518697113

Needs changes to the TravisCI configs. Maybe also a good time to investigate other CI options, such as CircleCI (it's looking pretty competitive and the ease of specifying several jobs might be helpful, considering #18 too).

miguelpduarte:
Maybe investigate using something like this: https://github.com/nodkz/mongodb-memory-server

(Also found mockgoose, but that is deprecated and forwards people to this)

imnotteixeira:
Why use a separate service/server anyway? We just need to mock the calls with the data we want to receive

miguelpduarte:
For ease of implementation.

miguelpduarte:
Whenever one requires `mongoose` (which is being used everywhere) the db instance is used. It would require having to manually pass dependencies everywhere just in order to simplify testing.

imnotteixeira:
As far as I can see this package only provides the "mongo" service. You still have to populate the DB for the tests, clean it afterwards, etc... and you cannot easily test failures, for example. What I am suggesting is to mock the mongoose calls themselves (not changing any of our service implementation), and when the service calls mongoose for whatever, the test code runs, returning whatever mock data or resulting in an error depending on the test. This is pretty easy to do with stubs. Some examples can be found here (https://codeutopia.net/blog/2016/06/10/mongoose-models-and-unit-tests-the-definitive-guide/)

miguelpduarte:
This was the idea I was aiming for as well but did not notise at the time that the package did not work like that.

Ideally we would mock everything, yes :slightly_smiling_face:

PR #22, opened 2019-08-06
Add audit task to ci

### LINKED ISSUE #18: Add auditing to CI
I think it would be interesting to add package auditing to the continuous integration to avoid (at least to filter `high` vulnerabilities) to keep our packages clean-ish.

xRuiAlves:
Closes #18

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
     # Any separate test environment variables go here, one per line each creating a different test environment
 
 script:
-  - npm audit --audit-level=high &> /dev/null
+  - npm run audit
   - docker-compose up --exit-code-from test test
 
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ env:
     # Any separate test environment variables go here, one per line each creating a different test environment
 
 script:
+  - npm audit --audit-level=high &> /dev/null
   - docker-compose up --exit-code-from test test
 
 cache:
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,29 +14,59 @@
       }
     },
     "@babel/core": {
-      "version": "7.12.9",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.12.9.tgz",
-      "integrity": "sha512-gTXYh3M5wb7FRXQy+FErKFAv90BnlOuNn1QkCK2lREoPAjrQCO49+HVSrFoe5uakFAF5eenS75KbO2vQiLrTMQ==",
+      "version": "7.12.10",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.12.10.tgz",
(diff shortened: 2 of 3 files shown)


xRuiAlves:
I have no idea what's failing in this build and this travis-ci is confusing af. Input @miguelpduarte ?

xRuiAlves:
Also, how come the ci build takes 35 minutes to fail? Please let's change this to `circle-ci`, it's at least 35 time faster (since builds take at most 1 minute from my experience). Plus, it's better documented :x

miguelpduarte:
> I have no idea what's failing in this build and this travis-ci is confusing af. Input @miguelpduarte ?

A snippet from one of the build logs:

```
$ git clone --depth=50 https://github.com/NIAEFEUP/nijobs-be.git NIAEFEUP/nijobs-be
Cloning into 'NIAEFEUP/nijobs-be'...
fatal: unable to access 'https://github.com/NIAEFEUP/nijobs-be.git/': Failed to connect to github.com port 443: Connection timed out

The command "eval git clone --depth=50 https://github.com/NIAEFEUP/nijobs-be.git NIAEFEUP/nijobs-be " failed. Retrying, 2 of 3.
```

I think GitHub was just down 😂 😂 (So the delay to fail was much bigger than expected, no way that the small amount of code in `develop` currently would run for 35 minutes for free hahaha)
Restarting the build manually after some OAuth(?) Let's see if it fixes that

.travis.yml
```diff
@@ -16,7 +16,8 @@ env:
     # Any separate test environment variables go here, one per line each creating a different test environment
 
 script:
+  - npm audit --audit-level=high
```
miguelpduarte: Not sure if this works due to the way we're using docker for some things, but let's see if the tests pass.

imnotteixeira:
Managed to re-run some failing builds, but the missing one is actually failing because of one of the tests is failing. Track here: https://travis-ci.org/NIAEFEUP/nijobs-be/builds/568557192?utm_source=github_status

GuilhermeJSilva:
The error comes from an after hook that erases all accounts in the test database, not from the actual tests. Running this tests locally resulted in 0 errors, in both the develop branch and this branch, however the hook was also badly designed, this hook actually erases the accounts after each test, which is not necessary, I think changing it to make the action before and after that specific collection of tests is more advisable. Furthermore, looking at the error it seems that it resulted from a timeout in the connection to the database

miguelpduarte:
Now it seems that the tests are passing... Weird... :upside_down_face: This was probably due to some misconfiguration of docker stuff or of TravisCI (1 of the 4 checks is still failing, for some reason - shouldn't they all be the same? Travis is weird), I'm guessing.

Well, I'd say that this is just more motivation to tackle #19 and #21 :sweat_smile:

miguelpduarte:
@xRuiAlves can you merge the changes from `develop` to here to see if the checks all pass?

xRuiAlves:
Well, build was failing due to `stdout` "overflood" when running audit. Since we're only looking for the return status of the command, ignoring the text output did the trick

imnotteixeira:
The acorn vulnerability has since been solved. @xRuiAlves Can you remove the `&> /dev/null` so that the output is shown?

imnotteixeira (approved):

Issue #17, opened 2019-08-06
Add sample .env file


xRuiAlves:
I think it might be insteresting to add a semple env file in the repo's root (e.g. `.env.example`) instead of explaining how to setup the env file in the README.

We could even add a `.env` file and setup [dotenv-flow](https://www.npmjs.com/package/dotenv-flow) to allow overriding the env file with a local version (e.g. `.env.local`).
This change would allow the README setup instructions to become simpler aswell: just run `cp .env .env.local` and fill in the variables.

If this is done, we should also add a `.env*.local` entry to the `.gitignore`.

miguelpduarte:
As replied in the FE issue, this sounds pretty good to me, even reducing the documentation strain in the README (would be able to just refer to the .env file and maintain it as the single source of truth).

I'm for it, and dotenv-flow is a good idea, I have had good experience with it 😄

PR #23, opened 2019-08-06
Feature/dotenv flow

xRuiAlves:
Closes https://github.com/NIAEFEUP/nijobs-be/issues/17


miguelpduarte:
Please change your PR message to something like `Closes` instead of `Tackles` or `Tackling` - that way the issues are closed when the PR is merged. I edited one of the PR texts but will leave this comment here so that you know 😅

.env.example
```diff
@@ -0,0 +1,6 @@
+DB_PORT=
+DB_HOST=
```
miguelpduarte: `DB_HOST` does not exist and is not used, even in #20

.env.example
```diff
@@ -0,0 +1,6 @@
+DB_PORT=
```
miguelpduarte: `DB_PORT` does not exist and is not used, even in #20
We are always using the default port for the mongo service, iirc

I see that you added the functionality to override, but I really don't see the point. If we use an external service that might be using another port, we would be using `MONGO_URI` anyway.
Sincerely, this just makes me want to scrap `DB_HOST` too and just use `MONGO_URI` for everything, but would have to look at the docker stuff, because I think this is like this due to some configs there.

.env.example
```diff
@@ -0,0 +1,6 @@
...
+DB_HOST=
+HOST_PORT=
+MONGO_URI=
+NODE_ENV=
```
miguelpduarte: This is something that is absolutely irrelevant to have in a `.env` file, as it is set either by `node`, `npm` or some docker configs depending on the context. It is used as an environment variable directly and thus has no place in this file.

README.md
```diff
@@ -81,9 +81,10 @@ This environment doesn't have hot reloading or dev extensions and is made to be
 
 ### Env File Specification
 
-- `HOST_PORT`= The port that the server will expose (http://localhost:<HOST_PORT>)
-- `MONGO_URI`= [Optional] Specify a URI for an external mongo database
-- `DB_HOSTNAME`=  The hostname of the DB, in the case of not wanting to specify the full URI (defaults to `localhost`) - Useful for setting up docker containers, for example (the current docker configuration already considers this)
+- `DB_PORT` **[Optional]** Specifies the database port, if `MONGO_URI` is not defined. Defaults to `27017`.
+- `DB_HOSTNAME` **[Optional]** Specifies the database port, if `MONGO_URI` is not defined. Defaults to `localhost`.
+- `HOST_PORT` Specifies the port where the server will be exposed (`http://localhost:<HOST_PORT>`)
```
miguelpduarte:
```suggestion
- `HOST_PORT` Specifies the port in which the server will be exposed (`http://localhost:<HOST_PORT>`)
```

README.md
```diff
@@ -81,9 +81,10 @@ This environment doesn't have hot reloading or dev extensions and is made to be
 
 ### Env File Specification
 
-- `HOST_PORT`= The port that the server will expose (http://localhost:<HOST_PORT>)
-- `MONGO_URI`= [Optional] Specify a URI for an external mongo database
-- `DB_HOSTNAME`=  The hostname of the DB, in the case of not wanting to specify the full URI (defaults to `localhost`) - Useful for setting up docker containers, for example (the current docker configuration already considers this)
+- `DB_PORT` **[Optional]** Specifies the database port, if `MONGO_URI` is not defined. Defaults to `27017`.
```
miguelpduarte: As mentioned above, useless. Please remove

README.md
```diff
@@ -81,9 +81,10 @@ This environment doesn't have hot reloading or dev extensions and is made to be
...
+- `DB_HOSTNAME` **[Optional]** Specifies the database port, if `MONGO_URI` is not defined. Defaults to `localhost`.
```
miguelpduarte: Copy paste mistake.

```suggestion
- `DB_HOSTNAME` **[Optional]** Specifies the database hostname, if `MONGO_URI` is not defined. Defaults to `localhost`.
```

miguelpduarte:
Also, for some discussion: Should we have example values in `.env.example`?
Or even comments either as the values or near the variables?

Trying to figure out the best way going forward, don't want to have the upkeep of updating the README if the source of truth could be elsewhere.

.env.example
xRuiAlves: Agreed. That being said, i'm going with
```javascript
const MONGO_URI = process.env.MONGO_URI || "mongodb://localhost:27017";
```
removing DB_HOST and DB_PORT. `MONGO_URI` can be overwritten if needed.

xRuiAlves:
I'm going to scrap the env variables explanation from the README and adding a description in the `.env.example` file. It's a bit annoying to have the need to update both the example file and the readme whenever the env variable change. This way, there's only one source of truth, more easy to maintain.

.env.example
miguelpduarte: LGTM, this was probably a leftover from a rushed initial config, tbh, as other repos that use MongoDB only have the URI available as an env variable.

Just keep in mind that the docker config stuff might break, as it uses the env variables we are going to kill. However, some simple string interpolation should solve that, I hope. If you need help, feel free to ping me.

miguelpduarte:
Agreed @xRuiAlves, sounds perfect. Exactly my point in steering the discussion towards that.

However, I'd still keep a section in the README describing the copying of the `.env.example` file and refering to it for the documentation of the environment variables.

src/db_controller.js
```diff
@@ -1,10 +1,6 @@
 const mongoose = require("mongoose");
 
-// Default mongo port
-const DB_PORT = 27017;
-const DB_HOST = process.env.DB_HOSTNAME || "localhost";
-
-const MONGO_URI = process.env.MONGO_URI || `mongodb://${DB_HOST}:${DB_PORT}`;
+const MONGO_URI = process.env.MONGO_URI || "mongodb://localhost:27017";
```
miguelpduarte: Should we maybe not have a default at all and just blow up when there is none specified (I'd think trying to connect to an undefined URI probably will do the trick, but adding a simple condition here might also work).

.env.example
```diff
@@ -0,0 +1,3 @@
+HOST_PORT=8000                        # Specifies the port in which the app will be exposed (when ran using Docker)
+MONGO_URI=mongodb://localhost:27017   # Specifies the URI for an external mongo database (OPTIONAL)
+PORT=8000                             # Specifies the port in which the app will be exposed (OPTIONAL, when ran without using Docker)
```
miguelpduarte: The difference between `HOST_PORT` and `PORT` needs clarification. One is for node, another is for docker.

Maybe the names can be changed or the variables merged (docker can pass a default port to node and remap it externally, if this is preferred), but at least make the explanation actually be different 😅

miguelpduarte:
Made TravisCI's checks no longer be required, as that is a problem with project structure (tests using a db) / dockerfile :spaghetti:. Will :eyes: changes now

.gitignore
```diff
@@ -1,6 +1,6 @@
-# Extras
+# Env related
 .env
```
miguelpduarte: Should `.env` still be ignored? Is it not a way to provide fallback values?

From what I'm thinking `.env` could have the place of `.env.example` (if .example is not being used for anything else) - having the explanation of the variables and a default configuration as well!

.travis.yml
```diff
@@ -10,7 +10,7 @@ services:
 
 env:
   global:
-    - HOST_PORT=3000
+    - PORT=3000
     - DB_HOSTNAME=mongo
```
miguelpduarte: If `DB_HOSTNAME` was removed, can we instead use `MONGO_URI` here? I think a value of `mongodb://mongo:27017` should work because the DB container is in the same docker configuration (and network).

I removed TravisCI's checks from being required, but if this makes them pass again then it would be cool as well, and a small fix to try.

README.md
```diff
@@ -33,9 +33,9 @@ The best approach to install `docker-compose` is to follow the offical guide [he
 ## Usage
 
 ### Development Environment
-To start developing, you must create a file `.env` with environment variables, which are explained in more detail [below](#env-file-specification).
+To start developing, copy `.env.example`, creating the local config file (this can be achieved by running `cp .env.example .env.local`). Then, fill in the variable values, as explained in `.env.example`file.
```
miguelpduarte: Same suggestion listed above, using `.env` or even `.env.test` for documentation and default values might be better than having a dedicated file for that which might become outdated.

docker-compose.yml
```diff
@@ -7,9 +7,9 @@ services:
     volumes:
       - ./src:/usr/src/nijobs-be/src
     ports:
-      - ${HOST_PORT}:3000
+      - ${PORT}:${PORT}
     environment:
-      - PORT=3000
+      - PORT=${PORT}
       - NODE_ENV=prod
       - DB_HOSTNAME=mongo
```
miguelpduarte: Same suggestion as with the Travis configs above, switch this to use `MONGO_URI` as `DB_HOSTNAME` is no longer being used, I think :sweat_smile:

docker-compose.yml
```diff
@@ -18,29 +18,29 @@ services:
     build:
       context: .
       dockerfile: Dockerfile
+    ports:
+      - ${PORT}:${PORT}
```
miguelpduarte: As per the [docker-compose docs](https://docs.docker.com/compose/compose-file/#ports), please use a string here instead of direct numbers.

docker-compose.yml
```diff
@@ -18,29 +18,29 @@ services:
...
-    ports:
-      - "${HOST_PORT}:3001"
     depends_on:
       - mongo
   test:
     build:
       context: .
       dockerfile: Dockerfile-test
+    ports:
+      - ${PORT}:${PORT}
```
miguelpduarte: Same suggestion as above, please use strings instead of numbers

.gitignore
miguelpduarte: Either `.env` or `.env.test` - don't know which is more appropriate.

miguelpduarte: After taking a look at [the docs](https://www.npmjs.com/package/dotenv-flow) I believe that `.env` should not be ignored and added with the example values (preferably ones that also work, as much as possible, without exposing any sensitive data) - basically `.env` should be what `.env.example` is at the moment.

miguelpduarte:
Ended up implementing the review suggestions myself, in order to speed up the process and tackle other issues relative to refactoring the project.

miguelpduarte:
Well, tested everything to make sure it worked, but don't exactly feel ok with LGTM'ing myself... 😕
What to do?

PR #29, opened 2019-08-20
Replace bcrypt-nodejs by bcrypt

nip10:
bcrypt-nodejs is deprecated ([source](https://www.npmjs.com/package/bcrypt-nodejs)).
Also, bcrypt encourages the use of async methods ([source](https://github.com/kelektiv/node.bcrypt.js#why-is-async-mode-recommended-over-sync-mode))

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55,8 +55,7 @@
     "abbrev": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
-      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
-      "dev": true
+      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q=="
     },
     "accepts": {
       "version": "1.3.5",
@@ -109,8 +108,7 @@
     "ansi-regex": {
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-3.0.0.tgz",
-      "integrity": "sha1-7QMXwyIGT3lGbAKWa922Bas32Zg=",
-      "dev": true
+      "integrity": "sha1-7QMXwyIGT3lGbAKWa922Bas32Zg="
     },
     "ansi-styles": {
       "version": "3.2.1",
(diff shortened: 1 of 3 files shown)


nip10:
I've always used `bcryptjs` in my own projects, ever since I had the need to use a hashing function for authentication. I believe this was the most used package at that time. 
When I installed these packages and saw you were using `bcrypt-nodejs` I decided to take a look, and found that note, which made me aware of the `bcrypt` package.

There's also some discussion around the "best" hash function and I believe the current "best" choise is actually Argon2 ([eg](https://security.stackexchange.com/questions/193351/in-2018-what-is-the-recommended-hash-to-store-passwords-bcrypt-scrypt-argon2)).

I discovered NIAEFEUP the first time I used the timetable selector. I was actually going to do a v2 of that project until I realized you were already working on it. Since I was going the React route, and the project uses Vue, I kind of dropped that idea and went to other projects. (as a MIEEC student who had to learn web-dev on my own, I had to choose wisely)

I also started working on Website-NIAEFEUP but as I was working on it I felt that the project required a full refactor, so I also dropped that idea.

Today I decided to take a look and this project seemed like a good one to contribute to!

I have some issues ready to post and PRs to fix them, if that is fine by you.

miguelpduarte:
I have also seen `bcrypt` as the norm for quite some time now, and it seems that it has been in a good state for a while now.
However, as with everything related to technology new ideas and solutions appear every day - however, despite what many say, this does not mean that it is the best idea to jump to them straight away, despite the fact that they might be objectively better. The maturity of a technology is also very important, and if one "jumps ship" to something cutting edge, one might end up being cut as well... :smile:

miguelpduarte:
I have also seen `bcrypt` as the norm for quite some time now, and it seems that it has been in a good state for a while now.
However, as with everything related to technology new ideas and solutions appear every day - however, despite what many say, this does not mean that it is the best idea to jump to them straight away, despite the fact that they might be objectively better. The maturity of a technology is also very important, and if one "jumps ship" to something cutting edge, one might end up being cut as well... :smile: 

Password hashing extreme security is not something that leaves me extremely worried as well - this might not be the best thing to say, but I believe that there are many other points related to security to address before worrying about using the "best" algorithm for hashing passwords :sweat_smile: 

We're kind of in a tough spot with the Vue implementation as well, as most of the branch's members are/were learning React. However, porting the project is not something that we are very keen on... :confused: 

Website-NIAEFEUP is in fact probably going to need a full refactor, but works fine for now, despite not being ideal. So the focus might not be on it for a while, but instead on newer projects :slightly_smiling_face:
It was kind of rushed at the time, as the need for a better website was critical, so its "poor" state is understandable.

Sure, go ahead! We are all for open source (otherwise the code would be private haha)!
Just keep in mind that, as we are on vacations, you might not get feedback very quickly.

Also, this project is prone to enter "full refactor mode" sometime soon (as per #19 and several others) so some of the things that you end up doing might be deferred or simply closed as it will become out of scope given the change in direction of the project.

Once again, thanks for your contribution! I accepted this, so I think it can be merged now :slightly_smiling_face:

nip10:
Completly agree on the "cutting-edge" stuff. I didn't phrase that properly: From what I've read, I believe Argon2 is probably the most promising, not the current best ready-for-production.
Also agree on having other priorities regarding security, in this type of project.

Thanks for the input on the other projects. I understand why refactoring two projects (one of them being a complete rebuild) is not a priority; but I'll keep an eye anyways!

I've seen that article on Reddit (Bulletproof arch) and had it on my bookmarks for a future project (just like Next, Nest, and 30 other libs/frameworks.. JS right...) so I may continue contributing!

Have a good vacation!

PR #30, opened 2019-08-20
Add linting to tests

miguelpduarte:
- Complements #28, this was probably forgotten or would have been
included there
- Fix linting errors. This was mostly automatic but there was also some
manual work relative to following mocha's suggestions for using
async/await in tests for the auth tests
(https://mochajs.org/#using-async-await)

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -1,8 +1,10 @@
-//// Base file for unit test utilities
+//----------------------------------
+// Base file for unit test utilities
+//----------------------------------
 
 require("dotenv-flow").config();
 
-//During the test the env variable must be set to test
+// During the test the env variable must be set to test
 if (process.env.NODE_ENV !== "test") {
     console.error("Entered test files without being in the test environment, aborting!");
     process.exit(55);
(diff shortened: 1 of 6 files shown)


test/common.js
```diff
@@ -1,8 +1,8 @@
-//// Base file for unit test utilities
+// // Base file for unit test utilities
```
xRuiAlves: Double comment makes the compiler ignore this twice, right? 😂

test/sample_tests.js
```diff
@@ -20,18 +20,19 @@ describe("# Sample boilerplate tests", () => {
         });
```
xRuiAlves: This file could be probabbly removed by now, right?

miguelpduarte: Out of the scope of this PR, imo

Leaving it for #19 - that would be more inside scope and there will be a better space for discussion and thought on the removal of it

test/common.js
miguelpduarte: :man_facepalming: 

Thanks eslint!

Couldn't change the rules so just added moar comments :slightly_smiling_face:

PR #34, opened 2019-08-20
Add cross-env package

### LINKED ISSUE #31: Add Windows support
Although this project already uses Docker, which basically eliminates issues that arrise from using different environments, there are still some issues that cause errors on Windows.

The **TLDR** is that `npm test` and pushing to git don't work on Windows.

The `test` script doesn't work in Windows due to the way env vars are set, plus the way two consecutive commands are executed in cmd. The best solution that I'm aware of is to use the [cross-env](https://www.npmjs.com/package/cross-env) package. The issue with pushing to git is due to the pre-push hook that calls that same npm script.

(I was previously able to edit the code and run the tests by manually editing the npm script)

nip10:
Fixes #31

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -743,11 +743,19 @@
         "capture-stack-trace": "^1.0.0"
       }
     },
+    "cross-env": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/cross-env/-/cross-env-5.2.0.tgz",
+      "integrity": "sha512-jtdNFfFW1hB7sMhr/H6rW1Z45LFqyI431m3qU6bFXcQ3Eh7LtBuG3h74o7ohHZ3crrRkkqHlo4jYHFPcjroANg==",
+      "requires": {
+        "cross-spawn": "^6.0.5",
+        "is-windows": "^1.0.0"
+      }
+    },
     "cross-spawn": {
       "version": "6.0.5",
       "resolved": "https://registry.npmjs.org/cross-spawn/-/cross-spawn-6.0.5.tgz",
       "integrity": "sha512-eTVLrBSt7fjbDygz805pMnstIs2VTBNkRm0qxZd+M7A5XDdxVRWO5MxGBXZhjY4cqLYLdtrGqRf8mBPmzwSpWQ==",
-      "dev": true,
       "requires": {
         "nice-try": "^1.0.4",
         "path-key": "^2.0.1",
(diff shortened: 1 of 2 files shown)


miguelpduarte:
Closing as per the discussion in #31. Removing git hooks will most likely be enough to improve windows support (#39).

miguelpduarte:
By the way, a suggestion for branch naming: use `fix/...`, `feature/...` et cetera, but note that `fix(something)/something` is not very common (at least in my experience) and is annoying because shells evaluate expressions structured like that, and thus there is a need to either escape it or make the branch name a string (using `'` or `"`).

This happened to me in `bash` and I'm certain it happens as well in `zsh`, which are most likely the two most used shells (excepting windows, but it also supports `bash` via WSL, so kinda windows too).

I'd say that the parens do not add that much (relevant) information, and thus can just be removed.

nip10:
The document I've used for commit naming schemas describes it as:

<type>(<scope>): <subject>
<body>
<footer>

But I've seen a couple different ones (like replacing : by / or having header+body). I'll start doing the way you guys do it here. I'd suggest adding this to the contributing.md ?

miguelpduarte:
What do you mean by commit naming schemas? The commit message?

I'm referring to the branch name here, the commit message itself does not bother me as much.
If I could, I would enforce [several rules](https://chris.beams.io/posts/git-commit/#seven-rules) that I think are a synonym of clean and organized git history, but since the branch changes members a lot and we can't control absolutely everything, I'm okay with being looser there.

I pointed out the branch name as I've found it to be a minor annoyance (the problem with having to escape parentheses I mentioned), just that.

Furthermore, I've never heard of the schema you mentioned in any git-related environment :sweat_smile: I'm not at all very experienced but have seen some repos maintained by people that I believe have some and I've never seen anything like it, hence me thinking it was strange (given I had never seen or heard of something like it before).

We do not have a contributing.md at all yet, since despite the projects being open source (and us being all in for that!) this was not a main concern - being the major one internal creation and consumption by the branch members.
As such, there most likely are several things that open source projects should have that we don't.

nip10:
I never had any issues with that (escaping the ()s) so this is news to me! I've also never had this discussion because the few projects that I've worked on, which I didn't maintain, were C# (Windows devs) and commits/branches names were a bit random.
Fun-fact: We don't even talk about Git in MIEEC.

I understand why its not a priority, but a single file can probably be used in all repos!

PR #35, opened 2019-08-21
Bump packages versions

nip10:
`npm audit` reports 41 vulnerabilities (39 high, 2 moderate)
These vulnerabilities were caused by `eslint`, `husky`, `nodemon `and `mongoose`.
Since these packages needed updates, I decided to update all packages. This resulted in 0 vulnerabilities.

Checked changelogs of packages which were updated in major versions (`eslint`, `husky`, `mocha`) and didn't find anything that needs to be changed.

However, Mongoose changed something in [v5.5.0](https://github.com/Automattic/mongoose/blob/master/History.md#550--2019-04-08) which causes failure in two tests. ([maybe related](https://github.com/Automattic/mongoose/pull/7744)).
Basically, in the `fields` and `technologies` fields of the `Ad` collection, the validation error message now prints the array index.

Before:
`"`BACK_END!!THIS_DOES_NOT_EXIST_FOR_SURE-00` is not a valid enum value for path `fields`."`
After:
`"`BACK_END!!THIS_DOES_NOT_EXIST_FOR_SURE-00` is not a valid enum value for path `fields.0`."`

The fix is straight forward but I'm not sure this PR is the right place to do it. Waiting for opinions on this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,18 +5,18 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
-      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
+      "version": "7.5.5",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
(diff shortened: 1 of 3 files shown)


miguelpduarte:
Yay for updates and for regression testing!

This looks good overall, thanks!

I'd say that this would, in fact, be the right place to do this. If bumping the packages causes test failure, then the same PR should address the problem - I think this should be the right way to proceed.

However, I'd block this on #37, just to ensure that the package updates do not cause any more problems (the additional tests implemented there should also pass).

imnotteixeira:
Closing this as stale.

`mocha` and `husky` are no longer being used.

Furthermore, the auditing problems have been fixed in the meantime, and the mongoose bumping problems have been dealt with as well

PR #36, opened 2019-08-21
Changed erase account hooks to only perform before and after the tests

GuilhermeJSilva:
Attempts to solve problems in #22 by changing the hook that caused the timeout.

### DIFF
--- a/test/auth.js
+++ b/test/auth.js
@@ -9,9 +9,9 @@ const Account = require("../src/models/Account");
 
 describe("Register endpoint test", () => {
 
-    afterEach("Clearing accounts", async () => {
-        await Account.deleteMany({});
-    });
+    before("Clearing accounts", () => Account.deleteMany({}));
+    after("Clearing accounts", () => Account.deleteMany({}));
+
 
     it("Should return a malformed request (missing username)", async () => {
         const res = await request()
--- a/test/auth.js
+++ b/test/auth.js
@@ -7,134 +7,127 @@ const {
 const ERROR_TYPES = require("../src/routes/errors/errorHandler");
 const Account = require("../src/models/Account");
 
-describe("Register endpoint test", () => {
+describe("# Authentication endpoint tests", () => {
+    describe("Register endpoint", () => {
 
-    before("Clearing accounts", () => Account.deleteMany({}));
-    after("Clearing accounts", () => Account.deleteMany({}));
+        describe("Validation tests", () => {
+            describe("Missing required fields", () => {
+                it("Should return a malformed request (missing username)", () => request()
+                    .post("/api/auth/register")
+                    .send({}).then((res) => {
+                        res.should.have.status(400);
+                        res.body.should.be.an("object");
+                        res.body.should.have.property("success").equal(false);
+                        res.body.should.have.property("reason").that.is.a("string");
+                        res.body.should.have.property("error_code").equal(ERROR_TYPES.MISSING_FIELD);
+                    }));
 
+                it("Should return a malformed request (missing password)", () => request()
+                    .post("/api/auth/register")
+                    .send({
+                        username: "user",
+                    }).then((res) => {
+                        res.should.have.status(400);
+                        res.body.should.be.an("object");
+                        res.body.should.have.property("success").equal(false);
+                        res.body.should.have.property("reason").that.is.a("string");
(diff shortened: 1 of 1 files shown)


test/auth.js
```diff
@@ -9,9 +9,9 @@ const Account = require("../src/models/Account");
 
 describe("Register endpoint test", () => {
 
-    afterEach("Clearing accounts", async () => {
-        await Account.deleteMany({});
-    });
+    before("Clearing accounts", () => Account.deleteMany({}));
+    after("Clearing accounts", () => Account.deleteMany({}));
```
miguelpduarte: The basis for the change here is that the methods themselves return a `Promise` and this will make the hooks await the promise, correct?

Despite this, I'm not sure if this is very clear. I feel that this might result in someone adding braces to the arrow function and then not knowing that an `await` is necessary.

*TL;DR*: Explicit > Implicit, it is better to handle promises explicitly than "hiding" them like this.
But this can use some other feedback, it is open for discussion.

test/auth.js
```diff
@@ -9,9 +9,9 @@ const Account = require("../src/models/Account");
 
 describe("Register endpoint test", () => {
 
-    afterEach("Clearing accounts", async () => {
```
miguelpduarte: A problem I have with this PR is that despite this being a good idea because it cleans up the test code and minimizes the number of DB accesses, which is always good, I believe that it might not solve the problem that happened in the other PR, as it seems weird that this was the problem.

My guess is a DB misconfiguration, because this is probably the first bit of test code that actually has to use the DB and thus the test times out. Probably something like that.

Finally, I think that the tests could be refactored a bit in terms of organization. The before and after hooks that are being changed here seem to only be needed for the final test inside this describe block.
I'd say that that calls for better organization of the tests, as they could be scoped like "Authentication tests" > "when a user already exists" > "something" or "Authentication tests" > "parameter validations/input validation" > "something".

More describe scopes might seem like clutter but might make the tests and test hooks more organized :slightly_smiling_face:

test/auth.js
GuilhermeJSilva: The objective is to return a Promise, if I understood correctly the suggestion is to include bracers and the return keyword, however eslint is configured so that this is not possible.

xRuiAlves:
Please squash the `Fixed typo` commit 😅

miguelpduarte:
@xRuiAlves I know what you're doing, and despite me not being entirely opposed to it, I'm not sure if we'll be able to enforce that level of commit juiciness (👌) everywhere. Just leaving that note here.

test/auth.js
miguelpduarte: I know that the objective is to return a promise 😅 I might not have been very clear, I'll try to explain it better:

Returning a Promise in a test hook makes the test execution wait for the hook completion.
However, returning a Promise there is being implicit behaviour (due to how arrow functions work).

I get what you mean about the `eslint` rules, but that was not what I had in mind.
I was thinking of something like:

```js
before("Hook", async () => {
    await Async.stuff();
});
```

Due to the way that `async` methods work, that is functionally the same as returning a `Promise`.
However, if done this way, it will be clear that the hook will wait for the async stuff to resolve, because it is more explicit.

My recommendation is to switch to this style in order to make sure that the code is clearer for other people that might look at it and reduce possible copy paste errors that might occurr (someone might copy paste the function and add braces and more statements, and then the promise would not be returned anymore - thus making the methods in the hook not be `await`ed for).

miguelpduarte: This can (and should, imo) also be changed in tests, that I see are following a similar pattern.  This would also make them more consistent with what's being done in #38 😄

miguelpduarte:
PS: Forced the failing build to re-run, and now it's passing... #JustTravisThings I guess - We should reeeeeally switch to CircleCI :/

miguelpduarte:
@GuilhermeJSilva Was this not tackled on #60 or the other PR in which I changed the tests from mocha+chai to jest? Want to close the dangling branches, since we currently have a lot of them, and a lot of open PRs I also would like to slim down :slightly_smiling_face: 

IMO we should close this and check if the issue still occurs. If so, then open this again as the code has changed a lot and maintaining this with the conflicts would also be unnecessary work.

Please give your input on this :slightly_smiling_face:

GuilhermeJSilva:
I think the changes this PR was trying to make have already been introduced (the delete hooks in account only occur per test group), added to the fact that the code being changed is substantially outdated I'm closing this PR

Issue #33, opened 2019-08-20
Improve validation


nip10:
In the /register route, both username and password lack proper validation. A string consisting of spaces will be valid and saved in the database.

A possible solution is to use `trim()` and then `length`. A more robust (but more complex) solution is to use something like [@hapi/joi](https://www.npmjs.com/package/@hapi/joi).

Mongoose models can also have `required` and `validate` fields.

miguelpduarte:
I see no problem with allowing spaces in the password. I am not entirely up to date on web development good practices for sure, but I don't see why the password's characters should be restricted at all.

Relative to the username - there I'd say you have a point. We could decide on certain requirements and follow them - such as "only alphanumerical characters" or something like that.

Finally, on using `@hapi/joi` - I'd have to say that I'm against it because despite seeming like a more robust solution, what I really see is just one more package that does not simplify things all that much - we could (and should, imo) make our own validators, which we can reuse to make more complex ones. The logic that it abstracts does not seem to be enough justification to make it worthwhile.

As you can see, I'm very purist in what concerns utility packages :sweat_smile:
I really hate bloat and will many times prefer to implement stuff myself instead of relying on other stuff.

I know, I know, "don't reinvent the wheel!" - However, my wheel is not dependant on other people making it not spin anymore without apparent reason, or providing no help when the wheel stops working for no reason! (which sadly is many times the case for `node` packages)

However, of course, this is not set in stone - if it makes sense to add later, then of course that would be best! It just doesn't seem to me like a necessary course of action for now, so I'd rather go with the simplest approach.

nip10:
The issue is not having spaces, but being entirely spaces. The `trim `function will only remove leading and trailing spaces, not spaces in the middle. (but since you mentioned it, I'm pro disallowing any spaces at all)

Using Joi was only a suggestion. I find it pretty useful (eg email validation, error handling) but I can see it being a bit overhead for the current schemas.

PR #37, opened 2019-08-21
More validation in account information

GuilhermeJSilva:
Includes:
* Added maximum and minimum lengths to the username field in account
* Added tests for the schema (#21)

Still missing:
* Minimum length for the password
* A consistent way to test input validation at the endpoints (#33)

### DIFF
--- a/src/routes/auth.js
+++ b/src/routes/auth.js
@@ -47,6 +47,12 @@ router.post("/register", async (req, res) => {
             "reason": "Username is too long",
             "error_code": ERROR_TYPES.MISSING_FIELD,
         });
+    } else if (req.body.username.length < 3) {
+        return res.status(500).json({
+            "success": false,
+            "reason": "Username is too short",
+            "error_code": ERROR_TYPES.MISSING_FIELD,
+        });
     }
 
     // Password is required
--- a/src/routes/auth.js
+++ b/src/routes/auth.js
@@ -41,6 +41,12 @@ router.post("/register", async (req, res) => {
             "reason": "No username specified",
             "error_code": ERROR_TYPES.MISSING_FIELD,
         });
+    } else if (req.body.username.length > 20) {
+        return res.status(500).json({
+            "success": false,
+            "reason": "Username is too long",
+            "error_code": ERROR_TYPES.MISSING_FIELD,
+        });
     }
 
     // Password is required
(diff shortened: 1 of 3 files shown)


src/routes/auth.js
```diff
@@ -41,6 +41,18 @@ router.post("/register", async (req, res) => {
...
+            "reason": "Username is too long",
+            "error_code": ERROR_TYPES.MISSING_FIELD,
+        });
+    } else if (req.body.username.length < 3) {
+        return res.status(500).json({
+            "success": false,
+            "reason": "Username is too short",
+            "error_code": ERROR_TYPES.MISSING_FIELD,
+        });
     }
```
miguelpduarte: Would you oppose to moving this to a separate file (of validators)?

I'm thinking that this type of input checking could be done by middlewares (which are many times called validation middlewares or validators), and this would also reduce the friction with #38's changes (and goes along #19's ideas).

This would also reduce the code in the controller, making it simpler to understand :slightly_smiling_face:

src/routes/auth.js
```diff
@@ -41,6 +41,18 @@ router.post("/register", async (req, res) => {
             "reason": "No username specified",
             "error_code": ERROR_TYPES.MISSING_FIELD,
         });
+    } else if (req.body.username.length > 20) {
+        return res.status(500).json({
+            "success": false,
+            "reason": "Username is too long",
+            "error_code": ERROR_TYPES.MISSING_FIELD,
+        });
+    } else if (req.body.username.length < 3) {
```
miguelpduarte: Also, please remove the `else`s if each if already returns :sweat_smile: 

It's functionally the same but results in reduced visual clutter in my opinion :slightly_smiling_face:

test/account_schema.js
```diff
@@ -0,0 +1,69 @@
...
+                        .equal(`Path \`username\` (\`${username}\`) is shorter than the minimum allowed length (3).`);
+                });
+            });
+
+            it("Above miximum should throw error", () => {
+                const username = "thisisa20pluscharacterusermane";
+                const account = new Account({
+                    username,
+                });
+                return account.validate((err) => {
```
miguelpduarte: Nitpick: add newline here (to separate variable definitions from the asserting).

nip10:
Validation errors sent to the client should have a 4XX status code.
I suggest sending 400 if a field is missing and 422 if validation fails (but sending 400 in any case is not that bad)

PR #38, opened 2019-08-21
Application Structural Refactor

### LINKED ISSUE #19: Improve code structure

### LINKED ISSUE #25: Switch tests from Mocha+Chai to Jest
I now believe that Jest tests are much cleaner. Plus, Jest can run tests in parallel for better performance and has mocking built-in, as well as coverage reports!

miguelpduarte:
Closes #19. Closes #25.

Currently going for a mix of the Bulletproof architecture mentioned in the issue, Ruby on Rails (it has a pretty good architecture imo, despite ruby being weird) and some of my own opinions.

This is in the very early stages, but opinions are welcome as this is a core concern for future development.


miguelpduarte:
Also closes #25

.env
```diff
@@ -13,4 +13,4 @@ DB_NAME=nijobs
 SESSION_SECRET=O Rui foi membro do IEEE.
 
 # Specifies the port in which the app will be exposed
-PORT=8000
+PORT=8087
```
imnotteixeira: Why this change? Anything in particular?

src/loaders/express.js
```diff
@@ -23,33 +25,35 @@ module.exports = (app) => {
...
+    // Adds route logging
+    app.use(morgan("common"));
 
-        // Request methods to allow
+    // Adding headers (CORS)
+    app.use((_, res, next) => {
+        // Allow connections for all origins
+        res.setHeader("Access-Control-Allow-Origin", "*");
+        // Allowed request methods
         res.setHeader("Access-Control-Allow-Methods", "GET, POST, OPTIONS, PUT, PATCH, DELETE");
```
imnotteixeira: Are we using all of these? Shouldn't we remove the ones we don't use?

src/loaders/express.js
```diff
@@ -23,33 +25,35 @@ module.exports = (app) => {
...
-    // Website to allow to connect
-        res.setHeader("Access-Control-Allow-Origin", "*");
+    // Adds route logging
+    app.use(morgan("common"));
 
-        // Request methods to allow
+    // Adding headers (CORS)
+    app.use((_, res, next) => {
+        // Allow connections for all origins
+        res.setHeader("Access-Control-Allow-Origin", "*");
```
imnotteixeira: We should have an `env` variable with the allowed origins, for more safety. It is not that hard and we could customize it to allow only our applications for now (nijobs-fe, for example). Surely for development environment, we could have the `*` the same way.

.env
miguelpduarte: Collided with a lot of stuff that opens listens on 8000 by default.

src/loaders/express.js
miguelpduarte: This is CORS, which will be necessary for any API.

.env
imnotteixeira: :+1:

src/loaders/express.js
miguelpduarte: If you are mentioning the allowed methods, this was probably just a left-over copy pasta from somewhere online for enabling CORS for node+express

src/loaders/express.js
miguelpduarte: This is a valid future concern, please create an issue mentioning your comment for increased context.

src/loaders/express.js
imnotteixeira: I was asking if all the http methods there will be used, if not, they can be removed from the `Access-Control-Allow-Methods` Header

miguelpduarte: I don't really see a problem, as they would not match any endpoint and just return a 404. But I get what you mean.

miguelpduarte:
Derped with the force pushes, fixed the problem and amended the commit but forgot to add the files 🙃 

Sorry for the spam there 😅

src/loaders/express.js
imnotteixeira: Here you go: #41

miguelpduarte: Thanks!

miguelpduarte:
The tests were hanging, and jest traced it to some `supertest` stuff. Found https://github.com/nodkz/mongodb-memory-server and in fact it seems that the problem was not disconnecting from the `mongo` server in the end of the test suites 🙃

src/api/middleware/errorHandler.js
```diff
@@ -0,0 +1,56 @@
+const { isCelebrate } = require("celebrate");
+const EscapeHtml = require("escape-html");
+
+const { errorExtractor } = require("../../lib/dbErrorExtractor");
+
+const ERROR_TYPES = Object.freeze({
```
imnotteixeira: Let's rename this variable to ErrorTypes, so it can be used as ErrorTypes.FORBIDDEN, for example, I think it is more common, at least in java. (does not have to be changed here, but it should, to ensure consistency)

src/lib/dbErrorExtractor.js
```diff
@@ -0,0 +1,41 @@
...
+    UNDETERMINED_ERROR: 1,
+    DUPLICATED_KEY: 2,
+});
+
+const errorExtractor = (err) => {
+    if (handlers.hasOwnProperty(err.code)) {
+        return handlers[err.code](err);
+    }
+    return {
+        type: DB_ERROR_TYPES,
```
imnotteixeira: This should be an actual error type, not the enum :D

src/lib/dbErrorExtractor.js
```diff
@@ -0,0 +1,41 @@
+const DB_ERROR_TYPES = Object.freeze({
```
imnotteixeira: Same thing, change to `DBErrorTypes`

src/loaders/express.js
```diff
@@ -0,0 +1,67 @@
...
+        res.setHeader("Access-Control-Allow-Headers", "X-Requested-With, content-type, authorization");
+        // Because we need the website to include cookies in the requests sent
+        // to the API (we are using sessions)
+        res.setHeader("Access-Control-Allow-Credentials", true);
+        // Continue to next layer of middleware
+        return next();
+    });
+
+    // Health check endpoint
+    app.get("/", (_, res) => res.status(200).json({ "online": true }));
```
imnotteixeira: The fact that the server is answering (200 response) is enough to know that it is online. Instead, we could send the API version in the response (add it to .env?)

src/services/auth.js
```diff
@@ -0,0 +1,21 @@
+// TODO: Use typedi or similar
+const Account = require("../models/Account");
+
+class AuthService {
+    constructor() {
```
imnotteixeira: Empty constructor is not required

src/services/auth.js
```diff
@@ -0,0 +1,21 @@
...
+
+    }
+
+    async register(username, password) {
+        const account = await Account.create({
+            username: username,
```
imnotteixeira: You can use the smaller syntax 
```js
{
    username,
    password
}
```

src/services/auth.js
```diff
@@ -0,0 +1,21 @@
...
+            password: password,
+        });
+
+        return {
+            username: account.username,
```
imnotteixeira: We can use `{ username }` directly, since it will be the same as `account.username`

src/setupTests.js
```diff
@@ -0,0 +1,35 @@
...
+//----------------------------------
+
+// During the test the env variable must be set to test
+if (process.env.NODE_ENV !== "test") {
+    console.error("Entered test files without being in the test environment, aborting!");
+    process.exit(55);
+}
+
+const mongoose = require("mongoose");
+// Ocurrs before all the tests, only once
```
imnotteixeira: Grammarly your code: `occurs`

src/api/middleware/errorHandler.js
```diff
@@ -32,10 +34,23 @@ const handler = () => (err, req, res, next) => {
...
+        return next(err);
+    }
+
+    const result = {
+        success: false,
+        error_code: ERROR_TYPES.DB_ERROR,
+        error: errorExtractor(err),
+    };
+
+    return res.status(500).send(result);
```
miguelpduarte: Currently, in the only case that this happens the status should probably be 400. But we should revisit this later, the status should be calculated from some more context, such as a `req.something` added by the request handler to give context to the db error handler.

src/api/middleware/validators/auth.js
```diff
@@ -0,0 +1,10 @@
+const { celebrate, Joi } = require("celebrate");
```
miguelpduarte: We should switch to [`express-validator`](https://www.npmjs.com/package/express-validator) because of this: https://www.npmtrends.com/celebrate-vs-express-validator

src/loaders/express.js
miguelpduarte: The json content could be empty, this content is just so that it is not, for easier checking of the app being online via a browser or something that might hide an empty response.

miguelpduarte: I don't think we have discussed versioning, so I'd rather not dive into that now (at least not in this PR, it's already huge)

src/services/auth.js
miguelpduarte: This had the purpose of setting up dependency injection using something like `typedi`, but was stopped in the interim due to other things being more relevant at the time.

miguelpduarte: However, it can be removed without any issue

src/services/auth.js
miguelpduarte: Ideally so, but some hooks in the Model might change it, due to trimming or something of the sort. I see no problem with leaving this as is.

src/setupTests.js
miguelpduarte: :cry:

miguelpduarte:
Resolved the conflicts via rebasing develop into this branch

src/services/auth.js
imnotteixeira: bump

miguelpduarte: A bit nitpicking, but fixed :stuck_out_tongue:

PR #47, opened 2019-11-21
Add API Rate limit + security improvements

### LINKED ISSUE #46: Limit API access rate + Security Fixes
According to LGTM analysis, we are performing database accesses without limiting the API calls. After some research, I found that the most used (and it seems to be an alive project, so it is updated) is express-rate-limit (https://www.npmjs.com/package/express-rate-limit).

The alternatives to store the access cache are the node process internal memory, redis, memcached and mongo.

As we might deploy the server in a cluster in the future, across multiple node processes in parallel, it's important to store the accesses in an external cache.  redis and memcached are optimized for reading, and it seems the implementation is quite easy, so I recommend using the redis approach, by adding this package as well: https://www.npmjs.com/package/rate-limit-redis.

Additionally, after looking at the express recommendations for security (https://expressjs.com/en/advanced/best-practice-security.html), we should adopt some of them (some are already in place, yay) such as the helmet plugin (https://expressjs.com/en/advanced/best-practice-security.html#use-helmet)

imnotteixeira:
Closes #46 

Choose to use the mongo storage approach, for the reasons described in the issue.

Also added the helmet dependency, as recommended in the express docs, linked in the issue.

### DIFF
--- a/src/config/env.js
+++ b/src/config/env.js
@@ -1,13 +1,13 @@
 require("dotenv-flow").config();
 
-module.exports = Object.freeze({
+const config = Object.freeze({
     // Database
     db_host: process.env.DB_HOST,
     db_port: process.env.DB_PORT,
+    db_name: process.env.DB_NAME,
     db_uri: process.env.DB_URI,
     db_user: process.env.DB_USER,
     db_pass: process.env.DB_PASS,
-    db_name: process.env.DB_NAME,
 
     session_secret: process.env.SESSION_SECRET,
 
@@ -20,4 +20,13 @@ module.exports = Object.freeze({
     admin_email: process.env.ADMIN_EMAIL,
     admin_password: process.env.ADMIN_PASSWORD,
     access_control_allow_origin: process.env.ACCESS_CONTROL_ALLOW_ORIGIN || "*",
+});
+
+module.exports = Object.freeze({
+    ...config,
+    db_uri: config.db_uri ||
+        (config.db_host
+            && config.db_port
+            && config.db_name
+            && `mongodb://${config.db_host}:${config.db_port}/${config.db_name}`),
 });
--- a/src/config/env.js
+++ b/src/config/env.js
@@ -1,17 +1,18 @@
 require("dotenv-flow").config();
 
-const config = Object.freeze({
+const generateDBUriFromEnv = () => {
+    if (!process.env.DB_HOST || !process.env.DB_PORT || !process.env.DB_NAME)
+        throw new Error("Missing DB Params to generate URI. Either pass a DB_URI or (DB_HOST, DB_PORT and DB_NAME) in .env");
+
+    return `mongodb://${process.env.DB_HOST}:${process.env.DB_PORT}/${process.env.DB_NAME}`;
+};
+
+module.exports = Object.freeze({
     // Database
     db_host: process.env.DB_HOST,
     db_port: process.env.DB_PORT,
     db_name: process.env.DB_NAME,
-    db_uri: process.env.DB_URI ||
-    (
-        process.env.DB_HOST
-            && process.env.DB_PORT
-            && process.env.DB_NAME
-            && `mongodb://${process.env.DB_HOST}:${process.env.DB_PORT}/${process.env.DB_NAME}`
-    ),
+    db_uri: process.env.DB_URI || generateDBUriFromEnv(),
     db_user: process.env.DB_USER,
     db_pass: process.env.DB_PASS,
 
@@ -24,5 +25,3 @@ const config = Object.freeze({
     admin_password: process.env.ADMIN_PASSWORD,
     access_control_allow_origin: process.env.ACCESS_CONTROL_ALLOW_ORIGIN || "*",
 });
-
-module.exports = config;
--- a/src/config/env.js
+++ b/src/config/env.js
@@ -5,28 +5,24 @@ const config = Object.freeze({
     db_host: process.env.DB_HOST,
     db_port: process.env.DB_PORT,
     db_name: process.env.DB_NAME,
-    db_uri: process.env.DB_URI,
+    db_uri: process.env.DB_URI ||
+    (
+        process.env.DB_HOST
+            && process.env.DB_PORT
+            && process.env.DB_NAME
+            && `mongodb://${process.env.DB_HOST}:${process.env.DB_PORT}/${process.env.DB_NAME}`
+    ),
     db_user: process.env.DB_USER,
     db_pass: process.env.DB_PASS,
 
+    // App
     session_secret: process.env.SESSION_SECRET,
-
     port: process.env.PORT,
-
     god_token: process.env.GOD_TOKEN,
-
     test_log_requests: JSON.parse(process.env.TEST_LOG_REQUESTS),
-
     admin_email: process.env.ADMIN_EMAIL,
     admin_password: process.env.ADMIN_PASSWORD,
     access_control_allow_origin: process.env.ACCESS_CONTROL_ALLOW_ORIGIN || "*",
 });
 
-module.exports = Object.freeze({
-    ...config,
-    db_uri: config.db_uri ||
-        (config.db_host
-            && config.db_port
-            && config.db_name
-            && `mongodb://${config.db_host}:${config.db_port}/${config.db_name}`),
-});
+module.exports = config;
--- a/src/loaders/express.js
+++ b/src/loaders/express.js
@@ -44,19 +44,22 @@ module.exports = (app) => {
 
     const mongo_uri = config.db_uri || `mongodb://${config.db_host}:${config.db_port}/${config.db_name}`;
 
-    const api_rate_limiter = new RateLimit({
-        store: new MongoStore({
-            uri: mongo_uri,
-        }),
-        windowMs: API_REQUEST_TIME_WINDOW_MS,
-        max: API_MAX_REQUESTS_PER_WINDOW,
-        message: {
-            error: "Too many requests issued from this IP, please try again after a while",
-        },
-    });
+    // Set the API call rate limit only on production
+    if (process.env.NODE_ENV === "production") {
+        const api_rate_limiter = new RateLimit({
+            store: new MongoStore({
+                uri: mongo_uri,
+            }),
+            windowMs: API_REQUEST_TIME_WINDOW_MS,
+            max: API_MAX_REQUESTS_PER_WINDOW,
+            message: {
+                error: "Too many requests issued from this IP, please try again after a while",
+            },
+        });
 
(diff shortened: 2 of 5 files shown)


src/loaders/express.js
```diff
@@ -34,6 +41,28 @@ module.exports = (app) => {
     app.use(passport.initialize());
     app.use(passport.session());
 
+    const mongo_uri = config.db_uri || `mongodb://${config.db_host}:${config.db_port}/${config.db_name}`;
```
GuilhermeJSilva: We have 2 separate definitions of this, even tough they are equal it would be best to have only one so if it ever needs be changed it would always stay consistent

src/loaders/express.js
```diff
@@ -34,6 +41,28 @@ module.exports = (app) => {
...
+            message: {
+                error: "Too many requests issued from this IP, please try again after a while",
+            },
+        });
+
+        // Adds rate limit to all routes
+        app.use(api_rate_limiter);
+    }
+
+    // Adds protection to common attacks
```
GuilhermeJSilva: It would be better to include a link to the list of protections it provides

src/loaders/express.js
imnotteixeira: Made it so that the uri is always available if it is provided, or if host port and db name are provided, that way, one can always access db_uri directly. There were already checks to assert that there exists either db_uri or db host/port and name. Ran the tests and it successfully connected so it should be fine.

imnotteixeira:
Fixed conflicts with `develop`, should be ready now.

src/loaders/express.js
```diff
@@ -35,6 +42,29 @@ module.exports = (app) => {
     app.use(passport.initialize());
     app.use(passport.session());
 
+    const mongo_uri = config.db_uri;
+
+    // Set the API call rate limit only on production
+    if (process.env.NODE_ENV === "production") {
```
miguelpduarte: We need to make sure that the servers in production are actually launched with `NODE_ENV=production` because of this. Commenting here so that I don't forget to quickly check this.

src/config/env.js
```diff
@@ -20,3 +20,12 @@ module.exports = Object.freeze({
...
     admin_password: process.env.ADMIN_PASSWORD,
 });
+
+module.exports = Object.freeze({
+    ...config,
+    db_uri: config.db_uri ||
+        (config.db_host
+            && config.db_port
+            && config.db_name
+            && `mongodb://${config.db_host}:${config.db_port}/${config.db_name}`),
```
miguelpduarte: The uri that was being used in the mongoose loader did not use the db_name directly, but instead used it in the options for the connection. Please change this so that the db_name is not used here.

Uses that desire to know the db_name shall fetch it as they need to, but db_uri should just be the uri to the mongodb database.

src/loaders/express.js
miguelpduarte: The problem is not in connecting to the tests, as they provide db_host and db_port. The problem is that the new changes would probably not work in production, in which the mongodb cluster uri is provided :sweat_smile:

I think the suggested changes solve this though.

src/loaders/express.js
miguelpduarte: `niployments` runs `Dockerfile-prod`. `nijobs-be`'s `Dockerfile-prod` runs `npm run prod`. However, `npm run prod` does not set `NODE_ENV` to `production`.

So, if you don't mind, this needs to be set in package.json for this PR to be effective :slightly_smiling_face:

miguelpduarte: (Furthermore, we should really update the node versions in the dockerfiles, as they are being used for production, CI and etc. Creating an issue to track that now)

src/loaders/express.js
```diff
@@ -8,6 +8,13 @@ const apiRoutes = require("../api");
 const config = require("../config/env");
 const { dbHandler } = require("../api/middleware/errorHandler");
 
+const helmet = require("helmet");
+
+const RateLimit = require("express-rate-limit");
+const MongoStore = require("rate-limit-mongo");
+
+const API_REQUEST_TIME_WINDOW_MS = 15 * 60 * 1000; // 15 minutes
+const API_MAX_REQUESTS_PER_WINDOW = 100;
```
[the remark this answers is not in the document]
imnotteixeira: Yeah, I don't know if those values are well thought, they are per IP, not in total, I think I got those from some default I found, but they can be changed

src/config/env.js
imnotteixeira: There is no case where you need the URI without the db_name, I updated the mongo connection call to pass the complete URI instead of a db_name param in the options which has the same effect. The other cases always use the complete URI. Which makes sense, in the way they are connecting to a specific db instance, not to the server where the db is located

miguelpduarte: I believe that we found out that this is not the case for the test environment :sweat_smile:

src/loaders/express.js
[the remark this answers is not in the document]
miguelpduarte: Okay, I guess we can roll with this for now and later on see if changes are necessary :slightly_smiling_face:

imnotteixeira:
Hey @miguelpduarte, just tested this and pushed the solution for the db_uri problem:
Basically, db_uri will be set (even if not provided), however, for example on the test scenario, if db_name is overriden, it will use that (i.e. if uri is mongodb:/<something>/db_name_1, and the .env.test sets db name to db_name_2, the connection will use db_name_2 even though the uri already has another db name in it)

GuilhermeJSilva (dismissed):

src/loaders/express.js
miguelpduarte: Resolving as this was tackled

miguelpduarte:
I need to give this a new good look, as it has been a while so this should either just be closed or merged instead of being eternally zombie. @imnotteixeira can you please test with a URI like the ones used in staging? (the cluster ones) I think that there may still be some issues with that format when running.

src/config/env.js
```diff
@@ -1,22 +1,27 @@
...
     db_port: process.env.DB_PORT,
-    db_uri: process.env.DB_URI,
+    db_name: process.env.DB_NAME,
+    db_uri: process.env.DB_URI ||
+    (
+        process.env.DB_HOST
+            && process.env.DB_PORT
+            && process.env.DB_NAME
+            && `mongodb://${process.env.DB_HOST}:${process.env.DB_PORT}/${process.env.DB_NAME}`
+    ),
```
miguelpduarte, line 14: If we are to go with this solution we should just make a method for this, I think it would be simpler or at least abstract from the rest of the configuration.

src/config/env.js
```diff
@@ -1,22 +1,27 @@
...
-
     god_token: process.env.GOD_TOKEN,
-
     test_log_requests: JSON.parse(process.env.TEST_LOG_REQUESTS),
-
     admin_email: process.env.ADMIN_EMAIL,
     admin_password: process.env.ADMIN_PASSWORD,
 });
+
+module.exports = config;
```
miguelpduarte, line 27: Rationale for this change? I liked the `module.exports =` approach and don't really see reasons to change

src/loaders/express.js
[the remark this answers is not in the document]
miguelpduarte: Going back to this, I think we should increase this a bit. Especially given the idea to have infinite scrolling on the search page. Scrolling quickly with a lot of entries might easily go over 100 requests in 15 minutes, given that the person might refresh the page and then go back to browsing for a bit.
However, like you said at the time - these values are heavily based on speculation, since we don't have real user data for now. As such we probably can just go with any value and adapt later. I just think it might be better to overshoot than undershoot 😛

miguelpduarte (changes requested):
I think we might be able to quickly move this forward.
By the way, I'm concerned that the rate-limiting might impact the free tier of storage we have going for the current MongoDB hosting solution. Can you provide any insight relative to this? We should probably be careful to not go over the quota.

src/config/env.js
imnotteixeira: I'm actually not sure why it changed, maybe I was exporting additional things and changed the structure, but it can be reverted.

src/loaders/express.js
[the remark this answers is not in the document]
imnotteixeira: I honestly think this is a good starting point since for the search it will probably be only 4-5 reloads (We won't have 1000 active offers in the near future - I hope we reach it someday, though :P)

I don't oppose changing the limit, it's just important to have one, really. But I don't think this will cause us problems for now.

imnotteixeira:
> I need to give this a new good look, as it has been a while so this should either just be closed or merged instead of being eternally zombie. @imnotteixeira can you please test with a URI like the ones used in staging? (the cluster ones) I think that there may still be some issues with that format when running.

I don't have access to the staging .env, but if it is a complete URI, it should work as per 

> Hey @miguelpduarte, just tested this and pushed the solution for the db_uri problem:
Basically, db_uri will be set (even if not provided), however, for example on the test scenario, if db_name is overriden, it will use that (i.e. if uri is mongodb://db_name_1, and the .env.test sets db name to db_name_2, the connection will use db_name_2 even though the uri already has another db name in it)

miguelpduarte:
The thing is that the URI has the specific requirement of being a cluster URI and having to specify the dbname separately, as mentioned elsewhere in this PR (can't really search it atm). We cannot use the URI to specify the dbname AFAIK, as when we tried that with the MongoDB Atlas deployment it no longer worked.

imnotteixeira:
> The thing is that the URI has the specific requirement of being a cluster URI and having to specify the dbname separately, as mentioned elsewhere in this PR (can't really search it atm). We cannot use the URI to specify the dbname AFAIK, as when we tried that with the MongoDB Atlas deployment it no longer worked.

Just tested this and basically, it first looks at the URI, and if present it uses it. Additionally, if there is a `DB_NAME` present it will override the one in the URI.

Finally, if no URI is present, DB_HOST and DB_PASS are used, together with the DB_NAME, for the tests use case, for example, where localhost is used.

Tl;dr, it works.

src/config/env.js
```diff
@@ -1,22 +1,26 @@
 require("dotenv-flow").config();
 
+const generateDBUriFromProps = () => {
```
miguelpduarte, line 3:
```suggestion
const generateDBUriFromEnv = () => {
```
Too much React? 😛

src/loaders/express.js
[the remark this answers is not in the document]
miguelpduarte: Resolving this for now though. Maybe we should create an issue to further analyze and update this later, though.

src/config/env.js
imnotteixeira: No React is too much React.

miguelpduarte (approved):
👍

PR #60, opened 2019-11-28
Update Account Schema

### LINKED ISSUE #52: Update Account Schema
In order to support Company and Admin accounts, the Account schema should be updated to contain the following fields:
- `email`
- `password`
- `isAdmin` : Boolean
- `company` : Ref to Company Model

Additionally, if the Account is admin (`isAdmin === true`), `company` must be null. If `company` is non-null, `isAdmin` must be false.

Furthermore, the following middlewares must be created:
- `isAdmin` - the logged in user is an admin
- `belongsToCompany(companyId)` - the logged in user is a representative of the company

GuilhermeJSilva:
Closes #52 

* Added `isAdmin` and `company` fields to Account Schema
* Force only one of the fields `isAdmin` and `company` to be required
* Rename Admin Token to God Token
* Make register endpoint force the created user to be an admin
* Only the god user can create admins

### DIFF
--- a/src/api/middleware/validators/auth.js
+++ b/src/api/middleware/validators/auth.js
@@ -3,19 +3,20 @@ const { body } = require("express-validator");
 const { useExpressValidators } = require("../errorHandler");
 const Account = require("../../../models/Account");
 
-const checkDuplicateUsername = async (username) => {
-    const acc = await Account.findOne({ username }).exec();
+const checkDuplicateEmail = async (email) => {
+    const acc = await Account.findOne({ email }).exec();
     if (acc) {
-        throw new Error("Username already exists");
+        throw new Error("Email already exists");
     }
 };
 
 const register = useExpressValidators([
-    body("username", "Invalid username")
-        .exists().withMessage("Username is required").bail()
-        .isString().withMessage("Username must be a String")
+    body("email", "Invalid email")
+        .exists().withMessage("Email is required").bail()
+        .isString().withMessage("Email must be a String")
+        .isEmail().withMessage("Email must be valid")
         .bail()
-        .custom(checkDuplicateUsername)
+        .custom(checkDuplicateEmail)
         .trim(),
 
     body("password", "Invalid password")
--- a/src/api/middleware/validators/auth.js
+++ b/src/api/middleware/validators/auth.js
@@ -13,8 +13,7 @@ const checkDuplicatedEmail = async (email) => {
 const register = useExpressValidators([
     body("email", "Invalid email")
         .exists().withMessage("Email is required").bail()
-        .isString().withMessage("Email must be a String")
-        .isEmail().withMessage("Email must be valid")
+        .normalizeEmail().isEmail().withMessage("Email must be valid")
         .bail()
         .custom(checkDuplicatedEmail)
         .trim(),
@@ -26,4 +25,15 @@ const register = useExpressValidators([
         .matches(/\d/).withMessage("Password must contain a number"),
 ]);
 
-module.exports = { register };
+const login =  useExpressValidators([
+    body("email", "Invalid email")
+        .exists().withMessage("Email is required").bail()
+        .normalizeEmail().isEmail().withMessage("Email must be valid")
+        .trim(),
+
+    body("password", "Invalid password")
+        .exists().withMessage("Password is required").bail()
+        .isString().withMessage("Password must be a String"),
+]);
+
+module.exports = { register, login };
--- a/src/api/middleware/validators/auth.js
+++ b/src/api/middleware/validators/auth.js
@@ -3,7 +3,7 @@ const { body } = require("express-validator");
 const { useExpressValidators } = require("../errorHandler");
 const Account = require("../../../models/Account");
 
-const checkDuplicateEmail = async (email) => {
+const checkDuplicatedEmail = async (email) => {
     const acc = await Account.findOne({ email }).exec();
     if (acc) {
         throw new Error("Email already exists");
@@ -16,7 +16,7 @@ const register = useExpressValidators([
         .isString().withMessage("Email must be a String")
         .isEmail().withMessage("Email must be valid")
         .bail()
-        .custom(checkDuplicateEmail)
+        .custom(checkDuplicatedEmail)
         .trim(),
 
     body("password", "Invalid password")
--- a/src/api/middleware/validators/auth.js
+++ b/src/api/middleware/validators/auth.js
@@ -7,7 +7,7 @@ const Account = require("../../../models/Account");
 const checkDuplicatedEmail = async (email) => {
     const acc = await Account.findOne({ email }).exec();
     if (acc) {
-        throw new Error("Email already exists");
+        throw new Error(ValidationReasons.ALREADY_EXISTS("email"));
     }
 };
 
--- a/src/api/middleware/validators/auth.js
+++ b/src/api/middleware/validators/auth.js
@@ -1,6 +1,7 @@
 const { body } = require("express-validator");
 
 const { useExpressValidators } = require("../errorHandler");
+const ValidationReasons = require("./validationReasons");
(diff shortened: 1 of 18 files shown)


src/api/middleware/validators/auth.js
```diff
@@ -3,19 +3,20 @@ const { body } = require("express-validator");
 const { useExpressValidators } = require("../errorHandler");
 const Account = require("../../../models/Account");
 
-const checkDuplicateUsername = async (username) => {
-    const acc = await Account.findOne({ username }).exec();
+const checkDuplicateEmail = async (email) => {
```
imnotteixeira:
```suggestion
const checkDuplicatedEmail = async (email) => {
```

src/api/middleware/validators/auth.js
```diff
@@ -3,19 +3,20 @@ const { body } = require("express-validator");
...
-    body("username", "Invalid username")
-        .exists().withMessage("Username is required").bail()
-        .isString().withMessage("Username must be a String")
+    body("email", "Invalid email")
+        .exists().withMessage("Email is required").bail()
+        .isString().withMessage("Email must be a String")
+        .isEmail().withMessage("Email must be valid")
         .bail()
-        .custom(checkDuplicateUsername)
+        .custom(checkDuplicateEmail)
```
imnotteixeira:
```suggestion
        .custom(checkDuplicatedEmail)
```

src/api/middleware/validators/auth.js
```diff
@@ -3,19 +3,20 @@ const { body } = require("express-validator");
...
     }
 };
 
 const register = useExpressValidators([
-    body("username", "Invalid username")
-        .exists().withMessage("Username is required").bail()
-        .isString().withMessage("Username must be a String")
+    body("email", "Invalid email")
+        .exists().withMessage("Email is required").bail()
+        .isString().withMessage("Email must be a String")
```
imnotteixeira: Do we need to check "is String" when we are checking "is Email"? It's kind of redundant

src/config/passport.js
```diff
@@ -4,21 +4,24 @@ const LocalStrategy = require("passport-local").Strategy;
...
+    usernameField: "email",
+    passwordField: "password",
+},
+(email, password, done) => {
+    Account.findOne({ email: email }, (err, user) => {
+        if (err) {
+            return done(err);
+        }
+        if (!user) {
+            return done(null, false, { message: "Incorrect email." });
```
imnotteixeira: Shouldn't this return a message like "Incorrect email or password."? I understand that in this scope, it makes sense that the errors are separated, but I believe this message is being served as an error, therefore it should be generalized

src/config/passport.js
```diff
@@ -4,21 +4,24 @@ const LocalStrategy = require("passport-local").Strategy;
...
+        }
+        if (!user.validatePassword(password)) {
+            return done(null, false, { message: "Incorrect password." });
```
imnotteixeira: Same as above ("Incorrect email or password.")

src/api/middleware/validators/auth.js
```diff
@@ -3,19 +3,19 @@ const { body } = require("express-validator");
...
     }
 };
 
 const register = useExpressValidators([
-    body("username", "Invalid username")
-        .exists().withMessage("Username is required").bail()
-        .isString().withMessage("Username must be a String")
+    body("email", "Invalid email")
+        .exists().withMessage("Email is required").bail()
+        .normalizeEmail().isEmail().withMessage("Email must be valid")
```
miguelpduarte: This should be using the validator reasons like in the other endpoint validators. The messages don't matter as their content will be decided in the Front-End. As such, we are using enums in the validators.

src/api/middleware/validators/auth.js
```diff
@@ -3,19 +3,19 @@ const { body } = require("express-validator");
 const { useExpressValidators } = require("../errorHandler");
 const Account = require("../../../models/Account");
 
-const checkDuplicateUsername = async (username) => {
-    const acc = await Account.findOne({ username }).exec();
+const checkDuplicatedEmail = async (email) => {
+    const acc = await Account.findOne({ email }).exec();
     if (acc) {
-        throw new Error("Username already exists");
+        throw new Error("Email already exists");
```
miguelpduarte: I'm not entirely sure what message is being sent in other custom validators, but we should probably also use the validator reasons enum here (and elsewhere).

src/api/middleware/validators/auth.js
```diff
@@ -25,4 +25,15 @@ const register = useExpressValidators([
...
-module.exports = { register };
+const login =  useExpressValidators([
+    body("email", "Invalid email")
+        .exists().withMessage("Email is required").bail()
+        .normalizeEmail().isEmail().withMessage("Email must be valid")
+        .trim(),
+
+    body("password", "Invalid password")
+        .exists().withMessage("Password is required").bail()
+        .isString().withMessage("Password must be a String"),
```
miguelpduarte: The messages should be changed, see above.

src/config/passport.js
```diff
@@ -4,21 +4,21 @@ const LocalStrategy = require("passport-local").Strategy;
...
-            }
-            return done(null, user);
-        });
-    }
+passport.use(new LocalStrategy({
+    usernameField: "email",
+    passwordField: "password",
+},
+(email, password, done) => {
+    Account.findOne({ email: email }, (err, user) => {
```
miguelpduarte:
```suggestion
    Account.findOne({ email }, (err, user) => {
```

src/models/Account.js
```diff
@@ -3,8 +3,39 @@ const bcrypt = require("bcrypt");
...
+    isAdmin: {
+        type: Boolean,
+        required: function() {
+            return !this.companyRef;
+        },
+        validate: {
+            validator: function(isAdmin) {
+                return isAdmin && !this.company;
+            },
+            message: "A user can not be an admin and a company representative",
```
miguelpduarte:
```suggestion
            message: "A user cannot be an admin and a company representative at the same time",
```

src/models/Account.js
```diff
@@ -3,8 +3,39 @@ const bcrypt = require("bcrypt");
...
+    company: {
+        type: Schema.Types.ObjectId, ref: "Company",
+        required: function() {
+            return !this.isAdmin;
+        },
+        validate: {
+            validator: function(companyRef) {
+                return !!companyRef && !this.isAdmin;
+            },
+            message: "A user can not be an admin and a company representative",
```
miguelpduarte: Same as above

test/end-to-end/auth.js
```diff
@@ -1,27 +1,33 @@
...
 
                 expect(res.status).toBe(422);
                 expect(res.body).toHaveProperty("error_code", ErrorTypes.VALIDATION_ERROR);
                 expect(res.body).toHaveProperty("errors");
                 expect(res.body.errors).toContainEqual({
                     "location": "body",
-                    "msg": "Username is required",
-                    "param": "username",
+                    "msg": "Email is required",
+                    "param": "email",
```
miguelpduarte: This (and all other tests that can do so) should be using the abstractions for testing validators that were introduced in the previous PR.

test/end-to-end/auth.js
```diff
@@ -1,27 +1,33 @@
 const { ErrorTypes } = require("../../src/api/middleware/errorHandler");
 const Account = require("../../src/models/Account");
 
+const test_god_token  = "testing_is_cool73";
```
miguelpduarte: Should be using `withGodToken`

GuilhermeJSilva:
>     * Fix custom validator for duplicate email to use a ValidatorReason
> 
>     * Add test for duplicate email responding with validation error
> 
> 
> Other than that, incredible! Good job!

The test for duplicated email was already done, but I added the reason to the ValidationReasons object

PR #61, opened 2020-01-22
Fix/date validators

### LINKED ISSUE #50: Fix validation (Dates)
When passing a past end date, and no publish date, it takes the default publish date which is `now`, however, since end date is in the past, it crashes without failing the validation. It should check the endDate vs publishDate on validation.

Another approach would be to prevent the creation of offers with an end date in the past.

Just to clarify, the model has a jobStartDate, and jobDuration, that refer to the job itself, however publish date and end date refer to the job posting, maybe we should change the endDate naming to make it clearer (maybe publishEndDate?)

miguelpduarte:
Closes #50

### DIFF
--- a/src/api/middleware/validators/offer.js
+++ b/src/api/middleware/validators/offer.js
@@ -22,7 +22,8 @@ const create = useExpressValidators([
 
     body("endDate", ValidationReasons.DEFAULT)
         .exists().withMessage(ValidationReasons.REQUIRED).bail()
-        .isISO8601({ strict: true }).withMessage(ValidationReasons.DATE)
+        .isISO8601({ strict: true }).withMessage(ValidationReasons.DATE).bail()
+        .isAfter().withMessage(ValidationReasons.DATE_EXPIRED)
         .toDate(),
 
     body("jobMinDuration", ValidationReasons.DEFAULT)
--- a/src/api/middleware/validators/offer.js
+++ b/src/api/middleware/validators/offer.js
@@ -20,7 +20,7 @@ const create = useExpressValidators([
         .isISO8601({ strict: true }).withMessage(ValidationReasons.DATE)
         .toDate(),
 
-    body("endDate", ValidationReasons.DEFAULT)
+    body("publishEndDate", ValidationReasons.DEFAULT)
         .exists().withMessage(ValidationReasons.REQUIRED).bail()
         .isISO8601({ strict: true }).withMessage(ValidationReasons.DATE).bail()
         .isAfter().withMessage(ValidationReasons.DATE_EXPIRED)
--- a/src/api/middleware/validators/offer.js
+++ b/src/api/middleware/validators/offer.js
@@ -17,13 +17,28 @@ const create = useExpressValidators([
 
     body("publishDate", ValidationReasons.DEFAULT)
         .optional()
-        .isISO8601({ strict: true }).withMessage(ValidationReasons.DATE)
+        .isISO8601({ strict: true }).withMessage(ValidationReasons.DATE).bail()
         .toDate(),
 
     body("publishEndDate", ValidationReasons.DEFAULT)
         .exists().withMessage(ValidationReasons.REQUIRED).bail()
         .isISO8601({ strict: true }).withMessage(ValidationReasons.DATE).bail()
-        .isAfter().withMessage(ValidationReasons.DATE_EXPIRED)
+        .isAfter().withMessage(ValidationReasons.DATE_EXPIRED).bail()
+        .custom((publishEndDateCandidate, { req }) => {
+            const { publishDate: publishDateRaw } = req.body;
+            // Default values and also handling if it is string or date object
+            const publishDate =
+                (publishDateRaw instanceof Date ? publishDateRaw.toISOString() : publishDateRaw)
+                || (new Date(Date.now())).toISOString();
+
+            if (publishEndDateCandidate <= publishDate) {
+                // end date is earlier than publish date, error!
+                throw new Error(ValidationReasons.MUST_BE_AFTER("publishDate"));
+            }
+
+            // Returning truthy value to indicate no error ocurred
+            return true;
+        })
         .toDate(),
 
     body("jobMinDuration", ValidationReasons.DEFAULT)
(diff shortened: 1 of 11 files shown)


miguelpduarte:
@imnotteixeira As mentioned in person (and just verified) the data is correctly inserted when the `publishEndDate` is larger that `publishDate` and they are both in the future. However (also like predicted) when they are swapped (but still both in the future) there is a mongoose validation error thrown in the server and an empty 500 response is returned.

Adding validation for this now.

miguelpduarte:
@imnotteixeira Should be good to go now. Re-requesting review. Sorry for the delay, it was caused by the surgery recovery mostly :sweat_smile:

imnotteixeira:
@miguelpduarte Didn't know there was spaghetti code fixing surgery 😈😈

src/api/middleware/validators/offer.js
```diff
@@ -17,13 +17,28 @@ const create = useExpressValidators([
...
+                || (new Date(Date.now())).toISOString();
+
+            if (value <= checkPublishDate) {
+                // end date is earlier than publish date, error!
+                throw new Error(ValidationReasons.MUST_BE_AFTER("publishDate"));
+            }
+
+            // Returning truthy value to indicate no error ocurred
+            return true;
+        })
```
imnotteixeira: Looks good! However, I would suggest some var names renaming:

* `value `-> `publishEndDateCandidate`
* `publishDate` -> `publishDateRaw`
* `checkPublishDate` -> `publishDate`

In this case the if would be way clearer: `if (publishEndDate <= publishDate)`

test/end-to-end/offer.js
```diff
@@ -150,7 +151,7 @@ describe("Offer endpoint tests", () => {
             test("Should successfully create an Offer", async () => {
                 const offer = {
                     title: "Test Offer",
-                    publishDate: "2019-11-17T00:00:00.000Z",
+                    publishDate: new Date(Date.now() - (24 * 3600 * 1000)),
```
imnotteixeira: That magic constant is being used twice here and in other files as well, can you create a global `const DAY_TO_MS` with that value?

test/utils/ValidatorTester.js
```diff
@@ -82,6 +82,27 @@ const ValidatorTester = (requestEndpoint) => (location) => (field_name) => ({
         });
     },
 
+    mustBeAfter: (field_name2) => {
+        test(`should be after ${field_name2}`, async () => {
+            const params = {
+                [field_name]: new Date(Date.now()).toISOString(),
+                [field_name2]: new Date(Date.now() + (24 * 3600 * 1000)).toISOString(),
```
imnotteixeira: `DAY_TO_MS` usage

src/api/middleware/validators/offer.js
miguelpduarte: Done. Though of other options but these ended up being the ones I agreed the most with.

test/end-to-end/offer.js
miguelpduarte: done

test/utils/ValidatorTester.js
miguelpduarte: also done ^^

test/utils/ValidatorTester.js
```diff
@@ -63,6 +64,46 @@ const ValidatorTester = (requestEndpoint) => (location) => (field_name) => ({
...
     },
 
+    mustBeFuture: () => {
+        test("should be a future Date", async () => {
+            const params = {
+                [field_name]: "1998-12-25",
+            };
+            const res = await requestEndpoint(params);
+
+            expect(res.status).toBe(422);
```
GuilhermeJSilva: Update to use the enum `HTTPStatus` from `http-status-codes`

test/utils/ValidatorTester.js
```diff
@@ -63,6 +64,46 @@ const ValidatorTester = (requestEndpoint) => (location) => (field_name) => ({
...
+    mustBeAfter: (field_name2) => {
+        test(`should be after ${field_name2}`, async () => {
+            const params = {
+                [field_name]: new Date(Date.now()).toISOString(),
+                [field_name2]: new Date(Date.now() + (DAY_TO_MS)).toISOString(),
+            };
+
+            const res = await requestEndpoint(params);
+
+            expect(res.status).toBe(422);
```
GuilhermeJSilva: Same as comment above

test/utils/ValidatorTester.js
miguelpduarte: A rebase was necessary, as this was actually done in #63. However, took the chance and extracted some of the common assertions in `ValidatorTester.js` to a separate function, which makes the code cleaner imo :slightly_smiling_face:

miguelpduarte: Please check it out now

test/utils/ValidatorTester.js
```diff
@@ -1,8 +1,19 @@
...
+
 const { ErrorTypes } = require("../../src/api/middleware/errorHandler");
 const ValidationReasons = require("../../src/api/middleware/validators/validationReasons");
 const { DAY_TO_MS } = require("../utils/TimeConstants");
 
+/**
+ * Checks the common parts of the error response: the status code, error_code property and that the body also has the error properties
+ * @param {*} res The response object
+ */
+const checkCommonErrorResponse = (res) => {
```
imnotteixeira: Shouldn't this be `checkValidationErrorResponse`? The first thought I had when reading was that this would check for 500 errors, as they are the common/generic. Since this checks for 422 (specific request validation error), the function name should be clearer

miguelpduarte: As the doc comment mentions, this "Checks the common parts of the error response". This method was created to reduce the code duplication since there were some parts that were repeated along this file.

Since this does not check the full response, and only the parts that are common in the response between all of the validator tests, this name was chosen. As such, I see no issue with it, and think that `checkValidationErrorResponse` would be misleading as it would seem like it would check for all of the response, but only performs checks on a small part of it, since the separate tests below perform the remaining checks.

miguelpduarte: Furthermore, since this function is only used in this file, has a doc comment that explains it, and is not even exported from the file, I think this is pointless bikeshedding :sweat_smile:

miguelpduarte: I guess it could be renamed to `checkCommonValidationErrorResponse` but that is just a mouthful, and sincerely irrelevant given the reasons above.

imnotteixeira: Sure, since this is a single-file thing (not exported), used to reduce repetition and the file is the validation one, the renaming is not as important, let's ship this

PR #64, opened 2020-01-26
Fix Account Schema issues and ensure Admin account exists on start-up

### LINKED ISSUE #53: Ensure Admin account exists on start-up
(This is probably not the ideal way to do this, but it is a good compromise in order to get things rolling quickly)

Requires #52 

When loading the connection to the `mongo` service, we should check if an Admin account already exists. If not, we should create it based on environment values (`ADMIN_EMAIL` and `ADMIN_PASSWORD`).

miguelpduarte:
Closes #53 and addresses some spaghetti left over from #52

### DIFF
--- a/src/loaders/mongoose.js
+++ b/src/loaders/mongoose.js
@@ -1,5 +1,6 @@
 const mongoose = require("mongoose");
 const config = require("../config/env");
+const Account = require("../models/Account");
 
 const setupDbConnection = async () => {
     if (!(config.db_uri || (config.db_host && config.db_port))) {
@@ -24,6 +25,7 @@ const setupDbConnection = async () => {
 
     try {
         await mongoose.connect(connection_uri, options);
+        await createDefaultAdmin();
     } catch (err) {
         console.error("Mongoose: failure in initial connection to the DB (aborting, will not retry)", err);
         process.exit(44);
@@ -32,6 +34,26 @@ const setupDbConnection = async () => {
     mongoose.connection.on("error", (err) => {
         console.error("Mongoose connection error:", err);
     });
+};
+
+const createDefaultAdmin = async () => {
+    if (process.env.NODE_ENV === "test") {
+        return;
+    }
+
+    const acc = await Account.findOne({ email: config.admin_email });
+    if (acc) {
+        // User with admin email already exists
+        return;
+    }
+
+    await Account.create({
+        email: config.admin_email,
+        password: config.admin_password,
+        isAdmin: true,
+    });
+
+    console.info("Created default admin!");
 };
 
 module.exports = setupDbConnection;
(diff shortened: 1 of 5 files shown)


src/loaders/mongoose.js
```diff
@@ -34,4 +36,24 @@ const setupDbConnection = async () => {
...
+const createDefaultAdmin = async () => {
+    if (process.env.NODE_ENV === "test") {
+        return;
+    }
+
+    const acc = await Account.findOne({ email: config.admin_email });
+    if (acc) {
+        // User with admin email already exists
+        return;
+    }
```
miguelpduarte: Should we instead check for an (any) admin user already existing?

imnotteixeira: I don't think so, we want THAT admin to exist, we don't want to find via trial and error which admin existed before

src/models/Account.js
```diff
@@ -8,17 +8,15 @@ const AccountSchema = new Schema({
...
         type: Boolean,
-        required: function() {
-            return !this.companyRef;
-        },
+        default: false,
         validate: {
             validator: function(isAdmin) {
-                return isAdmin && !this.company;
+                return isAdmin !== !!this.company;
             },
```
imnotteixeira: Can we swap to arrow function and make this a one-liner?

```js
validator: isAdmin => isAdmin !== !!this.company;
```

src/models/Account.js
```diff
@@ -29,10 +27,10 @@ const AccountSchema = new Schema({
             return !this.isAdmin;
         },
         validate: {
-            validator: function(companyRef) {
-                return !!companyRef && !this.isAdmin;
+            validator: function(company) {
+                return !!company !== this.isAdmin;
             },
```
imnotteixeira: Can we swap to arrow function and make this a one-liner?

```js
validator: company => !!company !== this.isAdmin;
```

src/models/Account.js
miguelpduarte: Arrow functions do not capture this, hence why the validators use the older `function` syntax :sweat_smile:

src/models/Account.js
miguelpduarte: See the other comment

src/loaders/mongoose.js
miguelpduarte: Ok, it was just checking for an admin with the given email vs checking for at least one admin being in the system. But this is fine for me as well. Resolving this thread then :slightly_smiling_face:

src/models/Account.js
imnotteixeira: Correct.. you would have to `.bind(this)`... nevermind

imnotteixeira:
Merge this when you think best, don't know if it should go now or after #61

miguelpduarte:
@imnotteixeira I'm waiting for #61 to merge first since it has more tests that might check for potential problems with conflicts here. But will merge it shortly after it's done.

miguelpduarte:
Fixed conflicts, please check this and merge it if it's okay, I'll be back later if this still needs work, but I think it's probably done :slightly_smiling_face:

Issue #54, opened 2019-11-26
Company Applications and Approvals


miguelpduarte:
- [x] Create CompanyApplication model: Has attributes of Account + Company (for creating them later on if the Application is accepted) + `approvalDate` + `submitDate` + `rejectDate` (the dates  - except approval - are null until changed to represent that that event did not yet happen)
- [x] List CompanyApplications endpoint (admin only)
- [x] CompanyApplication approval endpoint (admin only) -> Creates the Company and, in case of success, the Account for the Company representative
- [x] CompanyApplication rejection endpoint (admin only)

GuilhermeJSilva:
As discussed in person with @imnotteixeira, each `CompanyApplication` should have :
- `motivation` text field
- `rejectReaseon` text field (optional only when the application is pending or accepted)
- link to an Account (add flag for pending verification accounts to the AccountSchema)
- `comunicationEmail`, defaults to Account email
- link to a Company

A Company should have a flag `active` (has been approved).
An Account should have a flag `pending`(is pending approval and can't log in)

On a rejected application the Account is deleted.

GuilhermeJSilva:
@miguelpduarte Any thoughts?

imnotteixeira:
Additionally, it would be really helpful to have a virtual field `Status` which, based on the approvalDate + submitDate + rejectDate, returns a value like `APPROVED`, `PENDING` or `REJECTED`, which is clearer when filtering

miguelpduarte:
I think that this was already discussed, wasn't this in a photo of the whiteboard somewhere?

Don't feel like we should delete the applications.

Furthermore, wasn't an application supposed to have all the fields for the company? Or was that just for the account? Don't recall which one but this was definitely the case for one of them.

miguelpduarte:
Due to us working with mongo, I think it makes more sense to simply have all the data as a sort of blob in the companyapplication and then later on create the company and associated rep's account.

Since references in mongo are weird, don't feel like the suggestion of links to things is the way to go.

miguelpduarte:
Basically my opinion is to defualt to what was already decided in the issue description... The creation of a company and account and then tagging them with a disabled attribute just seems like unnecessary confusion to me.

Only creating them after approval makes more sense imo.

As such, there would be no trouble only listing the active companies or not allowing login to some accounts. These two models would retain their semantic value with no problem whatsoever.

[...]

PR #66, opened 2020-01-30
Add company application schema

GuilhermeJSilva:
Contributes to #54 

- Added company application schema
- Added tests to the schema
- Added test function to simplify tests to mongoose schemas
- Fixed bug in auth, where the program was not waiting for the validate password promise

### DIFF
--- a/src/models/CompanyApplication.js
+++ b/src/models/CompanyApplication.js
@@ -23,15 +23,27 @@ const CompanyApplicationSchema = new Schema({
     approvedAt: {
         type: Date,
         validate: [
-            validateDecisionDate,
-            "`approvedAt` must be after `submittedAt`",
+            {
+                validator: validateDecisionDate,
+                msg: "`approvedAt` must be after `submittedAt`",
+            },
+            {
+                validator: validateApprovedDate,
+                msg: "`approvedAt` and `rejectedAt` are mutually exclusive",
+            },
         ],
     },
     rejectedAt: {
         type: Date,
         validate: [
-            validateDecisionDate,
-            "`rejectedAt` must be after `submittedAt`",
+            {
+                validator: validateDecisionDate,
+                msg: "`rejectedAt` must be after `submittedAt`",
+            },
+            {
+                validator: validateRejectedDate,
+                msg: "`approvedAt` and `rejectedAt` are mutually exclusive",
+            },
         ],
     },
     rejectReason: {
@@ -46,6 +58,15 @@ const CompanyApplicationSchema = new Schema({
 
 function validateDecisionDate(value) {
     return !value || (value > this.submittedAt);
+}
+
+function validateApprovedDate(value) {
+
+    return !value || !this.rejectedAt;
+}
+
+function validateRejectedDate(value) {
+    return !value || !this.approvedAt;
 }
 
 const CompanyApplication = mongoose.model("CompanyApplication", CompanyApplicationSchema);
--- a/src/models/CompanyApplication.js
+++ b/src/models/CompanyApplication.js
@@ -1,7 +1,7 @@
 const mongoose = require("mongoose");
 const { Schema } = mongoose;
 
-const CompanyApplicaionSchema = new Schema({
+const CompanyApplicationSchema = new Schema({
     email: {
         type: String,
         trim: true,
@@ -16,7 +16,7 @@ const CompanyApplicaionSchema = new Schema({
         minlength: 10,
         required: true,
     },
-    submitedAt: {
+    submittedAt: {
         type: Date,
         required: true,
     },
@@ -24,19 +24,20 @@ const CompanyApplicaionSchema = new Schema({
         type: Date,
         validate: [
             validateDecisionDate,
-            "`approvalDate` must be after `submitDate`",
+            "`approvedAt` must be after `submittedAt`",
         ],
     },
     rejectedAt: {
         type: Date,
         validate: [
             validateDecisionDate,
-            "`rejectedDate` must be after `submitDate`",
+            "`rejectedAt` must be after `submittedAt`",
         ],
     },
     rejectReason: {
         type: String,
         maxlength: 1500,
+        minlength: 10,
         required: function() {
             return !!this.rejectedAt;
         },
@@ -44,8 +45,8 @@ const CompanyApplicaionSchema = new Schema({
 });
 
 function validateDecisionDate(value) {
-    return !value || (value > this.submitDate);
+    return !value || (value > this.submittedAt);
 }
 
-const CompanyApplicaion = mongoose.model("CompanyApplicaion", CompanyApplicaionSchema);
-module.exports = CompanyApplicaion;
+const CompanyApplication = mongoose.model("CompanyApplication", CompanyApplicationSchema);
+module.exports = CompanyApplication;
--- a/src/models/CompanyApplication.js
+++ b/src/models/CompanyApplication.js
@@ -1,5 +1,6 @@
 const mongoose = require("mongoose");
 const { Schema } = mongoose;
+const ApplicationStatus = require("./ApplicationStatus");
 
 const CompanyApplicationSchema = new Schema({
     email: {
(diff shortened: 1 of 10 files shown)


test/company_application_schema.js
```diff
@@ -0,0 +1,26 @@
+const CompanyApplication = require("../src/models/CompanyApplication");
+const testSchema = require("./utils/SchemaTester");
```
miguelpduarte: Nitpick: rename variable to `SchemaTester`

test/utils/SchemaTester.js
```diff
@@ -0,0 +1,108 @@
+const testSchema = (Model) => ({
```
miguelpduarte: Nitpick: rename to `SchemaTester`

test/utils/SchemaTester.js
```diff
@@ -0,0 +1,108 @@
...
+    fieldRequired: (fieldName, customObject) => {
+        test(`'${fieldName}' is required`, async () => {
+            const emptyObject = new Model(customObject || {});
```
miguelpduarte: Given that it might not be empty, it makes more sense for `emptyObject` to be renamed to `modelInstance` imo

test/utils/SchemaTester.js
```diff
@@ -0,0 +1,108 @@
+const testSchema = (Model) => ({
+    fieldRequired: (fieldName, customObject) => {
```
miguelpduarte:
```suggestion
    fieldRequired: (fieldName, customObject = {}) => {
```

Instead of using `|| {}` below, I think default argument values are clearer

Furthermore, `customObject` can be renamed to `instanceValues` or something similar, I think. I'm not sure of the best name for this variable, but I thin that this is clearer than `customObject`.

test/utils/SchemaTester.js
```diff
@@ -0,0 +1,108 @@
...
+                expect(err.errors[fieldName]).toBeDefined();
+                expect(err.errors[fieldName]).toHaveProperty("kind", "required");
+                expect(err.errors[fieldName]).toHaveProperty("message", `Path \`${fieldName}\` is required.`);
+            }
+        });
+    },
+
+    minLength: (fieldName, minimum) => {
+        describe(`${fieldName} must have a minimum of ${minimum} characters`, () => {
+            test("check for error", async () => {
```
miguelpduarte:
```suggestion
            test("less than minimum bound should throw error", async () => {
```

Clearer name, I think

test/utils/SchemaTester.js
```diff
@@ -0,0 +1,108 @@
...
+                    await model.validate();
+                } catch (err) {
+                    expect(err.errors[fieldName]).toBeDefined();
+                    expect(err.errors[fieldName]).toHaveProperty("kind", "minlength");
+                    expect(err.errors[fieldName]).toHaveProperty("message",
+                        `Path \`${fieldName}\` (\`${"a".repeat(minimum - 1)}\`) is shorter than the minimum allowed length (${minimum}).`
+                    );
+                }
+            });
+            test("check for success", async () => {
```
miguelpduarte:
```suggestion
            test("at minimum should not throw error", async () => {
```

I think this might be a clearer name? Not super sure about this one

test/utils/SchemaTester.js
```diff
@@ -0,0 +1,108 @@
...
+                const modelObj = {};
+                modelObj[fieldName] = "a".repeat(maximum);
+
+                const model = new Model(modelObj);
+                try {
+                    await model.validate();
+                } catch (err) {
+                    expect(err.errors[fieldName]).not.toBeDefined();
+                }
+            });
```
miguelpduarte: Please improve the test names, similar to what I've suggested above

src/models/CompanyApplication.js
```diff
@@ -0,0 +1,52 @@
...
+        required: function() {
+            return !!this.rejectedAt;
+        },
+    },
+});
+
+function validateDecisionDate(value) {
+    return !value || (value > this.submitedAt);
+}
+
```
imnotteixeira: Can you add the virtual field `status`?
Something like 
```js
if(!approvedAt && !rejectedAt) return ApplictaionStatus.PENDING
else if(approvedAt) return ApplictaionStatus.APPROVED
else return ApplictaionStatus.REJECTED
```

src/models/CompanyApplication.js
```diff
@@ -0,0 +1,52 @@
...
+        minlength: 10,
+        required: true,
+    },
+    submitedAt: {
+        type: Date,
+        required: true,
+    },
+    approvedAt: {
+        type: Date,
+        validate: [
```
imnotteixeira: I think we should have some type of validation against having approvedAt and rejectedAt set at the same time. They should be mutually exclusive.

test/utils/SchemaTester.js
```diff
@@ -0,0 +1,108 @@
...
+        });
+    },
+
+    dateAfterDate: (laterDate, earlierDate) => {
+        describe(`'${earlierDate}' must be after '${laterDate}'`, () => {
+            test("check for error", () => {
+                const modelObj = {};
+                modelObj[earlierDate] =  new Date("02/01/2020");
+                modelObj[laterDate] = new Date("01/01/2020");
+                const model = new Model(modelObj);
```
miguelpduarte:
```suggestion
                const model = new Model({
                    [earlierDate]: new Date("02/01/2020"),
                    [laterDate]: new Date("01/01/2020"),
                });
```

I think using [computed property names](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Computed_property_names) makes this clearer :slightly_smiling_face:

src/models/CompanyApplication.js
```diff
@@ -0,0 +1,52 @@
...
+            return !!this.rejectedAt;
+        },
+    },
+});
+
+function validateDecisionDate(value) {
+    return !value || (value > this.submitedAt);
+}
+
+const CompanyApplicaion = mongoose.model("CompanyApplicaion", CompanyApplicaionSchema);
```
imnotteixeira:
```suggestion
const CompanyApplication = mongoose.model("CompanyApplication", CompanyApplicationSchema);
```

test/utils/SchemaTester.js
```diff
@@ -0,0 +1,108 @@
...
+                    expect(err.errors[laterDate]).toHaveProperty("kind", "user defined");
+                    expect(err.errors[laterDate]).toHaveProperty("message", `\`${laterDate}\` must be after \`${earlierDate}\``);
+                });
+            });
+
+            test("check for success", () => {
+                const modelObj = {};
+                modelObj[earlierDate] =  new Date("01/01/2020");
+                modelObj[laterDate] = new Date("02/01/2020");
+                const model = new Model(modelObj);
```
miguelpduarte: Same suggestion as above, use computed property names :)

src/models/CompanyApplication.js
```diff
@@ -0,0 +1,52 @@
+const mongoose = require("mongoose");
+const { Schema } = mongoose;
+
+const CompanyApplicaionSchema = new Schema({
```
imnotteixeira:
```suggestion
const CompanyApplicationSchema = new Schema({
```

test/utils/SchemaTester.js
```diff
@@ -0,0 +1,108 @@
...
+
+    dateAfterDate: (laterDate, earlierDate) => {
+        describe(`'${earlierDate}' must be after '${laterDate}'`, () => {
+            test("check for error", () => {
+                const modelObj = {};
+                modelObj[earlierDate] =  new Date("02/01/2020");
+                modelObj[laterDate] = new Date("01/01/2020");
+                const model = new Model(modelObj);
+
+                return model.validate((err) => {
```
miguelpduarte: Please use the promise-based version of validate used above, since I think that this one doesn't always fail the tests since jest simply thinks there are no assertions in this test. Additionally, the version used above provides with better debug messages from what I've seen :slightly_smiling_face:

src/models/CompanyApplication.js
miguelpduarte: Agreed, good point

test/utils/SchemaTester.js
```diff
@@ -0,0 +1,150 @@
...
+                } catch (err) {
+                    expect(err.errors[laterDate]).not.toBeDefined();
+                }
+            });
+        });
+    },
+
+    mutuallyExclusive: (firstField, secondField, customObject) => {
+        describe(`${firstField} and ${secondField} must be mutually exclusive`, () => {
+            test(`if ${firstField} and ${secondField} are defined throws an error should `, async () => {
```
miguelpduarte:
```suggestion
            test(`if ${firstField} and ${secondField} are defined an error should be thrown`, async () => {
```

Typo here

PR #68, opened 2020-02-19
Add verification for applications with the same email

imnotteixeira:
* Moved constant defining files to 'constants' folder
* Added business rules error handler
* Added last resort error handler (returns generic message + HTTP 500)


src/api/routes/application.js
```diff
@@ -12,15 +11,15 @@ module.exports = (app) => {
...
+
         try {
             // This is safe since the service is destructuring the passed object and the fields have been validated
             const application = await (new ApplicationService()).create(req.body);
 
             return res.json(application);
         } catch (err) {
-            console.error(err);
-            return res.status(HTTPStatus.INTERNAL_SERVER_ERROR).send();
+            return next(err);
```
GuilhermeJSilva: Can you please substitute the catch in the offer endpoints (there are only a few) that are doing the error handling

src/services/application.js
```diff
@@ -1,11 +1,25 @@
...
+
+        const existingApplications = await CompanyApplication.find({ email });
+
+        if (existingApplications.some((application) =>
+            application.state === ApplicationStatus.PENDING ||
+            application.state === ApplicationStatus.APPROVED)
+        )  {
+            throw new NijobsBusinessRulesError(CompanyApplicationRules.ONLY_ONE_APPLICATION_ACTIVE_PER_EMAIL);
+        }
+
```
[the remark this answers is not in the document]
imnotteixeira: Done, although I could not add a `beforeEach(clean accounts and company applications)` on the company application tests that check for email duplication, because this would run in parallel with the auth tests and would delete the accounts that were expected to exist there

src/api/routes/application.js
```diff
@@ -12,15 +11,14 @@ module.exports = (app) => {
     /**
      * Creates a new Company Application
      */
-    router.post("/", validators.create, async (req, res) => {
+    router.post("/", validators.create, ...validators.businessRules, async (req, res, next) => {
```
GuilhermeJSilva: You're checking application uniqueness twice, once in the create validator and another in the business rules. I would think it would be better if it was kept only in the create validator since that rule is going to be specific to this endpoint.

src/models/CompanyApplication.js
```diff
@@ -64,18 +79,49 @@ CompanyApplicationSchema.virtual("state").get(function() {
...
 
-function validateRejectedDate(value) {
-    return !value || !this.approvedAt;
+const applicationUniqueness = async (email) => {
+    const existingApplications = await CompanyApplication.find({ email });
+    if (existingApplications.some((application) =>
+        application.state === ApplicationStatus.PENDING ||
+    application.state === ApplicationStatus.APPROVED)
+    ) {
+        throw new Error(CompanyApplicationRules.ONLY_ONE_APPLICATION_ACTIVE_PER_EMAIL.msg);
```
GuilhermeJSilva: In here you could simply return false and mongoose would use the correct message

src/api/middleware/errorHandler.js
```diff
@@ -1,13 +1,12 @@
 const HTTPStatus = require("http-status-codes");
 const { validationResult } = require("express-validator");
 
-const { errorExtractor } = require("../../lib/dbErrorExtractor");
-
 const ErrorTypes = Object.freeze({
     VALIDATION_ERROR: 1,
     // Possibly nested in the future
     DB_ERROR: 2,
```
GuilhermeJSilva: I think DB error is no longer in use can you remove it

PR #69, opened 2020-03-01
Feature/restrict offer creation

markmeehan99:
- Implemented a solution for limiting the creation of an offer by a given company.
- A company may only have, at a given time, a number of active offers, which is specified in MAX_ACTIVE_OFFERS_ALLOWED.

### DIFF
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -32,7 +32,7 @@ module.exports = (app) => {
     router.post("/", authMiddleware.isGod, validators.create, async (req, res) => {
         try {
             // This is safe since the service is destructuring the passed object and the fields have been validated
-            if (await Offer.find().activeOffersCount(req.body).length < OFFER_LIMIT_REACHED) {
+            if (await Offer.find().activeOffersCount(req.body).length < MAX_ACTIVE_OFFERS_ALLOWED) {
                 const offer = await (new OfferService()).create(req.body);
 
                 return res.json(offer);
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -32,7 +32,7 @@ module.exports = (app) => {
     router.post("/", authMiddleware.isGod, validators.create, async (req, res) => {
         try {
             // This is safe since the service is destructuring the passed object and the fields have been validated
-            if (await Offer.find().activeOffersCount(req.body).length < MAX_ACTIVE_OFFERS_ALLOWED) {
+            if (await Offer.find().activeOffersCount(req.body).length < OFFER_LIMIT_REACHED) {
                 const offer = await (new OfferService()).create(req.body);
 
                 return res.json(offer);
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -1,8 +1,10 @@
 const HTTPStatus = require("http-status-codes");
 const { Router } = require("express");
+const Offer = require("../../models/Offer");
 
 const authMiddleware = require("../middleware/auth");
 const validators = require("../middleware/validators/offer");
+const ValidationReasons = require("../middleware/validators/validationReasons");
 const OfferService = require("../../services/offer");
 
 const router = Router();
@@ -32,7 +34,7 @@ module.exports = (app) => {
             // This is safe since the service is destructuring the passed object and the fields have been validated
             if (await Offer.find().activeOffersCount(req.body).length < MAX_ACTIVE_OFFERS_ALLOWED) {
                 const offer = await (new OfferService()).create(req.body);
-                
+
                 return res.json(offer);
             } else throw new Error(ValidationReasons.OFFER_LIMIT_REACHED(req.body.owner));
 
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -30,12 +30,16 @@ module.exports = (app) => {
     router.post("/", authMiddleware.isGod, validators.create, async (req, res) => {
         try {
             // This is safe since the service is destructuring the passed object and the fields have been validated
-            const offer = await (new OfferService()).create(req.body);
+            if (await Offer.find().activeOffersCount(req.body).length < MAX_ACTIVE_OFFERS_ALLOWED) {
+                const offer = await (new OfferService()).create(req.body);
+                
+                return res.json(offer);
+            } else throw new Error(ValidationReasons.OFFER_LIMIT_REACHED(req.body.owner));
 
-            return res.json(offer);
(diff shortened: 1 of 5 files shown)


src/api/routes/offer.js
```diff
@@ -30,12 +32,16 @@ module.exports = (app) => {
     router.post("/", authMiddleware.isGod, validators.create, async (req, res) => {
         try {
             // This is safe since the service is destructuring the passed object and the fields have been validated
-            const offer = await (new OfferService()).create(req.body);
+            if (await Offer.find().activeOffersCount(req.body).length < Offer.MAX_ACTIVE_OFFERS_ALLOWED) {
```
imnotteixeira: This should be in the validators.create, as a custom validator, so that the error is automatically produced and standardized.

imnotteixeira: Also, you don't want to pass all the req.body, as it has a lot of properties, use req.body.owner instead, as it represents the company

src/api/routes/offer.js
```diff
@@ -30,12 +32,16 @@ module.exports = (app) => {
...
+                const offer = await (new OfferService()).create(req.body);
+
+                return res.json(offer);
+            } else throw new Error(ValidationReasons.OFFER_LIMIT_REACHED(req.body.owner));
```
imnotteixeira: By removing this if statement when creating the custom validator, do not forget to use this ValidatorReason as well

src/models/Offer.js
```diff
@@ -110,9 +110,28 @@ OfferSchema.query.current = function() {
     });
 };
 
+OfferSchema.query.activeOffersCount = function(canditateCompany) {
```
imnotteixeira: Rename this to `activeOffers`, as it returns the offers, not only a number (count)

src/models/Offer.js
```diff
@@ -110,9 +110,28 @@ OfferSchema.query.current = function() {
...
+        publishDate: {
+            $lte: new Date(Date.now()),
+        },
+        publishEndDate: {
+            $gt: new Date(Date.now()),
+        },
+        owner: {
+            $eq: canditateCompany.owner,
+        },
+    });
```
imnotteixeira: If you look closely, this has the same first two conditions as Offer.current query, thus you can chain them and have a simpler one here: 

```js
OfferSchema.query.activeOffers = function(owner) {
    return this.current.where({owner});
}
```

imnotteixeira: After looking at this, I noticed that the current query does not have the `if !isHidden`, should we add it? @miguelpduarte @GuilhermeJSilva? If we do, I'd change from `current` to `currentActive` or something like that, to make it clear that we are not only checking the dates

src/models/Offer.js
```diff
@@ -110,9 +110,28 @@ OfferSchema.query.current = function() {
...
+
 
 // Useful for testing correct field implementation
 // console.log("DBG: ", OfferSchema.path("location"));
 
-module.exports = Offer;
+module.exports = {
+    Offer,
+    MAX_ACTIVE_OFFERS_ALLOWED,
+};
```
imnotteixeira: This is probably a cause of problems.
Here, you are saying that you want to export an object which has two properties: `Offer`, and `MAX_ACTIVE_OFFERS_ALLOWED`. By not specifying the values for those keys, they will get the value of the variables with the same name, by default, which is ok. However, when you import it, you won't be importing the Offer model anymore, you will be importing an object which Offer key will be the Offer model, therefore you would actually need to do something like `const Offer = require("./Offer").Offer;` to import it, which is not that good.

What you can do instead is do it like this:
```js
module.exports = Offer;
module.exports.MAX_ACTIVE_OFFERS_ALLOWED = MAX_ACTIVE_OFFERS_ALLOWED;
```

With this, you can import the Offer model with `const Offer = require("./Offer");` and MAX_ACTIVE_OFFERS_ALLOWED with `const MAX_ACTIVE_OFFERS_ALLOWED = require("./Offer").MAX_ACTIVE_OFFERS_ALLOWED;`

**This was an explanation**, however, what I want you to do is use the constants file, in models/constants/Offer.js and create the maxActiveOffersAllowed rule there. This way, you only export the Offer model from this file

src/models/Offer.js
imnotteixeira: Also, this should be in the Company Model, I think, as you want the active offers of a specific Company, it doesn't make sense to query all offers to check which ones have the specified company as an owner, and then check if they are active. What you want to do is move this to the Company model, and then `Company.find({_id: req.body.owner}).activeOffers` in the validator

src/api/routes/offer.js
imnotteixeira: Another thing: instead of using `.length`, use `countDocuments()`, so that the query is more efficient, since you don't need the actual offers, only how many.

src/models/Offer.js
miguelpduarte: I do not recall what the idea behind `isHidden` was anymore... Do any of you remember? I'm thinking if we should scrap it altogether if it's not being used... Since this is mongo we can always just introduce it later anyway 😛

imnotteixeira: I am not sure, but when I was looking, I felt the need for a isActive, in case the company wants to hide/remove an offer, probably the isHidden is that field, that's why I suggested to use it in the `current` query

src/api/routes/offer.js
```diff
@@ -29,7 +29,6 @@ module.exports = (app) => {
      */
     router.post("/", authMiddleware.isGod, validators.create, async (req, res) => {
         try {
-            // This is safe since the service is destructuring the passed object and the fields have been validated
```
markmeehan99: This was a mistake, I'll add this line as it was

src/models/Offer.js
```diff
@@ -110,6 +110,14 @@ OfferSchema.query.current = function() {
     });
 };
 
+OfferSchema.query.activeOffers = function(canditateCompany) {
```
markmeehan99: As suggested, I renamed this query to `activeOffers`. However, this now returns a count, since I thought it made sense that countDoocuments was called in the query itself and not in the validator, since what we really want from the query is the count itself. Do you think it makes more sense this way, or should I change it?

imnotteixeira: I think this query should return the values, as it may be useful in other scenarios. In the validator you can add the `countDocuments`. Do not forget to move this to the Company model instead

imnotteixeira:
Closed due to being stale. Besides, the work is being done in #108

PR #71, opened 2020-03-30
Add mock management script and more enum schema values

brunodccarvalho:
The TechnologyType and FieldType enums only have a few rows, added early for testing purposes. We can extend these enums now that we have mock data. The new rows are based on popular StackOverflow tags, developer roles, and web developer roadmaps.

Add also a mock script to insert and clean the mongo database using the mock data.

Note: the data is intentionally left out for two reasons:
- so developers can use their own mock data
- the mock data would easily fall out of sync with changes in the offer and company schemas


brunodccarvalho:
> I believe the mock utility doesn't add anything to this project function, as it works as a helping hand to the developer. That being said, why not creating a node module, publishing it as niaefeup/nijobs-utils (so we can use for other utils as well if we see fit) and then we include it as a devDependency of this project?

Indeed, the script might as well be managed by developers too. Adding it to the repository is a bit noisy (just like the data and the scraper).

> The way I see it, this can be done with a few changes:
> 
>   * The util receives the mongoose connection, instead of setting one up
> 
>   * The populate receives the models instead of importing them (I would like for it to work for any type of mocking/populate for us to play in the future, but due to the way it is written now and due to the linking part, I would not worry with that for now)

If we need to pass the mongoose connection, what would the invocation of this script look like? Would it be an HTTP request to the server?

One way to incorporate developer hooks and scripts is to create a /dev endpoint that is only available when `NODE_ENV` is explicitly set to `"development"` and is populated by a 'dev.js' routes file maintained by the developer and not tracked by git. Under this route you could do any number of nasty things you need, like resetting the database and reading private tables.

This would do what you're describing with a curl POST request to e.g. /dev/populate, with the mock data in the payload. Is this what you have in mind?

For example:

```
  --- a/src/api/index.js
  +++ b/src/api/index.js
  @@ -7,9 +7,12 @@ const application = require("./routes/application");
   module.exports = () => {
       const app = Router();
       auth(app);
       example(app);
       offer(app);
       application(app);
  +    if (process.env.NODE_ENV === "development")
  +        require("./routes/dev")(app);
  +
       return app;
   };
```
@GuilhermeJSilva 

I don't see any way of transforming this into a dev dependency.

imnotteixeira:
@brunodccarvalho Ok, we are talking about two things. 

First is what I suggested: Creating a separate package registered in npm as niaefeup/nijobs-utils, which would contain, for now, the mocking util (populate, reset, etc..) which would be exported by the package.

Then, right here on this project, we can have that NON-PRODUCTION only route that imports the mock util from the nijobs-utils package (which should be installed as a dev-dependency), and calls the mock stuff, giving it the mongoose connection and the models, which can be imported from nijobs-be, because they are defined here.

One additional thing, instead of checking `if(NODE_ENV === "development")`, we can check   `if(NODE_ENV !== "production")`, as it might be useful for testing as well.

PR #74, opened 2020-07-18
Company Applications Review

imnotteixeira:
Addresses missing functionality on #54 

Added 
* GET `/applications/company/search` - list all applications, applying provided filters and with pagination
    * limit: number
    * offset: number
    * filters
    ```js
        companyName: String
        submissionDate: { - only one of the properties is necessary, but you can use both to define interval
            from: Date
            to: Date
        }
        state: String | Array - String for exact match, Array to provide set of options
    ```

* POST `/applications/company/:id/approve` (approves company application with id=:id)
* POST `/applications/company/:id/reject` (rejects company application with id=:id)

### DIFF
--- a/src/api/middleware/validators/validatorUtils.js
+++ b/src/api/middleware/validators/validatorUtils.js
@@ -16,6 +16,29 @@ const valuesInSet = (set) => (arr) => {
 };
 
 /**
+ * Returns a validator that checks whether all of the elements of an array belong to the provided set of values
+ * @param {Array} set
+ */
+const stringOrValuesInSet = (set) => (val) => {
+
+    const checkValid = (value) => (arr) => {
+        if (!arr.includes(value)) {
+            throw new Error(ValidationReasons.IN_ARRAY(set));
+        }
+    };
+
+    if (typeof val === "string") {
+        checkValid(val)(set);
+    } else {
+        for (const item of val) {
+            checkValid(item)(set);
+        }
+    }
+
+    return true;
+};
+
+/**
  * Throws an error if it already exists a account with the given email.
  * @param {String} email
  */
@@ -29,5 +52,6 @@ const checkDuplicatedEmail = async (email) => {
 
 module.exports = {
     valuesInSet,
+    stringOrValuesInSet,
     checkDuplicatedEmail,
 };
--- a/src/api/middleware/validators/validatorUtils.js
+++ b/src/api/middleware/validators/validatorUtils.js
@@ -21,17 +21,17 @@ const valuesInSet = (set) => (arr) => {
  */
 const stringOrValuesInSet = (set) => (val) => {
 
-    const checkValid = (value) => (arr) => {
+    const checkValid = (value, arr) => {
         if (!arr.includes(value)) {
             throw new Error(ValidationReasons.IN_ARRAY(set));
         }
     };
 
     if (typeof val === "string") {
-        checkValid(val)(set);
+        checkValid(val, set);
     } else {
         for (const item of val) {
-            checkValid(item)(set);
+            checkValid(item, set);
         }
     }
 
--- a/src/api/routes/review.js
+++ b/src/api/routes/review.js
@@ -4,6 +4,7 @@ const authMiddleware = require("../middleware/auth");
 const companyApplicationValidators = require("../middleware/validators/application");
 const ApplicationService = require("../../services/application");
 const mongoose = require("mongoose");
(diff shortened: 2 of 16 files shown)


miguelpduarte:
100% just style but wouldn't it look better to have `POST` being approve and `DELETE` being reject? I like varying my HTTP verbs 😛

imnotteixeira:
> 100% just style but wouldn't it look better to have `POST` being approve and `DELETE` being reject? I like varying my HTTP verbs 😛 

No. The reject endpoint does not delete the resource, therefore it should be POST

src/services/application.js
```diff
@@ -18,6 +38,160 @@ class CompanyApplicationService {
...
+                        $and: this.buildFiltersQuery(queryFilters),
+                    } : {})
+                    .sort(sortingOptions || { submittedAt: "desc" })
+                    .skip(offset)
+                    .limit(limit)
+                    .exec()
+                )
+                    .map(async (application) => ({
+                        ...application.toObject(),
+                        state: (await CompanyApplication.findById(application._id).exec()).state,
```
GuilhermeJSilva, line 142: The state can be determined by the other fields in the application, this will save a lot of unnecessary queries.

src/services/application.js
```diff
@@ -18,6 +38,160 @@ class CompanyApplicationService {
...
+                filter["$and"].push({ submittedAt: { "$lte": submissionDate.to } });
+            }
+            if (submissionDate.from) {
+                filter["$and"].push({ submittedAt: { "$gte": submissionDate.from } });
+            }
+            filterQueries.push(filter);
+        }
+
+        return filterQueries;
+    }
```
GuilhermeJSilva, line 101: For semantic porpuses, I would separate this function into two (one which returns a single filter). I think it would make the code more readable.

src/api/middleware/validators/validatorUtils.js
```diff
@@ -15,6 +15,29 @@ const valuesInSet = (set) => (arr) => {
...
+ * Returns a validator that checks whether all of the elements of an array belong to the provided set of values
+ * @param {Array} set
+ */
+const stringOrValuesInSet = (set) => (val) => {
+
+    const checkValid = (value) => (arr) => {
+        if (!arr.includes(value)) {
+            throw new Error(ValidationReasons.IN_ARRAY(set));
+        }
+    };
```
GuilhermeJSilva, line 28: This does not need to be a nested function, the return value of the outer function is never used more than once.

src/api/routes/review.js
```diff
@@ -0,0 +1,103 @@
...
+                        .json(buildErrorResponse(ErrorTypes.VALIDATION_ERROR, [err.message]));
+                } else if (err instanceof ApplicationService.CompanyApplicationAlreadyReiewed) {
+                    return res
+                        .status(HTTPStatus.CONFLICT)
+                        .json(buildErrorResponse(ErrorTypes.VALIDATION_ERROR, [err.message]));
+
+                } else if (err instanceof ApplicationService.CompanyApplicationEmailAlreadyInUse) {
+                    return res
+                        .status(HTTPStatus.CONFLICT)
+                        .json(buildErrorResponse(ErrorTypes.VALIDATION_ERROR, [err.message]));
```
GuilhermeJSilva, line 68: These 2 clauses can be joined into a single `else if`.

GuilhermeJSilva (changes requested):
Just some small changes

src/api/middleware/validators/validatorUtils.js
imnotteixeira: I don't understand. the `checkValid` function is only used in that function, I abstracted it in order to avoid replication in the if block below

src/services/application.js
imnotteixeira: Just found this is a thing: https://mongoosejs.com/docs/api.html#document_Document-toObject

Also took the chance and removed passwords from the returned objects so no more: `delete password; return retVal` ugliness

src/api/middleware/validators/validatorUtils.js
GuilhermeJSilva: I meant changing the function `checkValid` from being called like `checkValid(value)(arr)` to being invoked like `checkValid(value, arr)`

imnotteixeira: Got it, done.

GuilhermeJSilva (approved):

PR #77, opened 2020-10-01
Bump node version to 14.13.1 Current

### LINKED ISSUE #65: Update nodejs versions in Dockerfiles
The node versions we are using in the Dockerfiles seem to be LTS 8. However, the current LTS is 12.14.1.

We should update the app to this LTS, alongside the dockerfiles and any other configurations that need changing.

This is quite pressing as the version mismatch is pretty high.

xRuiAlves:
Closes #65

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:  
-  - "lts/*"
+  - "12"
 
 sudo: required
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:  
-  - "12"
+  - "14"
 
 sudo: required
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
 # Development dockerfile
-# Use node 8-LTS
-FROM node:8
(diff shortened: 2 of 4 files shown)


imnotteixeira (changes requested):
Update travis.yml to use node_js: 12 (https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions) in order to make it consistent as well

xRuiAlves:
> Update travis.yml to use node_js: 12 (https://docs.travis-ci.com/user/languages/javascript-with-nodejs/#specifying-nodejs-versions) in order to make it consistent as well

Yeap, missed that. Done :+1:

imnotteixeira (changes requested):
In two weeks nodejs 14 will be the current LTS, let's update now. https://nodejs.org/en/about/releases/

Also, I really want the optional chain operator. 👀

imnotteixeira:
> 
> 
> In two weeks nodejs 14 will be the current LTS, let's update now. https://nodejs.org/en/about/releases/
> 
> Also, I really want the optional chain operator. 👀

Don't forget to update the PR's title :P

imnotteixeira (approved):

PR #79, opened 2020-10-10
Update README docker-compose test command

### LINKED ISSUE #78: Fix README docker-compose test command
> After building the images/containers, the tests can be ran with:
> `docker-compose up test mongo --exit-code-from test`

Running this command raises an error (`ERROR: No such service: --exit-code-from`), since the flag should be specified before listing the services (this way `--exit-code-from` is being treated as a service).

The command should be updated to:

```
docker-compose up --exit-code-from test test mongo
```

xRuiAlves:
Closes #78

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ docker-compose build test
 After building the images/containers, the tests can be ran with:
 
 ```bash
-docker-compose up test mongo --exit-code-from test
+docker-compose up --exit-code-from test test
 ```
 > A `test.sh` file is available in the project's root folder to run these commands on linux environments (simply run `./test.sh [--build]`)
 


README.md
```diff
@@ -71,7 +71,7 @@ docker-compose build test
 After building the images/containers, the tests can be ran with:
 
 ```bash
-docker-compose up test mongo --exit-code-from test
+docker-compose up --exit-code-from test test mongo
```
imnotteixeira, line 74: `test` service actually has a depends_on on mongo, so you only need to start the `test` service 
```suggestion
docker-compose up --exit-code-from test test
```

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #82, opened 2020-10-11
Bump Dockerfile node version to 14.13.1 Current

xRuiAlves:
Forgot to update `Dockerfile` version in #77 :sweat_smile:

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,7 +15,7 @@
     },
 
     "parserOptions": {
-        "ecmaVersion": 9,
+        "ecmaVersion": 2020,
         "sourceType": "module"
     },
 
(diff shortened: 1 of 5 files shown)


src/services/application.js
```diff
@@ -142,7 +142,7 @@ class CompanyApplicationService {
                     .filter((application) => (stateFilter) ?
                         application.state === stateFilter ||
                         stateFilter.includes(application.state)
-                        : true
+                        : true,
```
imnotteixeira, line 145: this comma makes no sense, if you take it out, what is the eslint rule you break?

test/end-to-end/auth.js
```diff
@@ -116,7 +116,7 @@ describe("Login endpoint test", () => {
                     .send();
 
                 expect(res.status).toBe(HTTPStatus.UNAUTHORIZED);
-            }
+            },
```
imnotteixeira, line 119: This seems like it's adding a comma after the last argument of functions, remove that rule

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #84, opened 2020-12-01
Fix GET /applications/company/search filters params from body params to query params

imnotteixeira:
Title says it all. Besides, this was taking body parameters, which shouldn't be done in an HTTP GET.


miguelpduarte:
Given that this does not (or should not) have side effects, I hesitate on changing this to a POST. In theory searching makes sense via GET.
I believe that the validators and such can easily be changed to look at parameters instead of the body and as such the problem you stated (and thus the reasoning for the change I assume) would be resolved.

miguelpduarte (changes requested):
see above, I disagree with this PR (didn't send as a review 🙃 )

src/api/routes/review.js
```diff
@@ -10,44 +10,60 @@ const { buildErrorResponse, ErrorTypes } = require("../middleware/errorHandler")
...
 
-            const limit = parseInt(req.query.limit || MAX_LIMIT_RESULTS, 10);
-            const offset = parseInt(req.query.offset || 0, 10);
+            const { limit, offset, sortBy, ...filters } = req.query;
+            const computedLimit = parseInt(limit || MAX_LIMIT_RESULTS, 10);
+            const computedOffset = parseInt(offset || 0, 10);
             let sortingOptions;
 
-            if (!req.body.sortBy) {
+            if (!sortBy || sortBy.length === 0) {
```
miguelpduarte, line 56:
```suggestion
            if (!sortBy?.length) {
```
I think this also works and is simpler to look at. It might make sense to invert the if though (switch if/else order), to remove the `!` and simplify this further.

miguelpduarte (changes requested):
almost ready, just some small things.
Are the new validators being tested btw? It might make sense to do so, much like others were previously through the OP higher-order function thingies.

src/api/routes/review.js
```diff
@@ -10,44 +10,60 @@ const { buildErrorResponse, ErrorTypes } = require("../middleware/errorHandler")
 
 const router = Router();
 
+const parseSortingOptions = (sortingOptions) => sortingOptions
```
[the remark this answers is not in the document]
imnotteixeira: It's part of the route below, and it's not a generic parser, so it should not be in some "utils" folder, nor is it a validator. I understand what you're saying, but this is the "least bad" place to put I guess

src/api/routes/review.js
imnotteixeira: You're right, `let sortingOptions` already makes it `undefined` by default, no need for the extra if block

src/api/middleware/validators/validatorUtils.js
```diff
@@ -49,9 +49,15 @@ const checkDuplicatedEmail = async (email) => {
     }
 };
 
+/**
+ * Sanitize JSON string into a JS Array for validation
+ * @param {*} json string to sanitize as JS array from JSON
+ */
+const parseArrayJSON = (json) => JSON.parse(json);
```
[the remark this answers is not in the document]
imnotteixeira: Right, moved that to the validator side, calling JSON.parse and asserting that the result `isArray()`

.babelrc
```diff
@@ -0,0 +1,3 @@
+{
+    "plugins": ["@babel/plugin-proposal-optional-chaining"]
+}
```
miguelpduarte, line 3: nitpick but: no newline at the end of file bugs me 😛

Dockerfile-prod
```diff
@@ -15,4 +15,7 @@ COPY src/ src/
 # Copying .env files
 COPY .env* ./
 
+# Copying babel config
+COPY .babelrc ./
+
```
miguelpduarte, line 20: Shouldn't be necessary for prod, I'd say?

src/api/routes/review.js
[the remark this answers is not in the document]
miguelpduarte: gotcha, resolving.

Dockerfile-prod
imnotteixeira: The `sortBy?.length` would fail otherwise

src/api/middleware/validators/validatorUtils.js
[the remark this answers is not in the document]
miguelpduarte: That's perfect then! resolving this

src/api/routes/review.js
miguelpduarte: LGTM then

Dockerfile-prod
miguelpduarte: Nevermind, I read `eslint` instead of `babel` 🤦‍♂️

miguelpduarte (approved):
:shipit:

PR #85, opened 2020-12-02
Add extra info on /auth/me request

imnotteixeira:
`isAdmin` field is needed to customize the UI on some pages and even restrict others. I also added `company` information when available, in order to show the company name on the user menu, for example

### DIFF
--- a/test/end-to-end/auth.js
+++ b/test/end-to-end/auth.js
@@ -147,6 +147,8 @@ describe("Login endpoint test", () => {
 
             expect(res.status).toBe(HTTPStatus.OK);
             expect(res.body).toHaveProperty("data.email", test_user.email);
+            expect(res.body).toHaveProperty("data.isAdmin", true);
+            expect(res.body).not.toHaveProperty("data.company");
         });
 
         test("should be successful when loging out the current user", async () => {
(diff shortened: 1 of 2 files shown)


test/end-to-end/auth.js
```diff
@@ -147,6 +147,8 @@ describe("Login endpoint test", () => {
 
             expect(res.status).toBe(HTTPStatus.OK);
             expect(res.body).toHaveProperty("data.email", test_user.email);
+            expect(res.body).toHaveProperty("data.isAdmin", true);
```
miguelpduarte, line 150: Are there not other tests in which it would make sense to test that the users are specifically not admins? (Like a non-admin user using this endpoint?)

Additionally, for the following line a similar question: is there any test that has a user with a company and could be used to test this already?

miguelpduarte (changes requested):
Overall pretty simple and seems like nice changes, just some questions regarding testing.

test/end-to-end/auth.js
imnotteixeira: That feature is not yet being used (we only have logic specific for admins so far in the UI), that's why I did not work on that yet, so that we can move faster. They should be added once we start having logic for company users login. Created #86 for that

miguelpduarte: Thanks for creating the issue, this should be fine then

miguelpduarte (approved):

PR #93, opened 2020-12-14
Add Email Service and automatic emails for company applications

### LINKED ISSUE #54: Company Applications and Approvals

imnotteixeira:
Closes #54 

Added a generic Email Service that uses nodemailer to send emails as NIAEFEUP/NIJobs team.

Added the following automatic emails:
* Notify NIAEFEUP when a new application is created (sent to the MAIL_ADDRESS specified in .env file)
* Notify submission of application to the company email
* Notify approval of company to the company email

If no email is specified in the .env file, no emails will be sent and the backend won't crash


src/lib/nodemailer.js
```diff
@@ -0,0 +1,71 @@
...
+                refreshToken,
+                accessToken,
+            },
+            connectionTimeout: 30000
+        });
+    }
+
+    verifyConnection() {
+        if (!this.transporter) throw new Error("Trying to use Email Service without having a transporter set. Call init() first.");
+        return promisify(this.transporter.verify)();
```
miguelpduarte, line 28: Afaik `nodemailer` already does this internally?
> If `callback` argument is not set then the method returns a Promise object. Nodemailer itself does not use Promises internally but it wraps the return into a Promise for convenience.

In https://nodemailer.com/usage/#sending-mail

This is not extremely explicit in the "docs" though... They seem to be quite lacking...

src/lib/nodemailer.js
```diff
@@ -0,0 +1,71 @@
...
+        if (!message.attachments || !(message.attachments instanceof Array)) {
+            message.attachments = [];
+            message.attachments.push({
+                filename: "logo-niaefeup.png",
+                path: "https://ni.fe.up.pt/images/logo-niaefeup.png",
+                cid: "id_1234698",
+            });
+        }
+
+        message.html += EMAIL_SIGNATURE;
```
miguelpduarte, line 61: Should we consider use some sort of email templating to make this easier (and possibly safer)? Such as mustache/handlebars/etc

miguelpduarte: We could have base templates that would use a layout that would have the signature and everything necessary, and others would just "extend" these.

src/services/emails/companyApplicationApproval.js
```diff
@@ -0,0 +1,83 @@
+/* eslint-disable max-len */
+
```
miguelpduarte, line 2: Yeah, the string interpolation in the html makes me kind of wary of template injection. We should definitely use some templating library that handles this before shipping this. Maybe this might be in a separate PR, but before having this go live with any sort of email-sending capabilities.

The templating would also be easier to grok than these strings, most likely.

miguelpduarte: Someone in the board should also have a look at the emails, since they will be sent to companies. I think they are fine as-is, but they may benefit from some of their input or at least approval.

src/services/application.js
```diff
@@ -2,6 +2,13 @@ const CompanyApplication = require("../models/CompanyApplication");
 const CompanyApplicationRules = require("../models/CompanyApplication").CompanyApplicationRules;
 const hash = require("../lib/passwordHashing");
 const AccountService = require("./account");
+const EmailService = require("../lib/nodemailer");
+const {
+    NEW_COMPANY_APPLICATION_ADMINS,
+    NEW_COMPANY_APPLICATION_COMPANY,
+    APPROVAL_NOTIFICATION,
+} = require("./emails/companyApplicationApproval");
```
miguelpduarte, line 10: No rejection email?

src/lib/nodemailer.js
```diff
@@ -0,0 +1,71 @@
+const promisify = require("util").promisify;
+const nodemailer = require("nodemailer");
+const EMAIL_SIGNATURE = require("../services/emails/signature");
+
```
miguelpduarte, line 4: The name of the file should maybe be switched to `email.js` or `email_service.js` or something similar.

Having the name of the library as the file name might introduce some confusion as to whether the service or nodemailer itself is being used directly when browsing the code.

test/end-to-end/application.js
```diff
@@ -84,6 +88,39 @@ describe("Company application endpoint test", () => {
...
+                    html: adminEmailOptions.html,
+                    to: config.mail_address
+                });
+
+                expect(EmailService.sendMail).toHaveBeenNthCalledWith(2, {
+                    subject: companyEmailOptions.subject,
+                    text: companyEmailOptions.text,
+                    html: companyEmailOptions.html,
+                    to: application.email
+                });
```
miguelpduarte, line 121: Can these fail if they ever occur in a different order? I.e. is there a way to make this check regardless of the order the method is called in?

miguelpduarte (changes requested):
Overall looks good! Just have some suggestions!

Btw, the EmailService unit tests were quite tasty, props! :sunglasses:

src/services/application.js
imnotteixeira: I thought about it, but I didn't know what to put there. "You're not good enough" sounded too bad :P

@andrefmrocha could use your input here

src/lib/nodemailer.js
imnotteixeira: https://github.com/yads/nodemailer-express-handlebars this might work

src/services/application.js
andrefmrocha: What use cases are we thinking for rejecting a company application?

miguelpduarte: We could use the reject reason as some sort of message, besides the generic template.

Nevertheless, the notification should still exist and thus the email as well, regardless of not knowing what to say 😛

miguelpduarte: @andrefmrocha I've discussed this with you over PM but the idea is that a company is rejected if the admins don't like their application. For example their motivation is not valid, handling spam that might find the website and apply just because, etc.

andrefmrocha: It would seem that it would be tricky to create a generic email to reject a company since we have so many use cases. But probably an email saying they were rejected and forwarding any complaints to NI's email could be interesting. Wdyt @andrefmrocha @imnotteixeira ?

.env
```diff
@@ -28,3 +28,22 @@ ADMIN_EMAIL=ni@aefeup.pt
 ADMIN_PASSWORD=n1j0bs_ftw.12345
 #CORS allowed origin - OVERRIDE IN PRODUCTION
 ACCESS_CONTROL_ALLOW_ORIGIN= 
+
+# Mail service information. If you don't provide a MAIL_ADDRESS, no emails will be sent. The app will execute no-ops and won't crash
+# However, if you want to send emails, you need to fill all of the following 5 fields
+# Check this SO Post for details on how to configure your personal account for testing (https://stackoverflow.com/questions/51933601/what-is-the-definitive-way-to-use-gmail-with-oauth-and-nodemailer)
+
+# The email address from which the emails are sent
+MAIL_ADDRESS=
```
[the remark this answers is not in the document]
imnotteixeira: Sure

.env
```diff
@@ -28,3 +28,22 @@ ADMIN_EMAIL=ni@aefeup.pt
...
+MAIL_CLIENT_ID=
+
+# Client Secret used for OAuth
+MAIL_CLIENT_SECRET=
+
+# OAuth refresh token (for oauth token auto-refreshing)
+MAIL_REFRESH_TOKEN=
+
+#Initial OAuth access token, if expired, it will refresh using REFRESH_TOKEN
+MAIL_ACCESS_TOKEN=
```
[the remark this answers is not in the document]
imnotteixeira: Confirmed that it doesn't work in our prod email. Going back to simple SMTP only

package.json
```diff
@@ -8,8 +8,8 @@
     "prod": "NODE_ENV=production node src/index.js",
     "lint": "eslint src test --max-warnings 0",
     "lint-fix": "npm run lint -- --fix",
-    "test": "jest --runInBand",
-    "ci": "npm run lint && npm test",
+    "test": "NODE_ENV=test jest --runInBand",
+    "ci": "NODE_ENV=test npm run lint && npm test",
```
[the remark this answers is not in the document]
imnotteixeira: Definitely is... :P

src/lib/nodemailer.js
imnotteixeira: Yep, I knew about that (I am using it in the sendMail function), that one must have slipped...

src/lib/nodemailer.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    async sendMail(message, options) {
+        if (!this.transporter) {
+            console.warn(`Email service transporter not defined. 
+        Either you forgot to call init(), or MAIL_ADDRESS is not present in .env. Executing no-op.`);
+            return Promise.resolve();
+        }
+
+        const { sendFailureNotification = true } = { ...options }; // To allow undefined options object
```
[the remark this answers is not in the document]
imnotteixeira: I think as a developer, if there is a default value, it should be applied whenever it is isn't specified, that includes the empty object. Therefore, I am leaving as is, but I improved the explanation comment and added that test case

src/services/application.js
imnotteixeira: Well, it depends... Since the beginning I've always thought of the Reject Reason as an internal thing. It might be really bad PR to have a "bad" reject reason leak in an email by some lousy admin. In the end, it's up to you @andrefmrocha 

The options we have are:
* Use a generic email saying they were rejected (we can give some common causes of rejection such as spam) and if they think they were wrongly rejected to come back to us @ niMail
* Give them the actual reject reason in the email (must be well written every time, so it involves a bigger responsibility when reviewing)

src/services/emails/companyApplicationApproval.js
imnotteixeira: Will add https://github.com/yads/nodemailer-express-handlebars.

@andrefmrocha if you could review the emails, that would be great. Otherwise just delegate someone instead :upside_down_face:

test/end-to-end/application.js
imnotteixeira: They never failed so far, since they are sent "synchronously", one after the other. But I could use an .includes check on the calls array instead, since the order doens't really matter

src/services/application.js
andrefmrocha: I agree that the second option can only be taken if there's a review of such email. 

But perhaps this is not worth the hassle, at least for now. Why don't we go with a generic email for now? Less implementation overhead and overall less risky.

src/services/emails/companyApplicationApproval.js
```diff
@@ -0,0 +1,83 @@
...
+    NEW_COMPANY_APPLICATION_COMPANY: (companyName, applicationId) => ({
+        subject: "Your NIJobs Application",
+        text:
+            `We have successfully received your application!
+            \n
+            We will now review your application, and in case you're approved, you will receive another email with further instructions in order to complete your registration. 
+            \n
+            Your Application ID is ${applicationId} and you registered ${companyName}
+            \n
+            If you did not request this or if you need anything else, don't hesitate to contact us!
```
andrefmrocha, line 39: Even though this does send the contact details in the signature, it's probably best to mention it again here? This seems a very common pattern in these sort of emails. Wdyt?

src/services/emails/companyApplicationApproval.js
```diff
@@ -0,0 +1,83 @@
...
+            <br>
+            <p>Sincerely,</p>
+            <p>NIJobs team at NIAEFEUP</p>
+            
+        `,
+    }),
+    APPROVAL_NOTIFICATION: (companyName) => ({
+        subject: "Your NIJobs Application",
+        text:
+            `Glad to have you on board, ${companyName}!
```
andrefmrocha, line 59: How about: `It is our pleasure to welcome you aboard`?
Not sure if the current one relays that they have been accepted and since this is the first sentence to be seen, even before the email has been opened.

src/services/emails/companyApplicationApproval.js
```diff
@@ -0,0 +1,83 @@
...
+            <br>
+            <p>If you did not request this or if you need anything else, don't hesitate to contact us!</p>
+            <br>
+            <p>Sincerely,</p>
+            <p>NIJobs team at NIAEFEUP</p>
+            
+        `,
+    }),
+    APPROVAL_NOTIFICATION: (companyName) => ({
+        subject: "Your NIJobs Application",
```
andrefmrocha, line 57: It is common for the subject on an approval message to already mention that the application was accepted in some way, perhaps something simple such as `Welcome to NIJobs!`? Not sure if this is the best one, but something on these terms sounds better

andrefmrocha (changes requested):
Perhaps we can improve these email messages :thinking:

src/services/emails/companyApplicationApproval.js
imnotteixeira: Do you mean the subject or the first sentence on the body. If it is the body, they are pretty much the same imo.

Since you commented the subject specifically below, I am confused

andrefmrocha: It was about the first sentence of the body, but with the subject change, I think we can assume that's clear enough.

src/services/application.js
miguelpduarte: +1 on @imnotteixeira's first option. This way we have a consistent "face" to show to the companies. If this was a website aimed at a broader / less "serious" audience, I'd advocate for including the reject reason, as many websites do. However, what you mentioned made me prefer the other option.

As an "intermediate" alternative (for later, as it would be more effort and requires handling things that are not yet implemented) we could have some predefined reject reasons and have a generic email that includes them as simple snippets, based on what the operators chose. If the "other" reason is chosen a text field can be filled for internal consumption and the message would just be the generic part. This might be worth implementing later on when the project is already up and running, so as to provide more feedback to companies.

src/email-templates/new_company_application_admins.handlebars
```diff
@@ -0,0 +1,10 @@
+<h3>A company has requested access to NIJobs</h3>
+<br>
+<p>{{companyName}} <{{email}}> has requested to participate in NIJobs. Below is their motivation:</p>
+<blockquote><i>{{motivation}}</i></blockquote>
```
miguelpduarte, line 4: Any reason to use the `<i>` here? [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i) seems to suggest it would not make much sense.

src/email-templates/rejection_notification.handlebars
```diff
@@ -0,0 +1,8 @@
+<p>Hello {{companyName}},</p>
+<br>  
+<p>We have reviewed your application for {{companyName}} and decided to reject your application.</p>    
+<p>Some common causes of rejection are spam, abuse and innapropriate motivation.
```
miguelpduarte, line 4: `s/and/or`

miguelpduarte: innapropriate sounds weird as well, we might want to clarify with something like "or a motivation that does not align with the platform's principles" or "our principles". Not sure though, the big boss @andrefmrocha should call this one 😁

test/end-to-end/review.js
```diff
@@ -400,6 +400,25 @@ describe("Company application review endpoint test", () => {
                         expect(res.body.companyName).toBe(pendingApplication.companyName);
                     });
 
+                    test("Should send approval email to company email", async () => {
```
miguelpduarte, line 403: `s/approval/rejection`

miguelpduarte (changes requested):
Some small fixes, almost ready imo

src/email-templates/rejection_notification.handlebars
andrefmrocha: "or a motivation that does not align with our principles" sounds great in this case.

src/email-templates/new_company_application_admins.handlebars
imnotteixeira: It renders the text as italic. I think it helps to differentiate the text better (the blockquote simply indents it)

src/email-templates/rejection_notification.handlebars
imnotteixeira: Yep, this was a mere placeholder for review. I didn't really know what to put there... Changing

andrefmrocha (approved):
Lgtm :shipit:

miguelpduarte (approved):

PR #105, opened 2021-01-09
Enable proxy in express-session only in prod

### LINKED ISSUE #104: Fix express session in order to use secure cookies in prod
Currently, due to the reverse proxy setup done with [niployments](https://github.com/NIAEFEUP/niployments/), express cannot correctly parse the requester, and cannot deal with the secure cookies (check https://github.com/NIAEFEUP/niployments/pull/24 and https://github.com/expressjs/session/issues/281).

From the [express-session docs](https://github.com/expressjs/session#proxy):
> **proxy**
>
> Trust the reverse proxy when setting secure cookies (via the "X-Forwarded-Proto" header).
>
>The default value is **undefined**.
>
>    * **true** The "X-Forwarded-Proto" header will be used.
>    * **false** All headers are ignored and the connection is considered secure only if there is a direct TLS/SSL connection.
>    * **undefined** Uses the "trust proxy" setting from express

This will prompt express to look into the "X-Forwarded-Proto" header (added in #NIAEFEUP/niployments/pull/24) and make the secure cookies (and thus authentication) work in prod :sunglasses:

imnotteixeira:
Closes #104 

The issue says pretty much everything, this ensures that secure cookies work and fixes the auth problem in prod.

### DIFF
--- a/src/loaders/express.js
+++ b/src/loaders/express.js
@@ -30,6 +30,11 @@ module.exports = (app) => {
         secret: config.session_secret,
         resave: true,
         saveUninitialized: true,
+
+        // Set to true only in prod. Necessary due to reverse proxy setup
+        // Check https://github.com/expressjs/session/issues/281 for more details
+        proxy: process.env.NODE_ENV === "production",
+
         cookie: {
             secure: process.env.NODE_ENV === "production",
         },


miguelpduarte:
The motivation behind not having this for dev is just the pita of setting up https in localhost right? If so, I'll give this an approve. If not, please don't merge this and let's discuss it further.

miguelpduarte (dismissed):

imnotteixeira:
> 
> 
> The motivation behind not having this for dev is just the pita of setting up https in localhost right? If so, I'll give this an approve. If not, please don't merge this and let's discuss it further.

Basically yes, secure cookies are also not being used in dev due to that reason

miguelpduarte (approved):

PR #107, opened 2021-01-11
Feature/get offer info

### LINKED ISSUE #96: Get offer info
Create an endpoint to return the information about a specific Offer.

Something like GET /offers/:offer-id will do.

It should return:
* location
* jobType
* title
* description
* jobStartDate (might be undefined)
* minDuration and maxDuration
* vacancies
* isPaid
* requirements (if implemented after #95 )

DoStini:
Closes #96

### DIFF
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -28,8 +28,10 @@ module.exports = (app) => {
         }
     });
 
+    /**
+     * Gets an offer from the database with its id
+    */
     router.get("/:offer", validators.get, async (req, res, next) => {
-
         try {
             const offer = await (new OfferService()).getOfferByID({ id: req.params.offer });
             return res.json(offer);
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -3,6 +3,7 @@ const { Router } = require("express");
 const authMiddleware = require("../middleware/auth");
 const validators = require("../middleware/validators/offer");
 const OfferService = require("../../services/offer");
+const HTTPStatus = require("http-status-codes");
 
 const router = Router();
 
@@ -31,11 +32,16 @@ module.exports = (app) => {
     /**
      * Gets an offer from the database with its id
     */
-    router.get("/:offer", validators.get, async (req, res, next) => {
+    router.get("/:offerId", validators.get, async (req, res, next) => {
         try {
-            const offer = await (new OfferService()).getOfferByID(req.params.offer);
-
-            return res.json(offer);
+            const offer = await (new OfferService()).getOfferById(req.params.offerId);
+            if (offer !== null) {
+                return res.json(offer);
+            } else {
+                return res.status(HTTPStatus.UNAUTHORIZED).json({
+                    reason: `No offer with id:${req.params.offer}`,
+                });
+            }
 
         } catch (err) {
             return next(err);
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -28,6 +28,16 @@ module.exports = (app) => {
         }
     });
 
+    router.get("/:offer", validators.get, async (req, res, next) => {
+
+        try {
+            const offer = await (new OfferService()).getOfferByID({ id: req.params.offer });
+            return res.json(offer);
+        } catch (err) {
+            return next(err);
+        }
+    });
+
     /**
      * Creates a new Offer
      */
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -34,7 +34,27 @@ module.exports = (app) => {
     router.get("/:offer", validators.get, async (req, res, next) => {
         try {
             const offer = await (new OfferService()).getOfferByID({ id: req.params.offer });
-            return res.json(offer);
+
+            // Later add here requirements
+            const filters = [
+                "title",
+                "jobStartDate",
+                "jobMinDuration",
+                "jobMaxDuration",
+                "description",
+                "vacancies",
+                "isPaid",
+                "location",
+                "jobType"
+            ];
+
+            const filteredOffer = {};
+
+            filters.forEach((key) => {
+                filteredOffer[key] = offer[key];
+            });
+            return res.json(filteredOffer);
+
         } catch (err) {
             return next(err);
         }
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -35,10 +35,12 @@ module.exports = (app) => {
     */
     router.get("/:offerId", validators.getOfferById, async (req, res, next) => {
         try {
-            const offer = await (new OfferService()).getOfferById(req.params.offerId);
+            const offer = await (new OfferService()).getOfferById(req);
 
             if (!offer) {
-                return res.status(HTTPStatus.NOT_FOUND).json({ reason: APIErrorTypes.OFFER_NOT_FOUND(req.params.offerId) });
+                return res.status(HTTPStatus.NOT_FOUND).json({
+                    reason: APIErrorTypes.OFFER_NOT_FOUND(req.params.offerId)
+                });
             }
 
             return res.json(offer);
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -35,7 +35,7 @@ module.exports = (app) => {
     */
     router.get("/:offerId", validators.getOfferById, async (req, res, next) => {
         try {
-            const offer = await (new OfferService()).getOfferById(req);
+            const offer = await (new OfferService()).getOfferById(req.params.offerId, req.user);
 
             if (!offer) {
                 return res.status(HTTPStatus.NOT_FOUND).json({
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -33,7 +33,7 @@ module.exports = (app) => {
     /**
      * Gets an offer from the database with its id
     */
-    router.get("/:offerId", validators.get, async (req, res, next) => {
+    router.get("/:offerId", validators.getOfferById, async (req, res, next) => {
         try {
             const offer = await (new OfferService()).getOfferById(req.params.offerId);
 
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -4,6 +4,7 @@ const authMiddleware = require("../middleware/auth");
 const validators = require("../middleware/validators/offer");
 const OfferService = require("../../services/offer");
 const HTTPStatus = require("http-status-codes");
+const ValidationReasons = require("../middleware/validators/validationReasons");
 
 const router = Router();
 
@@ -35,13 +36,12 @@ module.exports = (app) => {
     router.get("/:offerId", validators.get, async (req, res, next) => {
         try {
             const offer = await (new OfferService()).getOfferById(req.params.offerId);
-            if (offer !== null) {
-                return res.json(offer);
-            } else {
-                return res.status(HTTPStatus.UNAUTHORIZED).json({
-                    reason: `No offer with id:${req.params.offer}`,
-                });
+
+            if (!offer) {
+                return res.status(HTTPStatus.NOT_FOUND).json({ reason: ValidationReasons.OFFER_NOT_FOUND(req.params.offerId) });
             }
+
+            return res.json(offer);
 
         } catch (err) {
             return next(err);
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -33,27 +33,9 @@ module.exports = (app) => {
     */
     router.get("/:offer", validators.get, async (req, res, next) => {
         try {
-            const offer = await (new OfferService()).getOfferByID({ id: req.params.offer });
+            const offer = await (new OfferService()).getOfferByID(req.params.offer);
 
-            // Later add here requirements
-            const filters = [
-                "title",
-                "jobStartDate",
-                "jobMinDuration",
-                "jobMaxDuration",
-                "description",
-                "vacancies",
-                "isPaid",
-                "location",
-                "jobType"
-            ];
-
-            const filteredOffer = {};
-
-            filters.forEach((key) => {
-                filteredOffer[key] = offer[key];
-            });
-            return res.json(filteredOffer);
+            return res.json(offer);
 
         } catch (err) {
             return next(err);
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -4,7 +4,7 @@ const authMiddleware = require("../middleware/auth");
 const validators = require("../middleware/validators/offer");
 const OfferService = require("../../services/offer");
 const HTTPStatus = require("http-status-codes");
-const ValidationReasons = require("../middleware/validators/validationReasons");
+const APIErrorTypes = require("../APIErrorTypes");
 
 const router = Router();
 
@@ -38,7 +38,7 @@ module.exports = (app) => {
             const offer = await (new OfferService()).getOfferById(req.params.offerId);
 
             if (!offer) {
-                return res.status(HTTPStatus.NOT_FOUND).json({ reason: ValidationReasons.OFFER_NOT_FOUND(req.params.offerId) });
+                return res.status(HTTPStatus.NOT_FOUND).json({ reason: APIErrorTypes.OFFER_NOT_FOUND(req.params.offerId) });
             }
 
             return res.json(offer);
--- a/src/services/offer.js
+++ b/src/services/offer.js
@@ -94,7 +94,7 @@ class OfferService {
         return constraints.length ? { "$and": constraints } : {};
     }
 
-    async getOfferByID({ id = 0 }) {
+    async getOfferByID( id = 0 ) {
         const offer = await Offer.findById(id).current();
 
         return offer;
--- a/src/services/offer.js
+++ b/src/services/offer.js
@@ -96,12 +96,10 @@ class OfferService {
         return constraints.length ? { "$and": constraints } : {};
     }
 
-    async getOfferByID({ id = 0, showHidden = false }) {
-        let offer = await Offer.findById(id).current();
-        //if (!showHidden) offer = undefined;
+    async getOfferByID({ id = 0 }) {
+        const offer = await Offer.findById(id).current();
 
         return offer;
-
     }
 
 }
--- a/src/services/offer.js
+++ b/src/services/offer.js
@@ -1,6 +1,4 @@
 const Company = require("../models/Company");
-const { Mongoose } = require("mongoose");
-const { off } = require("../models/Offer");
 const Offer = require("../models/Offer");
 
 class OfferService {
--- a/src/services/offer.js
+++ b/src/services/offer.js
@@ -94,10 +94,10 @@ class OfferService {
         return constraints.length ? { "$and": constraints } : {};
     }
 
-    async getOfferById(req) {
-        const offer = await Offer.findById(req.params.offerId);
+    async getOfferById(offerId, user) {
+        const offer = await Offer.findById(offerId);
 
-        if (offer?.isHidden && !(req.user?.isAdmin || offer.owner.toString() === req.user?.company?._id.toString())) return null;
+        if (offer?.isHidden && !(user?.isAdmin || offer.owner.toString() === user?.company?._id.toString())) return null;
 
         return offer;
     }
--- a/src/services/offer.js
+++ b/src/services/offer.js
@@ -94,8 +94,12 @@ class OfferService {
         return constraints.length ? { "$and": constraints } : {};
     }
 
-    getOfferById(id) {
-        return Offer.findById(id);
+    async getOfferById(req) {
+        const offer = await Offer.findById(req.params.offerId);
+
+        if (offer?.isHidden && !(req.user?.isAdmin || offer.owner.toString() === req.user?.company?._id.toString())) return null;
+
+        return offer;
     }
 
 }
--- a/src/services/offer.js
+++ b/src/services/offer.js
@@ -1,4 +1,6 @@
 const Company = require("../models/Company");
+const { Mongoose } = require("mongoose");
+const { off } = require("../models/Offer");
 const Offer = require("../models/Offer");
 
 class OfferService {
@@ -93,6 +95,15 @@ class OfferService {
 
         return constraints.length ? { "$and": constraints } : {};
     }
+
+    async getOfferByID({ id = 0, showHidden = false }) {
+        let offer = await Offer.findById(id).current();
+        //if (!showHidden) offer = undefined;
+
+        return offer;
+
+    }
+
 }
 
 module.exports = OfferService;
--- a/src/services/offer.js
+++ b/src/services/offer.js
@@ -94,10 +94,8 @@ class OfferService {
         return constraints.length ? { "$and": constraints } : {};
     }
 
-    async getOfferByID( id = 0 ) {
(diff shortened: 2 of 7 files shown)


DoStini:
Still need to implement tests.
Unused optional values should be returned as `vacancies : undefined` or not show up in the response?

andrefmrocha:
> Unused optional values should be returned as `vacancies : undefined` or not show up in the response?

Undefined variables in javascript are under a primitive type that already guarantees that they will not show up in the response. Moreover, whenever a field that doesn't show up in the response, it will return undefined

src/api/routes/offer.js
```diff
@@ -28,6 +28,38 @@ module.exports = (app) => {
...
+                "isPaid",
+                "location",
+                "jobType"
+            ];
+
+            const filteredOffer = {};
+
+            filters.forEach((key) => {
+                filteredOffer[key] = offer[key];
+            });
```
imnotteixeira, line 55: If you are fetching an offer by its id, don't bother having filters as it will be redundant

src/api/routes/offer.js
```diff
@@ -28,6 +28,38 @@ module.exports = (app) => {
         }
     });
 
+    /**
+     * Gets an offer from the database with its id
+    */
+    router.get("/:offer", validators.get, async (req, res, next) => {
+        try {
+            const offer = await (new OfferService()).getOfferByID({ id: req.params.offer });
```
imnotteixeira, line 36: Since the getOfferById only receives an id, the argument should be the id and not an object with an id.

Also, rename to `getOfferById` (lowercase i) to make it consistent with other `*byId` methods

src/services/offer.js
```diff
@@ -64,6 +64,13 @@ class OfferService {
         ).map((o) => o.withCompany()));
 
     }
+
+    async getOfferByID({ id = 0 }) {
+        const offer = await Offer.findById(id).current();
```
imnotteixeira, line 69: `.current()` should be removed since you are fetching an entity by Id, therefore you don't want to filter by date, sometimes you might want to just fetch it and do some operations, regardless if it is expired or not (maybe we'll want to "revive" offers in the future for example)

imnotteixeira (changes requested):
Good start! Some comments to save you some work :P

src/api/routes/offer.js
DoStini: So, just return the actual offer object found in the DB?

imnotteixeira: Exactly, and return an "error response" in case of error/not found

src/api/routes/offer.js
```diff
@@ -28,6 +29,25 @@ module.exports = (app) => {
         }
     });
 
+    /**
+     * Gets an offer from the database with its id
+    */
+    router.get("/:offer", validators.get, async (req, res, next) => {
```
imnotteixeira, line 35: You are using the same middleware for generic offer search (`validators.get`), you should create another one since you only receive one query param which is the offer id

imnotteixeira, line 35: rename to `:offerId` since that is actually what you will receive

src/api/routes/offer.js
```diff
@@ -28,6 +29,25 @@ module.exports = (app) => {
...
+        try {
+            const offer = await (new OfferService()).getOfferById(req.params.offer);
+            if (offer !== null) {
+                return res.json(offer);
+            } else {
+                return res.status(HTTPStatus.UNAUTHORIZED).json({
+                    reason: `No offer with id:${req.params.offer}`,
+                });
+            }
```
imnotteixeira, line 44: It's usually a good practice to improve readability to have only the error case in the if block, and the happy path outside, like so:

```js
            if (offer === null) {
                    res.status(HTTPStatus.UNAUTHORIZED).json({
                    reason: `No offer with id:${req.params.offer}`,
                });
            }
            return res.json(offer);
```

src/api/routes/offer.js
```diff
@@ -28,6 +29,25 @@ module.exports = (app) => {
...
+     * Gets an offer from the database with its id
+    */
+    router.get("/:offer", validators.get, async (req, res, next) => {
+        try {
+            const offer = await (new OfferService()).getOfferById(req.params.offer);
+            if (offer !== null) {
+                return res.json(offer);
+            } else {
+                return res.status(HTTPStatus.UNAUTHORIZED).json({
+                    reason: `No offer with id:${req.params.offer}`,
```
imnotteixeira, line 42: Create an entry in ValidationReasons.js for this, so that we can later access it outside of this file and make sure it follows the same schema as the others (`lowercase-words-with-hyphen:variable`)

src/api/routes/offer.js
```diff
@@ -28,6 +29,25 @@ module.exports = (app) => {
...
+    /**
+     * Gets an offer from the database with its id
+    */
+    router.get("/:offer", validators.get, async (req, res, next) => {
+        try {
+            const offer = await (new OfferService()).getOfferById(req.params.offer);
+            if (offer !== null) {
+                return res.json(offer);
+            } else {
+                return res.status(HTTPStatus.UNAUTHORIZED).json({
```
imnotteixeira, line 41: If an offer is not found, the error should not be `unauthorized` but `not found` instead (the famous HTTP 404)

src/services/offer.js
```diff
@@ -64,6 +64,12 @@ class OfferService {
         ).map((o) => o.withCompany()));
 
     }
+
+    async getOfferById(id) {
+        const offer = await Offer.findById(id);
+        return offer;
+    }
```
imnotteixeira, line 71: This can be simplified since you can return the find promise directly:

```js
getOfferById(id) {return Offer.findById(id)}
```

This works since `Offer.findById` returns a promise that resolves to the found value. You don't need the `async/await` in this case since you are not waiting for the resolving in this function, but actually in the function that calls `OfferService.getOfferById`, so you are just creating this intermediary function. LifeProTip: Never wait for a promise when you don't need to read its value.

imnotteixeira (changes requested):
Good job, don't forget to update the description of this PR with a "Closes #issue-id" so that it closes when we merge it.

I'll be waiting for the tests code :sunglasses:

src/api/routes/offer.js
DoStini: What should be in this middleware? Something like check if it's a valid id?

miguelpduarte: @DoStini exactly, just verifying the parameters it should receive. In this case it will only be an id which should be a mongoose object id. I think we have an example of that in a validator somewhere.

miguelpduarte: Note that this means renaming the usages of the parameter (the accesses to the `req` variable)

src/api/routes/offer.js
miguelpduarte: Additionally, shouldn't this be a simple `!offer` instead of a strict check with null? @imnotteixeira

src/api/routes/offer.js
miguelpduarte: Shouldn't that only be the case for the validators? Seems like scope creep to use **validator** messages for other errors.

src/api/routes/offer.js
imnotteixeira: Mongoose API states that when there is no match it will return null, however we can play it safe and use a !offer check instead

src/api/routes/offer.js
imnotteixeira: Hmmm, I see what you mean, and I requested that since I think there are a couple use cases already doing it like that. We can separate them to an ErrorReason object instead to separate if you agree

DoStini: So should I replace this `validationReason` ? https://github.com/NIAEFEUP/nijobs-be/blob/fd4bfd8d0c778376da8e384506cc2b7222fcaf65/src/api/routes/offer.js#L41

imnotteixeira: I'd say so, yes. Create a similar file called APIErrorTypes and put it in `src/api`

test/end-to-end/offer.js
```diff
@@ -603,4 +604,71 @@ describe("Offer endpoint tests", () => {
             });
         });
     });
+
+    describe("GET /offers/offerId", () => {
```
imnotteixeira, line 608:
```suggestion
    describe("GET /offers/:offerId", () => {
```

test/end-to-end/offer.js
```diff
@@ -603,4 +604,71 @@ describe("Offer endpoint tests", () => {
...
+
+        beforeAll(async () => {
+            await Offer.deleteMany({});
+        });
+
+        describe("Id Validation", () => {
+            test("Invalid ID", async () => {
```
imnotteixeira, line 615: change the test title to represent what should happen in the form "should ...."

test/end-to-end/offer.js
```diff
@@ -603,4 +604,71 @@ describe("Offer endpoint tests", () => {
...
+                const res = await request()
+                    .get("/offers/new/123");
+
+                expect(res.status).toBe(HTTPStatus.NOT_FOUND);
+                expect(res.body).toHaveProperty("reason", ValidationReasons.OBJECT_ID);
+            });
+            test("Offer not found", async () => {
```
imnotteixeira, line 622: Same comment for the test title

test/end-to-end/offer.js
```diff
@@ -603,4 +604,71 @@ describe("Offer endpoint tests", () => {
...
+                const id = "5facf0cdb8bc30016ee58952";
```
imnotteixeira, line 623: Wrap this in a Types.ObjectId from mongoose, to make sure that this string is always a valid mongo objected, otherwise it will throw something and the test fails. 

Then to be used below you might need to use id.toString(), but I'm not sure so try without it first but keep it in mind

test/end-to-end/offer.js
```diff
@@ -603,4 +604,71 @@ describe("Offer endpoint tests", () => {
...
+                const res = await request()
+                    .get(`/offers/new/${id}`);
```
imnotteixeira, line 625: This is probably not the endpoint you're testing :P

test/end-to-end/offer.js
```diff
@@ -603,4 +604,71 @@ describe("Offer endpoint tests", () => {
...
+                publishDate: "2019-11-22T00:00:00.000Z",
+                publishEndDate: "2019-11-28T00:00:00.000Z",
+                description: "For Testing Purposes",
+                contacts: ["geral@niaefeup.pt", "229417766"],
+                jobType: "SUMMER INTERNSHIP",
+                fields: ["FRONT_END", "BACK_END"],
+                technologies: ["Flutter", "Node.js"],
+                // owner: Will be set in beforeAll,
+                location: "Testing Avenue, Test State, 321",
+            };
```
imnotteixeira, line 656: In #108 I made a change request to have a function that generates test offers to clean the code a bit. I'm not sure about which PR will be ready first, they are quite tied in progress, so check that and apply it here as well (the idea would be to reuse the function that was created there)

test/end-to-end/offer.js
```diff
@@ -603,4 +604,71 @@ describe("Offer endpoint tests", () => {
...
+            beforeAll(async () => {
+                await Offer.create([test_offer_1, test_offer_2]);
```
imnotteixeira, line 658: Need to set the owner and ownerName of the offers before creating, check other tests for examples

test/end-to-end/offer.js
```diff
@@ -603,4 +604,71 @@ describe("Offer endpoint tests", () => {
...
+            });
+
+            test("Offer 1", async () => {
```
imnotteixeira, line 661: same for test title

test/end-to-end/offer.js
```diff
@@ -603,4 +604,71 @@ describe("Offer endpoint tests", () => {
...
+                const res = await request().get(`/offers/${test_offer_1.id}`);
+                expect(res.status).toBe(HTTPStatus.OK);
+                expect(res.body).toBe(test_offer_1);
+            });
+
+            test("Offer 1", async () => {
+                const res = await request().get(`/offers/${test_offer_1.id}`);
+                expect(res.status).toBe(HTTPStatus.OK);
+                expect(res.body).toBe(test_offer_1);
+            });
```
imnotteixeira, line 671: This should only be 1 test, since it is the same test. Just update the title and merge the two tests (basically making the two requests and the `expects` of the first between them

test/end-to-end/offer.js
```diff
@@ -603,4 +604,71 @@ describe("Offer endpoint tests", () => {
...
+
+            test("Offer 1", async () => {
+                const res = await request().get(`/offers/${test_offer_1.id}`);
```
imnotteixeira, line 668:
```suggestion
                const res = await request().get(`/offers/${test_offer_1._id}`);
```

test/end-to-end/offer.js
```diff
@@ -603,4 +604,71 @@ describe("Offer endpoint tests", () => {
...
+                expect(res.status).toBe(HTTPStatus.OK);
+                expect(res.body).toBe(test_offer_1);
```
imnotteixeira, line 670: I'm not sure if this works due to the fields mongoose puts in the offer object, check other tests that verify offer returns as well.

test/end-to-end/offer.js
```diff
@@ -603,4 +604,71 @@ describe("Offer endpoint tests", () => {
...
+            });
+        });
+    });
```
imnotteixeira, line 673: Add test for non-admins and non-owners fetching a hidden offer (and a valid user fetching it and working correctly)

To use sessions in tests you have some examples above in the `GET offers` tests which also has logic to limit the `showHidden` parameter

imnotteixeira (changes requested):
Almost done!

src/services/offer.js
```diff
@@ -65,8 +65,12 @@ class OfferService {
...
     }
 
-    getOfferById(id) {
-        return Offer.findById(id);
+    async getOfferById(req) {
+        const offer = await Offer.findById(req.params.offerId);
+
+        if (offer?.isHidden && !(req.user?.isAdmin || offer.owner.toString() === req.user?.company?._id.toString())) return null;
+
+        return offer;
```
imnotteixeira, line 73: The service should be unaware of the routing logic, pass the user and the offerId instead

```suggestion
    async getOfferById(offerId, user) {
        const offer = await Offer.findById(offerId);

        if (offer?.isHidden && !(user?.isAdmin || offer.owner.toString() === user?.company?._id.toString())) return null;

        return offer;
```

src/api/routes/offer.js
```diff
@@ -35,10 +35,12 @@ module.exports = (app) => {
     */
     router.get("/:offerId", validators.getOfferById, async (req, res, next) => {
         try {
-            const offer = await (new OfferService()).getOfferById(req.params.offerId);
+            const offer = await (new OfferService()).getOfferById(req);
```
imnotteixeira, line 38:
```suggestion
            const offer = await (new OfferService()).getOfferById(req.params.offerId, req.user);
```

test/end-to-end/offer.js
```diff
@@ -633,41 +632,139 @@ describe("Offer endpoint tests", () => {
...
+                publishEndDate: "2021-02-28T00:00:00.000Z",
                 description: "For Testing Purposes",
                 contacts: ["geral@niaefeup.pt", "229417766"],
                 jobType: "SUMMER INTERNSHIP",
                 fields: ["FRONT_END", "BACK_END"],
                 technologies: ["Flutter", "Node.js"],
+                isHidden: false,
                 // owner: Will be set in beforeAll,
                 location: "Testing Avenue, Test State, 321",
             };
```
imnotteixeira: These date changes will "expire" after some time, you need a test_offer generation function that uses a valid date every time. We can sync in PM about this

test/end-to-end/offer.js
```diff
@@ -633,41 +632,139 @@ describe("Offer endpoint tests", () => {
...
+
+            const test_offer_3 = {
```
imnotteixeira, line 659: this should be called hidden_offer or something to make it clearer

test/end-to-end/offer.js
```diff
@@ -633,41 +632,139 @@ describe("Offer endpoint tests", () => {
...
+                contacts: ["geral@niaefeup.pt", "229417766"],
+                jobType: "SUMMER INTERNSHIP",
+                fields: ["FRONT_END", "BACK_END"],
+                technologies: ["Flutter", "Node.js"],
+                isHidden: true,
+                // owner: Will be set in beforeAll,
+                location: "Area 51, Secret State, 321",
+            };
+
+            let test_1, test_2, test_3;
```
imnotteixeira, line 673: These variable names are too generic (or even wrong, since they are not "tests")

test/end-to-end/offer.js
```diff
@@ -633,41 +632,139 @@ describe("Offer endpoint tests", () => {
...
+                await Account.create({
+                    email: test_user_admin.email,
+                    password: await hash(test_user_admin.password),
+                    isAdmin: true
+                });
+                await Account.create({
+                    email: test_user_company.email,
+                    password: await hash(test_user_company.password),
+                    company: test_company._id
+                });
```
imnotteixeira: Something similar is being done a test/describe above. Is there a way we can reuse the other one (maybe by moving this logic to a common describe between both?

test/end-to-end/offer.js
```diff
@@ -633,41 +632,139 @@ describe("Offer endpoint tests", () => {
...
+                    await Promise.all([test_offer_1, test_offer_2, test_offer_3].map(async (offer) => {
+                        offer.owner = test_company._id.toString();
+                        const { _id } = await Offer.create(offer);
+                        offer["_id"] = _id.toString();
+                        return offer;
+                    }));
+                // test_user_non_admin = {
+                //     email: "company@email.com",
+                //     password: "password123",
+                // };
```
imnotteixeira, line 717: remove comment

test/end-to-end/offer.js
```diff
@@ -633,41 +632,139 @@ describe("Offer endpoint tests", () => {
...
+                    password: await hash(test_user_company.password),
+                    company: test_company._id
+                });
+                [test_1, test_2, test_3] =
+                    await Promise.all([test_offer_1, test_offer_2, test_offer_3].map(async (offer) => {
+                        offer.owner = test_company._id.toString();
+                        const { _id } = await Offer.create(offer);
+                        offer["_id"] = _id.toString();
+                        return offer;
+                    }));
```
imnotteixeira, line 713:
```suggestion
                    const createOffers = async (offer) => {
                        offer.owner = test_company._id.toString();
                        offer.ownerName = test_company.name;
                        const { _id } = await Offer.create(offer);
                        return {
                            ...offer, 
                            _id: _id.toString()
                        };
                    }
                    
                    await Promise.all([test_offer_1, test_offer_2, test_offer_3].map(createOffers));
```

test/end-to-end/offer.js
```diff
@@ -633,41 +632,139 @@ describe("Offer endpoint tests", () => {
...
+                // test_user_non_admin = {
+                //     email: "company@email.com",
+                //     password: "password123",
+                // };
+            });
+
+            test("should return offer", async () => {
+                let res = await test_agent.get(`/offers/${test_1["_id"]}`);
```
imnotteixeira, line 721:
```suggestion
                const res1 = await test_agent.get(`/offers/${test_1._id}`);
```

test/end-to-end/offer.js
```diff
@@ -633,41 +632,139 @@ describe("Offer endpoint tests", () => {
...
+                expect(res.status).toBe(HTTPStatus.OK);
+                const extracted_data_1 = res.body;
+
+                res = await test_agent.get(`/offers/${test_2["_id"]}`);
+                expect(res.status).toBe(HTTPStatus.OK);
+                const extracted_data_2 = res.body;
+
+                expect(extracted_data_1).toMatchObject(test_1);
```
imnotteixeira, line 729: Move this above (before the second request) so that if it fails, no unnecessary requests are made

test/end-to-end/offer.js
```diff
@@ -633,41 +632,139 @@ describe("Offer endpoint tests", () => {
...
+                //     password: "password123",
+                // };
+            });
+
+            test("should return offer", async () => {
+                let res = await test_agent.get(`/offers/${test_1["_id"]}`);
+                expect(res.status).toBe(HTTPStatus.OK);
+                const extracted_data_1 = res.body;
+
+                res = await test_agent.get(`/offers/${test_2["_id"]}`);
```
imnotteixeira, line 725:
```suggestion
                const res2 = await test_agent.get(`/offers/${test_2._id}`);
```

test/end-to-end/offer.js
```diff
@@ -633,41 +632,139 @@ describe("Offer endpoint tests", () => {
...
+                expect(res.status).toBe(HTTPStatus.OK);
+                const extracted_data_2 = res.body;
+
+                expect(extracted_data_1).toMatchObject(test_1);
+                expect(extracted_data_2).toMatchObject(test_2);
+            });
+
+            test("should fail if not admin or owner", async () => {
+                const res = await test_agent.get(`/offers/${test_3["_id"]}`);
```
imnotteixeira, line 734:
```suggestion
                const res = await test_agent.get(`/offers/${test_3._id}`);
```

test/end-to-end/offer.js
```diff
@@ -633,41 +632,139 @@ describe("Offer endpoint tests", () => {
...
 
-            test("Offer 1", async () => {
-                const res = await request().get(`/offers/${test_offer_1.id}`);
+            test("should return hidden offer as admin", async () => {
+                await test_agent
+                    .post("/auth/login")
+                    .send(test_user_admin)
+                    .expect(200);
+
+                const res = await test_agent.get(`/offers/${test_3["_id"]}`);
```
imnotteixeira, line 744:
```suggestion
                const res = await test_agent.get(`/offers/${test_3._id}`);
```

test/end-to-end/offer.js
```diff
@@ -633,41 +632,139 @@ describe("Offer endpoint tests", () => {
...
 
-            test("Offer 1", async () => {
-                const res = await request().get(`/offers/${test_offer_1.id}`);
+            test("should return hidden offer as company", async () => {
+                await test_agent
+                    .post("/auth/login")
+                    .send(test_user_company)
+                    .expect(200);
+
+                const res = await test_agent.get(`/offers/${test_3["_id"]}`);
```
imnotteixeira, line 758:
```suggestion
                const res = await test_agent.get(`/offers/${test_3._id}`);
```

test/end-to-end/offer.js
```diff
@@ -633,41 +632,139 @@ describe("Offer endpoint tests", () => {
...
-                expect(res.body).toBe(test_offer_1);
+                const extracted_data = res.body;
+
+                expect(extracted_data).toMatchObject(test_3);
+            });
+
+            afterAll(async () => {
+                await Company.deleteMany({});
+                await Offer.deleteMany({});
             });
```
imnotteixeira: Unnecessary, each describe does its setup in the beforeAll (cleaning existing stuff and creating whatever)

But if it was necessary, it should be before any test:

describe
\ - beforeAll
\ - afterAll
\ - beforeEach
\ - afterEach
\ - tests

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #110, opened 2021-01-25
Feature/edit offer

### LINKED ISSUE #59: Edit Offers
A Company Account should be able to edit its Company's Offers.

DoStini:
Closes #59


DoStini:
- Not really sure if verifying if the user is the correct owner should be done under param `id`, maybe outside middleware?
- Both publishDate and publishEndDate are being verified when both params are being edited and when one of the is being edited alone. Not really sure if using `Date.parse("").toISOString` is the cleanest way of doing this conversion.

src/api/middleware/validators/offer.js
```diff
@@ -130,6 +130,160 @@ const create = useExpressValidators([
         .isArray(),
 ]);
 
+const edit = useExpressValidators([
+    body("_id", ValidationReasons.DEFAULT)
```
imnotteixeira, line 134: This should be a URL param, not a body param

src/api/routes/offer.js
```diff
@@ -68,4 +68,12 @@ module.exports = (app) => {
             return next(err);
         }
     });
+
+    router.post("/edit", authMiddleware.isCompanyOrGod, validators.edit, (req, res, next) => {
```
imnotteixeira, line 72: Change to `/edit/:offerId` and access the offer id through req.params

src/api/middleware/validators/offer.js
```diff
@@ -130,6 +130,160 @@ const create = useExpressValidators([
...
+                const offer = await (new OfferService()).getOfferById(_id, req.user);
+                if (!offer) throw new Error(ValidationReasons.OFFER_NOT_FOUND(_id));
+                if (req.user?.company.toString() !== offer.owner.toString()) throw new Error(ValidationReasons.NOT_OFFER_OWNER(_id));
+            } catch (_e) {
+                // Also catches any fail to the DB
+                throw new Error(_e);
+            }
+
+            return true;
+        }),
```
imnotteixeira, line 148: This should be another middleware, as you said, something like `hasPermissionsToEdit` which verifies if the logged in user is the company owner, or if it is admin or god (these 2 are missing it seems)

src/api/middleware/validators/offer.js
```diff
@@ -130,6 +130,160 @@ const create = useExpressValidators([
...
+                    // end date is earlier than publish date, error!
+                    throw new Error(ValidationReasons.MUST_BE_BEFORE("publishEndDate"));
+                }
+
+            } catch (_e) {
+                // Also catches any fail to the DB
+                throw new Error(_e);
+            }
+            return true;
+        })
```
imnotteixeira, line 185: This validator got too big and is confusing to read and review (same as publishEndDate one). Consider moving the inner function outside, and even splitting the function in smaller more clear bits

src/api/routes/offer.js
imnotteixeira, line 72: Add a middleware validator that checks if the offer is editable in the first place (if not .current() or in grace period <- TODO)

An easy way to check if offer is "current" is to `Offer.find({_id}).current()`. If it returns an offer, it will be current, otherwise it will return null. Additionally, remember that offers that have a publishEndDate in the past (already past the "current" status) cannot be edited as well for history keeping reasons

imnotteixeira (changes requested):
Did some preliminary review. Did not review the validators fully but left some hopefully helpful comments on those

src/api/middleware/validators/offer.js
```diff
@@ -130,6 +130,160 @@ const create = useExpressValidators([
...
+        .isAfter().withMessage(ValidationReasons.DATE_EXPIRED).bail()
+        .custom(async (publishEndDateCandidate, { req }) => {
+            try {
+                const offer = await (new OfferService()).getOfferById(req.body._id, req.user);
+
+                const { publishDate: publishDateRaw } = req.body;
+                const publishDate = !publishDateRaw
+                    ? publishDateRaw
+                    : (new Date(Date.parse(publishDateRaw))).toISOString();
+                const publishEndDate = (new Date(Date.parse(publishEndDateCandidate))).toISOString();
```
imnotteixeira, line 200: You should use the same naming for both cases (avoid *Raw vs *Candidate) -> rename publishDateRaw to publishDateCandidate

Also, same as some other place: The date parsing is not necessary, since you are already validating that the received string is a ISO8601 Date string

imnotteixeira (changes requested):
Forgot these :open_mouth:

src/api/middleware/validators/offer.js
DoStini: So should I create a custom function `hasPermissionsToEdit`. But should I do it in `id` aswell? Like `param("offerId).custom(hasPermissionsToEdit)`

imnotteixeira: No, it should be another middleware (that runs before the request middleware), similar to the isCompanyOrGod, for example.

You'd have something like isCompanyOrGod => isEditableOffer => hasPermissionsToEdit => validators.edit

DoStini: I see, will do!

src/api/routes/offer.js
DoStini: Current indicates if the offer is active?

imnotteixeira: Current indicates that "today" is between publishDate and publishEndDate so it might be visible in search

The offer can however be hidden and stop being shown (so I don't like to use the word "active" for this)

src/services/offer.js
```diff
@@ -54,6 +54,65 @@ class OfferService {
...
+        const offer = await Offer.findOneAndUpdate(
+            query,
+            edits,
+            { new: true },
+            (err) => {
+                if (err) {
+                    throw err;
+                }
+            }
+        );
```
imnotteixeira, line 115:
```suggestion

        const fields = {
                title,
                publishDate,
                publishEndDate,
                jobMinDuration,
                jobMaxDuration,
                jobStartDate,
                description,
                contacts,
                isPaid,
                vacancies,
                jobType,
                fields,
                technologies,
                isHidden,
                location,
                coordinates,
                requirements,
            }
 
       try {
        const offer = await Offer.findByIdAndUpdate(
            _id,
            fields,
            { new: true, omitUndefined:true },
        );
        } catch (err) {
           console.error(err)
           throw err
       }
```

If there is an error here it will be caught in the route file. Just adding logging so we can know if it fails

test/end-to-end/offer.js
```diff
@@ -516,7 +517,7 @@ describe("Offer endpoint tests", () => {
 
             beforeAll(async () => {
 
-                await Company.deleteMany({});
+                // await Company.deleteMany({});
```
imnotteixeira, line 520: Why are you not deleting?

test/end-to-end/offer.js
```diff
@@ -1103,4 +1109,400 @@ describe("Offer endpoint tests", () => {
             });
         });
     });
+
+    describe("POST /offers/edit/:/offerId", () => {
```
imnotteixeira, line 1113:
```suggestion
    describe("POST /offers/edit/:offerId", () => {
```

test/end-to-end/offer.js
```diff
@@ -1103,4 +1109,400 @@ describe("Offer endpoint tests", () => {
...
+                });
+                return {
+                    ...offer,
+                    owner: owner.toString(),
+                    ownerName,
+                    _id: _id.toString(),
+                    jobMinDuration,
+                    jobMaxDuration
+                };
+            };
```
imnotteixeira, line 1140: It feels like you should modify the `generateTestOffer` to allow any parameter customization. Something like:

```js
const generateTestOffer = ({...params}) => ({
title: <deafultTitle>
otherparam: <defaultvalue>
...params // this will override the  default params with the passed in values inside the object argument
})
```

test/end-to-end/offer.js
```diff
@@ -1103,4 +1109,400 @@ describe("Offer endpoint tests", () => {
...
+            clean_test_offer = await createOffer(generateTestOffer(
+                (new Date(Date.now() + (2 * DAY_TO_MS))).toISOString(),
+                (new Date(Date.now() + (3 * DAY_TO_MS))).toISOString()
+            ));
+        });
+
+        test("should fail not logged in", async () => {
+            await test_agent
+                .post("/offers/edit/notavalidid")
+                .expect(HTTPStatus.UNAUTHORIZED);
```
imnotteixeira, line 1168: Assert the error message as well (check the res.body.errors as the other tests)

Also, you should have a test with a valid id, that should fail as well

test/end-to-end/offer.js
```diff
@@ -1103,4 +1109,400 @@ describe("Offer endpoint tests", () => {
...
+                    .expect(HTTPStatus.OK);
+            });
+
+            describe("testing edit dates in the past", () => {
+                test("should fail if publishDate in the past", async () => {
+                    const res = await test_agent
+                        .post(`/offers/edit/${future_test_offer._id.toString()}`)
+                        .send(withGodToken({ "publishDate": "2019-02-01" }))
+                        .expect(HTTPStatus.UNPROCESSABLE_ENTITY);
+                    expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.DATE_EXPIRED);
```
imnotteixeira, line 1242: assert the field that is throwing the error as well  (i.e. check the `param` field of `res.body.errors[0]` as well as the `msg`)

test/end-to-end/offer.js
```diff
@@ -1103,4 +1109,400 @@ describe("Offer endpoint tests", () => {
...
+                });
+
+                test("should fail if publishEndDate in the past", async () => {
+                    const res = await test_agent
+                        .post(`/offers/edit/${future_test_offer._id.toString()}`)
+                        .send(withGodToken({ "publishEndDate": "2019-02-01" }))
+                        .expect(HTTPStatus.UNPROCESSABLE_ENTITY);
+                    expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.DATE_EXPIRED);
```
imnotteixeira, line 1250: assert the field that is throwing the error as well  (i.e. check the `param` field of `res.body.errors[0]` as well as the `msg`)

test/end-to-end/offer.js
```diff
@@ -1103,4 +1109,400 @@ describe("Offer endpoint tests", () => {
...
+                test("should edit if is after offer's publishEndDate", async () => {
+                    const newDate = (new Date(new Date(future_test_offer.publishEndDate).getTime() - DAY_TO_MS));
+                    const res = await test_agent
+                        .post(`/offers/edit/${future_test_offer._id.toString()}`)
+                        .send(withGodToken({ "publishDate": newDate.toISOString() }))
+                        .expect(HTTPStatus.OK);
+                    expect(res.body).toHaveProperty("publishDate", newDate.toISOString());
+                });
+
+                test("should edit if is after offer's publishEndDate", async () => {
```
imnotteixeira, line 1299: These two tests have the same name, please differentiate them (it's not a problem if the title is long, as long as it's understandable what the test is testing)

test/end-to-end/offer.js
```diff
@@ -1103,4 +1109,400 @@ describe("Offer endpoint tests", () => {
...
+                    const edits = { "title": "This is a new title" };
+                    const res = await test_agent.post(`/offers/edit/${clean_test_offer._id}`)
+                        .send(edits)
+                        .expect(HTTPStatus.OK);
+                    expect(res.body).toMatchObject({
+                        ...clean_test_offer,
+                        ...edits
+                    });
+                    clean_test_offer = res.body;
+                    delete clean_test_offer["updatedAt"];
```
imnotteixeira, line 1484: this test should not mutate an offer that is available to other tests, as it can cause bugs in the future for someone who doesn't know how the tests are done. It would be really hard to debug

imnotteixeira: Alternatively, you should create an offer within this describe and edit that instead, or, if the clean_test_offer is not used anywhere else, move it to this describe block

imnotteixeira (changes requested):

src/api/middleware/validators/offer.js
```diff
@@ -135,6 +138,232 @@ const create = useExpressValidators([
...
+
+        // If the new publishEndDate is after the new publishDate, the verification will be done in publishEndDate
+        if (jobMinDuration >= offer.jobMaxDuration.toString() &&
+            !jobMaxDuration) {
+
+            // end date is earlier than publish date, error!
+            throw new Error(ValidationReasons.MUST_BE_BEFORE("jobMaxDuration"));
+        }
+    } catch (err) {
+        throw new Error(err);
```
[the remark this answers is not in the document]
DoStini: I think I used new Error(err) because of eslint, but will verify again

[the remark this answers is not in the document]
DoStini: "Unnecessary try/catch wrapper.eslintno-useless-catch"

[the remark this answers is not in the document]
DoStini: But with console.error it's fine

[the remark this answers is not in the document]
imnotteixeira: Yep, a try catch that just throws the error is useless, but we should add the console.error, so it's fine

src/api/middleware/validators/offer.js
```diff
@@ -135,6 +138,235 @@ const create = useExpressValidators([
...
+
+    // Verify if offer editing grace period is over
+    const timeDiff = currentDate - offer.createdAt;
+    const diffInHours = timeDiff / HOUR_IN_MS;
+
+    // Should implement here grace period verification instead of true
+    if (offer.publishEndDate.toISOString() <= currentDate.toISOString() ||
+        (offer.publishDate.toISOString() <= currentDate.toISOString() && diffInHours > OFFER_EDIT_GRACE_PERIOD)) {
+        return res.status(HTTPStatus.FORBIDDEN).json({
+            reason: ValidationReasons.OFFER_EDIT_PERIOD_OVER(req.params.offerId),
```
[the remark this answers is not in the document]
DoStini: In what units should this be done?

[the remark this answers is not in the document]
imnotteixeira: Hours, I think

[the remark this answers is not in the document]
DoStini: Ok

src/api/middleware/validators/offer.js
```diff
@@ -135,6 +138,232 @@ const create = useExpressValidators([
...
+        if (publishDateCandidate) {
+            try {
+                publishDate = (new Date(Date.parse(publishDateCandidate))).toISOString();
+            } catch (_e) {
+                return false;
+            }
+        }
+        publishDate = publishDate || offer.publishDate.toISOString();
+
+        // If the new publishEndDate is after the new publishDate, this verification is already done in publishDate
```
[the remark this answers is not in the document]
DoStini: Forgot about this, I'll verify this aswell

test/end-to-end/offer.js
DoStini: I wanted first to hear feedback

src/services/offer.js
DoStini: `web-dev_1   | Title:  undefined`
`web-dev_1   | PublishDate:  null`
`web-dev_1   | PublishEndDate:  null`
This causes PublishDate to be edited to null

test/end-to-end/offer.js
imnotteixeira: I meant deleting as calling the `deleteMany()`, why did you remove it? Every describe should delete + recreate the required data

src/services/offer.js
imnotteixeira: The problem is in the request validator that is not failing with null

imnotteixeira: Actually, there are two things here:
* If this is call by the route logic, it will never be null, since it's always a string, and it is verified by the validators
* If called by any other place, null should actually be allowed, since it is a valid update in terms of the model (it's like unpublishing an offer)

DoStini: Ok, found the problem. Needed to place `.optional` in date sanitizers, otherwise it would try to convert `undefined`

src/api/middleware/validators/offer.js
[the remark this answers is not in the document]
DoStini: Currently implementing this, but I think instead of a throw I should return a custom response, I don't really like the look it will have. Any other suggestion?
`   { "error_code": 1,
    "errors": [
        {
            "value": "2022-03-01",
            "msg": "cant-exceed:6-got:12.00",
            "param": "publishEndDate",
            "location": "body"
        }
    ]
}`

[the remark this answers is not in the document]
DoStini: Or maybe I should just return the ammount of months passed, like here https://github.com/NIAEFEUP/nijobs-be/pull/110#discussion_r567444139

src/api/middleware/auth.js
```diff
@@ -48,9 +57,40 @@ const isAdmin = (req, res, next) => {
...
+    }
+    if (offer.owner.toString() !== req.user?.company?._id.toString() &&
+                req.body.god_token !== config.god_token &&
+                !req.user?.isAdmin) {
+        return res.status(HTTPStatus.FORBIDDEN).json({
+            reason: ValidationReasons.NOT_OFFER_OWNER(offerId),
+            error_code: ErrorTypes.FORBIDDEN,
+        });
+    }
+    return next();
```
imnotteixeira, line 85:
```suggestion
    try {
        const offer = await (new OfferService()).getOfferById(offerId, req.user);
        
        if (offer.owner.toString() !== req.user?.company?._id.toString() &&
                req.body.god_token !== config.god_token &&
                !req.user?.isAdmin) {
        return res.status(HTTPStatus.FORBIDDEN).json({
            reason: ValidationReasons.NOT_OFFER_OWNER(offerId),
            error_code: ErrorTypes.FORBIDDEN,
        });
    }
    return next();
    } catch (err) {
        console.error(err);
        throw err;
    }
    
```

I'm not sure how the code was before when I told you to separate it to another try/catch, but it's really not necessary, since inside the verifications are returns, not throws, so the catch would only trigger when the offer does not exist. 

Next time you can refute if you notice stuff like this :P

imnotteixeira: After seeing another comment on this, I understood. So yep, you can use only one try catch as mentioned above, and keep the return, instead of the throw you had before, exactly like the suggestion

src/api/middleware/validators/offer.js
```diff
@@ -20,19 +29,17 @@ const create = useExpressValidators([
...
+        .isISO8601({ strict: true }).withMessage(ValidationReasons.DATE).bail(),
 
     body("publishEndDate", ValidationReasons.DEFAULT)
         .exists().withMessage(ValidationReasons.REQUIRED).bail()
         .isISO8601({ strict: true }).withMessage(ValidationReasons.DATE).bail()
         .isAfter().withMessage(ValidationReasons.DATE_EXPIRED).bail()
         .custom((publishEndDateCandidate, { req }) => {
-            const { publishDate: publishDateRaw } = req.body;
+            const { publishDate: publishDateCandidate } = req.body;
+            // publishDateRaw will be a Date instance because it's sanitized on the publishDateValidator
```
imnotteixeira, line 40: remove comment, since publishDateRaw does not exist, remember to check dangling comments always, as they do more harm than good

src/api/middleware/validators/offer.js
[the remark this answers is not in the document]
imnotteixeira: Just to clarify: there's nothing needed here, right? Regarding this PR, at least

imnotteixeira (changes requested):

src/api/middleware/auth.js
DoStini: So should I just keep the changes I made or revert them? I'm a little bit confused

src/services/offer.js
```diff
@@ -54,6 +54,60 @@ class OfferService {
...
+            },
+            { new: true, omitUndefined: true },
+            (err) => {
+                if (err) {
+                    throw err;
+                }
+            }
+        );
 
         return offer;
```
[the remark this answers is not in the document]
DoStini: https://github.com/NIAEFEUP/nijobs-be/pull/110#discussion_r569855749 Weird, thought I had removed this, probably got lost in some commit maybe, will fix.
https://github.com/NIAEFEUP/nijobs-be/pull/110#discussion_r569855749 This comment is probably not very clear, so I will clarify it

src/api/middleware/auth.js
imnotteixeira: just apply the suggested changes, there's a mix of keeping stuff and applying new things

Before, you had a try with different throws inside, that's why I told you to separate them. Now, since there's only one throw, there's no need for separation, hence the 1 try/catch with the throw and return inside

test/end-to-end/offer.js
```diff
@@ -1103,4 +1120,423 @@ describe("Offer endpoint tests", () => {
...
+                "jobMaxDuration": 5,
+            }
+            ));
+        });
+
+        test("should fail not logged in", async () => {
+            const res = await test_agent
+                .post(`/offers/edit/${future_test_offer._id}`)
+                .expect(HTTPStatus.UNAUTHORIZED);
+            expect(res.body).toHaveProperty("reason", "Insufficient Permissions");
```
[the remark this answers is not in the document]
DoStini: I don't see any constant related to this in ValidationReasons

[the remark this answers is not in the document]
imnotteixeira: It should, please create the entry and use it in any validator/middleware that needs it

imnotteixeira (approved):
Good job, this concludes your graduation from NIJobs. You should be able to take on any issue from now on :sunglasses:

PR #111, opened 2021-01-25
Feature/offer requirements

BrunoRosendo:


src/models/Offer.js
```diff
@@ -84,6 +84,7 @@ const OfferSchema = new Schema({
     ownerName: { type: String, required: true },
 
     location: { type: String, required: true },
+    requirements: { type: [String], required: true },
```
imnotteixeira, line 87: This only validates that the present field is an array, so empty arrays would be valid. Check this: https://mongoosejs.com/docs/migrating_to_5.html#array-required

test/end-to-end/offer.js
```diff
@@ -983,6 +976,24 @@ describe("Offer endpoint tests", () => {
                     expect(extracted_data).toContainEqual(prepared_test_offer);
                 });
             });
+
+            describe("Search for offer requirements", () => {
```
imnotteixeira, line 980: This is not search, but normal fetching. You can rename to simply "Offer Requirements"

imnotteixeira (changes requested):
You're missing the request validation (on the route definition). Must check that requirements is provided, is array, and has at least 1 requirement, similar to the `contacts` field.

Then add the test for that int the `end-to-end/offer` (`Input Validation` describe block already existing) as well as the schema test in `offer_schema.js`

test/offer_schema.js
```diff
@@ -114,6 +114,18 @@ describe("# Offer Schema tests", () => {
                     expect(err.errors.location).toHaveProperty("message", "Path `location` is required.");
                 }
             });
+
+            test("'requirements' is required", async () => {
+                const offer = new Offer({});
+                // Returning the validation promise to ensure the test doesn't finish before all the assertions do
```
imnotteixeira, line 120: This is why we don't usually put too many comments, they can get outdated and lose meaning.
Please dele this comment and all others in that file ( and other schema test files if they use the same structure)

imnotteixeira (changes requested):

test/offer_schema.js
```diff
@@ -155,7 +157,6 @@ describe("# Offer Schema tests", () => {
                 test("Inside the range should not throw error", async () => {
                     const submitted_fields = [];
                     for (let i = 0; i < MIN_FIELDS; ++i) {
-                        // Preventing interference from duplicate error
```
imnotteixeira, line 158: I didn't mean to remove these, only the "returning promise" ones :/
These actually make sense

test/offer_schema.js
```diff
@@ -187,7 +188,6 @@ describe("# Offer Schema tests", () => {
                 test("Above maximum throws error", async () => {
                     const submitted_technologies = [];
                     for (let i = 0; i < MAX_TECHNOLOGIES + 1; ++i) {
-                        // Preventing interference from duplicate error
```
imnotteixeira, line 190: Add it back

test/offer_schema.js
```diff
@@ -208,7 +208,6 @@ describe("# Offer Schema tests", () => {
                 test("Inside the range does not throw error", async () => {
                     const submitted_technologies = [];
                     for (let i = 0; i < MIN_TECHNOLOGIES + 1; ++i) {
-                        // Preventing interference from duplicate error
```
imnotteixeira, line 211: Add it back

test/offer_schema.js
```diff
@@ -362,7 +361,6 @@ describe("# Offer Schema tests", () => {
             const field_to_insert = FieldTypes[0];
             const submitted_fields = [];
             for (let i = 0; i < MIN_FIELDS; ++i) {
-                // Preventing interference from duplicate error
```
imnotteixeira, line 365: Add it back

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #117, opened 2021-02-05
Refactor OR* Express Middleware

### LINKED ISSUE #116: Refactor express OR middleware
Currently we have middleware like `isAdminOrGod` or even `isCompanyOrAdminOrGod`.

We should only have the primitives (such as `isGod`, `isAdmin` etc...) and have an `or` function that would compose them to achieve the same effect without creating multiple combinations forever.

After trying to develop a function that would seamlessly integrate with the current validators, I realized that it's really hard, since validators are calling res.status().json() on error, in order to return error responses. When running a function that checks for multiple middleware, we would need to mock the res calls so that a validator would not stop the execution of others.

An easier solution is to refactor the current validators to throw errors (instead of instantly calling res functions). Then, the error handler app middleware (executed on every request) would catch those, and call the res methods there, ensuring that a proper response is given.

In order to know which response to issue, it would require an errors array (or a single error string), a status code, and an error_code, and it would call `res.status(status).json(buildErrorResponse(error_code, errors))`

This way, the `or` util would harness the exceptions to know which middleware failed and act accordingly.

Another issue is what would the or error message be? Well, it will only error out if all validators fail (since it's an OR operation), so it would have **n** error messages, n being the number of given validators. However, it does not make sense to return a response with "must be admin" and "must be company", for example if we are actually executing an OR between them.

We could wrap the response in an or field like this:
```
{
  error_code: ?,
  errors: [
    {or: ["must-be-admin", "must-be-company"]}
  ]
}
```
but then the structure would be different from the standard.

So, we need to have a standard response, by having an `error_code`, an `errors` array where each entry is an object with a `msg` field. However, we can also add some fields, which can give more context, something like the idea above with the or field, showing the possibilities.

Considering all this, the `or` function's signature should be the following:

```js
const or = ([...middleware], {error_code = DEFAULT_ERROR_CODE, msg = DEFAULT_ERROR_MSG}, status_code = DEFAULT_STATUS_CODE) => async (req, res, next) => {}
```

This ensures that we return a standard response, and have some defaults, in case the user does not provide them.

imnotteixeira:
Closes #116 

Throwing inside middleware does not work in Express v4 for async functions, so we can just return the error inside the next() call. In the default error handler, if it receives an APIError, it will send it, otherwise sends an "Unknown error" as current behavior


src/api/middleware/utils.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+const or = (
+    [...middlewares],
+    {
+        error_code = DEFAULT_ERROR_CODE,
+        msg = DEFAULT_ERROR_MSG,
+        status_code = DEFAULT_STATUS_CODE
+    } = {}
+) => async (initialReq, res, next) => {
+    let success = false;
```
DoStini, line 29:
```suggestion
```

src/api/middleware/utils.js
```diff
@@ -0,0 +1,56 @@
...
+    const errors = [];
+    for (const middleware of middlewares) {
+        const req = lodash.cloneDeep(initialReq);
```
DoStini, line 32: Maybe we should use this lib in other places that use `JSON.parse(stringify)`, since this seems to be more efficient

src/api/middleware/utils.js
```diff
@@ -0,0 +1,56 @@
...
+        if (success) return next();
+        try {
+            await middleware(req, res, (error) => {
+                if (error) errors.push(error);
+                else success = true;
+            });
+        } catch (err) {
+            errors.push(err);
```
DoStini, line 40:
```suggestion
            errors.push(err.toObject());
```

src/api/routes/example.js
```diff
@@ -22,10 +23,21 @@ module.exports = (app) => {
         });
     });
 
+    const p1 = async (req, res, next) => {
+        await Promise.resolve();
+        if (req.query.p !== "1") throw new APIError(HTTPStatus.BAD_GATEWAY, ErrorTypes.FORBIDDEN, "MUST BE 1");
```
DoStini, line 28: Not understanding why we use `throw` here and in `p2` `next(new Error)`

test/unit/utils.js
```diff
@@ -0,0 +1,120 @@
+const { APIError } = require("../../src/api/middleware/errorHandler");
+const { or, DEFAULT_ERROR_CODE, DEFAULT_STATUS_CODE, DEFAULT_ERROR_MSG } = require("../../src/api/middleware/utils");
+
+describe("Middleware utils", () => {
+    test("Should pass with all functions not erroring", async () => {
```
DoStini, line 5:
```suggestion
    test("Should pass with all functions without errors" async () => {
```

test/unit/utils.js
```diff
@@ -0,0 +1,120 @@
...
+        const m2 = (req, res, next) => next();
+        const m3 = (req, res, next) => next();
+
+        const nextMock = jest.fn();
+        await or([m1, m2, m3])({}, {}, nextMock);
+
+        expect(nextMock).toHaveBeenCalledWith();
+    });
+
+    test("Should pass with only one function not erroring", async () => {
```
DoStini, line 16:
```suggestion
    test("Should pass with only one function without errors", async () => {
```

test/unit/utils.js
```diff
@@ -0,0 +1,120 @@
...
+            throw new APIError(400, 1, "error_message", { test: 1 });
+        };
+
+        const nextMock = jest.fn();
+        await or([m1, m2, m3])({}, {}, nextMock);
+
+        expect(nextMock).toHaveBeenCalledWith();
+    });
+
+    test("Should pass with only one function not erroring", async () => {
```
DoStini, line 29:
```suggestion
    test("Should pass with only one function without error", async () => {
```

test/unit/utils.js
```diff
@@ -0,0 +1,120 @@
...
+        };
+        const m3 = (req, res, next) => next();
+
+        const nextMock = jest.fn();
+        await or([m1, m2, m3])({}, {}, nextMock);
+
+        expect(nextMock).toHaveBeenCalledWith();
+    });
+
+    test("Should fail if every middleware erroring", async () => {
```
DoStini, line 42:
```suggestion
    test("Should fail if every middleware has errors", async () => {
```

src/api/middleware/utils.js
```diff
@@ -0,0 +1,56 @@
...
+ *  - msg: the message in case of error (default: ValidationReasons.UNKNOWN)
+ *  - status_code: The status used in the HTTP Response in case of error (default: BAD_REQUEST (400))
+ */
+const or = (
+    [...middlewares],
+    {
+        error_code = DEFAULT_ERROR_CODE,
+        msg = DEFAULT_ERROR_MSG,
+        status_code = DEFAULT_STATUS_CODE
+    } = {}
```
DoStini, line 27: Why use here this method instead of `{error_code = DEFAULT, ...params}` like in offer creations? We could also change offer creation to this, if this is better

DoStini (changes requested):

src/api/middleware/utils.js
```diff
@@ -0,0 +1,56 @@
...
+        if (success) return next();
+        try {
+            await middleware(req, res, (error) => {
+                if (error) errors.push(error);
+                else success = true;
+            });
+        } catch (err) {
+            errors.push(err);
+        }
+    }
```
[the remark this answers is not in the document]
imnotteixeira: Middleware can fail by calling next() with an error argument, or unexpectedly with an exception, hence the need for both cases

src/api/middleware/utils.js
```diff
@@ -0,0 +1,56 @@
...
+    } = {}
+) => async (initialReq, res, next) => {
+    let success = false;
+    const errors = [];
+    for (const middleware of middlewares) {
+        const req = lodash.cloneDeep(initialReq);
+        if (success) return next();
+        try {
+            await middleware(req, res, (error) => {
+                if (error) errors.push(error);
```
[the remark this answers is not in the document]
imnotteixeira: Errors might not be `APIError` so we cannot call .toObject() on every error. On the current version, that would be equivalent, but I'm thinking we might want to only show errors as is if they are APIError instances, and unknown errors otherwise, to protect leaks.

In this case, we need to first check if the error is an APIError, and in that case push `err.toObject()`, otherwise push a generic unknown error .toObject so that db errors or similar end up leaking.

src/api/middleware/utils.js
imnotteixeira: They don't have the same behavior, since JSON.parse(JSON.stringify(obj)) will not maintain some types such as Dates, or custom class instances or event functions. So it might actually be harder than simply replacing to execute that refactor. Since that is not on the scope of this PR, I won't do it here.

You can create a separate issue to propose that

src/api/middleware/utils.js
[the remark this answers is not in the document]
DoStini: I see

src/api/routes/example.js
imnotteixeira: It was to simulate that both work and we end up with both errors being caught correctly. In earlier versions of express, throwing inside async functions would make it hang (due to unresolved promises) and the request would not even finish. Now it does, responding with unknown error or the specific APIError, depending on what is thrown.

This is useful in cases where middleware has DB calls, that can throw, but we don't want those errors to leak, but at the same time we want the request to fail, and not hang.

src/api/middleware/utils.js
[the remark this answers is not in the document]
DoStini: I don't understand then why calling map.((e)=>e.toObject()) worked

src/api/middleware/utils.js
imnotteixeira: I'd say they are equivalent. However, in generateTestOffer, the return is an object base on the input directly, so we can save some lines. Here, a clearer separation between arguments and function body is more appropriate, I think, and we would not save any lines by choosing another approach, since we'd have to do the restructuring inside the function body

src/api/middleware/utils.js
DoStini: Yes, I was just pointing out something I noticed

src/api/middleware/utils.js
```diff
@@ -0,0 +1,56 @@
...
+                else success = true;
```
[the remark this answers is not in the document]
imnotteixeira: That would be great.... If that wasn't a lambda function :P

I had that at first, but returning there won't return on the outer function, so it will continue the execution. Having a success flag is necessary to pass the information to the outer function that it should stop executing

src/api/middleware/utils.js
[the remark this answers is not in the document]
imnotteixeira: That's the thing, it won't ahaha

I am making the change now, and will call the toObject inside the for loop, knowing that we'll only deal with APIErrors there. If they are not APIErrors, they will be converted, and toObject can be safely called

src/api/middleware/utils.js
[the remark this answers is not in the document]
DoStini: Oh yeah, right! I totally missed that. Maybe after the try then?

src/api/middleware/utils.js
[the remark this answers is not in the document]
DoStini: Oh ok, I didn't realize it in the previous reply

src/api/routes/example.js
DoStini: Oh I see. That's great, I think I had noticed requests hanging because of async throws. Great that it's solved!

src/api/middleware/utils.js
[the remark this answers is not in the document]
imnotteixeira: Yep, simplified a bit since we don't need the check on the last lines of the function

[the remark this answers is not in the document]
DoStini: Great

DoStini (dismissed):
Everything seems good!

DoStini (approved):

PR #119, opened 2021-02-07
Add verification for jobMaxDuration greater than jobMinDuration on requests level

BrunoRosendo:
Added a new custom validator to the offers creation endpoint that verifies that jobMaxDuration > jobMinDuration, if they are present

### DIFF
--- a/src/api/middleware/validators/offer.js
+++ b/src/api/middleware/validators/offer.js
@@ -53,7 +53,20 @@ const create = useExpressValidators([
 
     body("jobMaxDuration", ValidationReasons.DEFAULT)
         .optional()
-        .isInt().withMessage(ValidationReasons.INT),
+        .isInt().withMessage(ValidationReasons.INT)
+        .custom((jobMaxDuration, { req }) => {
+            try {
+                const { jobMinDuration } = req.body;
+
+                if (jobMinDuration >= jobMaxDuration) {
+                    throw new Error(ValidationReasons.MUST_BE_AFTER("jobMinDuration"));
+                }
+            } catch (err) {
+                console.error(err);
+                throw err;
+            }
+            return true;
+        }),
 
     body("jobStartDate", ValidationReasons.DEFAULT)
         .optional()
--- a/src/api/middleware/validators/offer.js
+++ b/src/api/middleware/validators/offer.js
@@ -18,6 +18,27 @@ const {
     OFFER_EDIT_GRACE_PERIOD_HOURS,
 } = require("../../../models/constants/TimeConstants");
 
+const jobMaxDurationGreaterOrEqualThanJobMinDuration = (jobMaxDuration, { req }) => {
+    try {
+        const { jobMinDuration } = req.body;
+
+        // jobMinDuration is required if jobMaxDuration was specified
+        if (!jobMinDuration) {
+            throw new Error(ValidationReasons.JOB_MIN_DURATION_NOT_SPECIFIED);
+        }
+
+        if (jobMinDuration > jobMaxDuration) {
+            throw new Error(ValidationReasons.MUST_BE_AFTER("jobMinDuration"));
+        }
+
+        return true;
+
+    } catch (err) {
+        console.error(err);
+        throw err;
+    }
+};
+
 const create = useExpressValidators([
     body("title", ValidationReasons.DEFAULT)
         .exists().withMessage(ValidationReasons.REQUIRED).bail()
@@ -54,19 +75,7 @@ const create = useExpressValidators([
     body("jobMaxDuration", ValidationReasons.DEFAULT)
         .optional()
         .isInt().withMessage(ValidationReasons.INT)
-        .custom((jobMaxDuration, { req }) => {
-            try {
-                const { jobMinDuration } = req.body;
-
-                if (jobMinDuration >= jobMaxDuration) {
-                    throw new Error(ValidationReasons.MUST_BE_AFTER("jobMinDuration"));
-                }
-            } catch (err) {
-                console.error(err);
-                throw err;
-            }
-            return true;
-        }),
+        .custom(jobMaxDurationGreaterOrEqualThanJobMinDuration),
 
     body("jobStartDate", ValidationReasons.DEFAULT)
         .optional()
@@ -159,7 +168,7 @@ const jobMinDurationEditable = async (jobMinDurationCandidate, { req }) => {
         const { jobMaxDuration: jobMaxDurationCandidate } = req.body;
 
         // If the new publishMinDuration is after the new publishMaxDuration, the verification will be done in publishMaxDurationEditable
-        if (jobMinDurationCandidate >= offer.jobMaxDuration.toString() &&
+        if (jobMinDurationCandidate > offer.jobMaxDuration.toString() &&
             !jobMaxDurationCandidate) {
 
             // end date is earlier than publish date, error!
@@ -179,7 +188,7 @@ const jobMaxDurationEditable = async (jobMaxDurationCandidate, { req }) => {
         const { jobMinDuration: jobMinDurationCandidate } = req.body;
 
         const jobMinDuration = jobMinDurationCandidate || offer.jobMinDuration.toString();
-        if (jobMinDuration >= jobMaxDurationCandidate) {
+        if (jobMinDuration > jobMaxDurationCandidate) {
 
             throw new Error(ValidationReasons.MUST_BE_AFTER("jobMinDuration"));
         }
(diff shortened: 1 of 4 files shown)


src/models/Offer.js
```diff
@@ -116,7 +116,7 @@ function validateEndDate(value) {
 
 // jobMaxDuration must be larger than jobMinDuration
 function validateJobMaxDuration(value) {
-    return value >= this.jobMinDuration;
+    return value > this.jobMinDuration;
```
BrunoRosendo, line 119: I changed this to be coherent with the validators on the router layer. Before this change, an offer with equal jobMaxDuration and jobMinDuration would be accepted in the model and rejected in the router

imnotteixeira: We should keep the = on both sides, since that's a special case for offers that have a strict duration

e.g., for offers that strictly last 3 months, both `jobMinDuration` and `jobMaxDuration` will be 3

src/api/middleware/validators/offer.js
```diff
@@ -53,7 +53,20 @@ const create = useExpressValidators([
...
+
+                if (jobMinDuration >= jobMaxDuration) {
+                    throw new Error(ValidationReasons.MUST_BE_AFTER("jobMinDuration"));
+                }
+            } catch (err) {
+                console.error(err);
+                throw err;
+            }
+            return true;
+        }),
```
imnotteixeira, line 69:
```suggestion
const jobMaxDurationGreaterOrEqualThanJobMinDuration = (jobMaxDuration, { req }) => {
            try {
                const { jobMinDuration } = req.body;

                if (jobMinDuration >= jobMaxDuration) {
                    throw new Error(ValidationReasons.MUST_BE_AFTER("jobMinDuration"));
                }
            } catch (err) {
                console.error(err);
                throw err;
            }
            return true;
        }


...
        .custom(jobMaxDurationGreaterOrEqualThanJobMinDuration),
```

Move this to a function so that we can name the validation and make it clearer

src/api/middleware/validators/offer.js
```diff
@@ -53,7 +53,20 @@ const create = useExpressValidators([
 
     body("jobMaxDuration", ValidationReasons.DEFAULT)
         .optional()
-        .isInt().withMessage(ValidationReasons.INT),
+        .isInt().withMessage(ValidationReasons.INT)
+        .custom((jobMaxDuration, { req }) => {
+            try {
+                const { jobMinDuration } = req.body;
+
+                if (jobMinDuration >= jobMaxDuration) {
```
imnotteixeira, line 61: You need to also check if jobMinDuration is defined before comparing. If it is not defined (since it's optional) this validation should pass

Add `if(!jobMinDuration) return true;` before that check

src/api/middleware/validators/offer.js
```diff
@@ -53,7 +53,20 @@ const create = useExpressValidators([
...
+                    throw new Error(ValidationReasons.MUST_BE_AFTER("jobMinDuration"));
+                }
+            } catch (err) {
+                console.error(err);
+                throw err;
+            }
```
imnotteixeira, line 67: This is an unnecessary wrap in try-catch, since you are throwing the only possible error.

test/end-to-end/offer.js
```diff
@@ -479,6 +479,42 @@ describe("Offer endpoint tests", () => {
...
+                    ownerName: test_company.name,
+                });
+
+                const res = await request()
+                    .post("/offers/new")
+                    .send(withGodToken(offer_params));
+
+                expect(res.status).toBe(HTTPStatus.OK);
+            });
+        });
```
imnotteixeira, line 517: add a test for when jobMinDuration is not provided (it should pass)

imnotteixeira (changes requested):

test/end-to-end/offer.js
BrunoRosendo: I want to check something out.
```
jobMinDuration: {
    type: Number,
    required: function() {
        // jobMinDuration is required if jobMaxDuration was specified
        return !!this.jobMaxDuration;
    },
},
```
As you can see, in the model validator there's a comment (and function) saying that the jobMinDuration is required in case jobMaxDuration was specified, which goes against your current request.
This way, I think the scenario would be (based on the current model):

- jobMinDuration and jobMaxDuration both specified  => pass
- jobMinDuration unspecified and jobMaxDuration specified => fail
- jobMinDuration specified and jobMaxDuration unspecified => pass
- jobMinDuration and jobMaxDuation both unspecified => pass

Is this intended or do you want me to change something?

imnotteixeira: Yep, sorry, I was thinking about the edit scenario, where jobMinDuration can be omitted and it will use the one currently in the offer.

Your bulletpoints are correct.

imnotteixeira:
Just had a 5 min confusing when reading the PR title. For a moment I thought that we were limiting the job duration (which doesn't make sense), and I didn't recall having that issue either. 

So 2 things: 
* Create PRs with meaningful titles (instead of the automatic from the branch name) - in this case something like "Add  verification for jobMaxDuration greater than jobMinDuration on requests level"
* Always link the relevant issue(s) in the description with some specific keyword ("Closes #x" for auto closing, or "Related to #x" for just mentioning it") This allows reviewers (an anyone checking the PR in the future) to have context on why this was opened

Bonus: On the description, try to _describe_ briefly what the solution is based on (what did you do) - in this case you added a new custom validator to the offers creation endpoint that verifies that jobaxDuration > jobMinDuration, if they are present

src/api/middleware/validators/offer.js
```diff
@@ -18,6 +18,27 @@ const {
...
 } = require("../../../models/constants/TimeConstants");
 
+const jobMaxDurationGreaterOrEqualThanJobMinDuration = (jobMaxDuration, { req }) => {
+    try {
+        const { jobMinDuration } = req.body;
+
+        // jobMinDuration is required if jobMaxDuration was specified
+        if (!jobMinDuration) {
+            throw new Error(ValidationReasons.JOB_MIN_DURATION_NOT_SPECIFIED);
+        }
```
imnotteixeira, line 28: create a "mustSpecifyJobMinDurationIfJobMaxDurationSpecified" custom validator and put this logic there instead, calling it before this one, since that is another rule and it will become clearer that way

imnotteixeira: By doing that, you can remove the trycatch wrapper, and this function will get really small :)

src/api/middleware/validators/offer.js
```diff
@@ -146,7 +168,7 @@ const jobMinDurationEditable = async (jobMinDurationCandidate, { req }) => {
         const { jobMaxDuration: jobMaxDurationCandidate } = req.body;
 
         // If the new publishMinDuration is after the new publishMaxDuration, the verification will be done in publishMaxDurationEditable
-        if (jobMinDurationCandidate >= offer.jobMaxDuration.toString() &&
+        if (jobMinDurationCandidate > offer.jobMaxDuration.toString() &&
```
imnotteixeira, line 171: I just noticed that this would fail for offers that don't have a jobMaxDuration (possible since it is optional) since undefined.toString() throws an error

Make it 

```js
if ((offer.jobMaxDuration && jobMinDurationCandidate > offer.jobMaxDuration.toString()) && !jobMaxDurationCandidate)
```

src/api/middleware/validators/offer.js
```diff
@@ -166,7 +188,7 @@ const jobMaxDurationEditable = async (jobMaxDurationCandidate, { req }) => {
         const { jobMinDuration: jobMinDurationCandidate } = req.body;
 
         const jobMinDuration = jobMinDurationCandidate || offer.jobMinDuration.toString();
-        if (jobMinDuration >= jobMaxDurationCandidate) {
+        if (jobMinDuration > jobMaxDurationCandidate) {
```
imnotteixeira, line 191: Add a check before this `if (!jobMinDuration) throw...` since jobMinDuration is required if jobMaxDuration is present

imnotteixeira (changes requested):

src/api/middleware/validators/offer.js
DoStini: Yes, I didn't thought of this problem. 

In that case, maybe we should throw an error saying something like "Must add minimum offer duration" or just assume it is 0, in that case it should be accepted.

imnotteixeira: Since there is no default 0 on the model side, let's not assume 0 and just throw the error

BrunoRosendo: @DoStini I reused the same error I was using in the creation validator (JOB_MIN_DURATION_NOT_SPECIFIED)

imnotteixeira (approved):

PR #120, opened 2021-02-08
Feature/application finish

### LINKED ISSUE #91: Add "complete registration" endpoint for companies
In order to complete #6, companies need to be able to provide their contact information when applying to nijobs.

Discussions were already started regarding the validation and structure of the field here: #51

Additionally, in order to complete their profile, they need a bio and a logo. This endpoint should receive those 3 and complete the Company instance that was created when approving the application.

In order to make this process easier on the frontend, I'd suggest adding a boolean field to the Company meaning that it has completed the registration. We would use that field to show the "Complete registration widget" but also to prevent companies from creating offers before completing the registration (purely backend limitation, since the frontend won't show the button before the company is fully registered)  

NIAEFEUP/Website-NIAEFEUP#199 is a good starting point on the logo storing feature.

GuilhermeJSilva:
Closes #91
Two new endpoints are added:
- List companies with pagination (added to check the end result of the finish endpoint)
   - Receives _limit_ and _offset_
- Finish the company profile
   - Only a company representative can use the endpoint
   - Companies which have already finished their profile can not access the endpoint (flag added to the company model)
   - The endpoint receives a request in `multipart/form-data`
   - The fields accepted are: _logo_ (png or jpeg), bio (string), contacts (array)
   - After all fields are validated the image is saved locally
   - If a CLOUDINARY_URL is defined as an env variable, then the file is going to be uploaded using the cloudinary NodeJS SDK (after it is erased locally)
   - After a successful upload (either local or to the cloud) the bio, contacts and logo path are saved to the database


src/api/routes/company.js
```diff
@@ -0,0 +1,60 @@
...
+                const companyService = new CompanyService();
+                const { bio, contacts } = req.body;
+                let logo = `static/${req.file.filename}`;
+                if (req.file.url)
+                    logo = req.file.url;
+                const company_id = req.user.company;
+                await companyService.changeAttributes(company_id, { bio, contacts, logo, finished: true });
+                return res.json({});
+            } catch (err) {
+                return next(err);
```
imnotteixeira, line 38: Add a console.error(err) as well

src/services/company.js
```diff
@@ -15,6 +12,49 @@ class CompanyService {
...
+     */
+    findById(company_id) {
+        return Company.findById(company_id);
+    }
+
+
+    /**
+     * Changes the attributes of a company
+     * @param {*} company_id id of the company
+     * @param {*} attributes attributes to change in company
```
imnotteixeira, line 52: Make it clearer that this should be an object

test/end-to-end/company.js
```diff
@@ -0,0 +1,212 @@
...
+const ValidationReasons = require("../../src/api/middleware/validators/validationReasons");
+const CompanyConstants = require("../../src/models/constants/Company");
+const fs = require("fs");
+const path = require("path");
+
+const getCompanies = async (options) =>
+    [
+        ...(await Company
+            .find(options)
+            .exec())
```
imnotteixeira, line 14: no .exec needed. await Company.find() returns the results already

test/end-to-end/company.js
```diff
@@ -0,0 +1,212 @@
...
+    ]
+        .map((c) => {
+            const obj = c.toObject();
+            obj._id = obj._id.toString();
+            return obj;
+        });
```
imnotteixeira, line 20:
```suggestion
        .map((company) => ({
            ...company.toObject(),
            _id: company._id.toString()
        }));
```

test/end-to-end/company.js
```diff
@@ -0,0 +1,212 @@
...
+                password: "password123",
+            };
+
+            const company_data = {
+                name: "Company Ltd"
+            };
+
+            beforeEach(async () => {
+                await Company.deleteMany({});
+                const c = await Company.create({ name: company_data.name });
```
imnotteixeira, line 71:
```suggestion
                const test_company = await Company.create({ name: company_data.name });
```

test/end-to-end/company.js
```diff
@@ -0,0 +1,212 @@
...
+                    .attach("logo", "test/data/logo-niaefeup.png")
+                    .field("bio", "A very interesting and compelling bio")
+                    .field("contacts", ["contact1", "contact2"])
+                    .expect(HTTPStatus.OK);
+
+                const c = [... await Company.find({}).exec()][0];
+                expect([...c.contacts]).toEqual(["contact1", "contact2"]);
+                expect(c.finished).toBe(true);
+                expect(c.bio).toBe("A very interesting and compelling bio");
+                const filename = path.join(__dirname, `../../${c.logo.replace("static", "public")}`);
```
imnotteixeira, line 94: Why is this replace needed? What's going on between static vs public?

src/api/routes/company.js
```diff
@@ -0,0 +1,60 @@
...
+const fileMiddleware  = require("../middleware/files");
+const { authRequired } = require("../middleware/auth");
+
+module.exports = (app) => {
+    app.use("/company", router);
+
+    /**
+     * Creates a new Company Application
+     */
+    router.post("/finish",
```
imnotteixeira, line 18:
```suggestion
    router.post("/application/finish",
```

src/models/Company.js
```diff
@@ -12,11 +12,20 @@ const CompanySchema = new Schema({
...
     contacts: {
         type: [String],
+        minlength: CompanyConstants.contacts.min_length,
+        maxlength: CompanyConstants.contacts.max_length,
     },
     bio: {
         type: String,
         maxlength: CompanyConstants.bio.max_length,
     },
+    finished: {
```
imnotteixeira, line 22:
```suggestion
    hasFinishedRegistration: {
```

src/config/multer.js
```diff
@@ -0,0 +1,15 @@
...
+
+const storage =  multer.memoryStorage();
+const limits = { fileSize: 5000000 };
+const fileFilter = (req, file, cb) => {
+    const fileTypes = /png|jpeg|jpg/;
+    const mimeType = fileTypes.test(file.mimetype);
+    if (mimeType) {
+        cb(null, true);
+    } else {
+        cb(new Error(ValidationReasons.IMAGE_FORMAT));
```
imnotteixeira, line 12: Might need to change this into an APIError instance, after #117 to make it appear on the response, otherwise a generic error will be shown instead

src/api/middleware/files.js
```diff
@@ -0,0 +1,110 @@
...
+            .json({
+                error_code: ErrorTypes.FILE_ERROR,
+                errors: [
+                    {
+                        location: "body",
+                        param: req.file.fieldname,
+                        msg: ValidationReasons.FAILED_SAVE
+                    }
+                ]
+            });
```
imnotteixeira, line 68: same thing (#117)

src/api/middleware/files.js
```diff
@@ -0,0 +1,110 @@
...
```
imnotteixeira, line 102: same thing (#117)

src/api/middleware/files.js
```diff
@@ -0,0 +1,110 @@
...
+                .status(HTTPStatus.UNPROCESSABLE_ENTITY)
+                .json({
+                    error_code: ErrorTypes.VALIDATION_ERROR,
+                    errors: [
+                        {
+                            location: "body",
+                            param,
+                            msg: message
+                        }]
+                });
```
imnotteixeira, line 39: same thing (#117)

src/api/middleware/files.js
```diff
@@ -0,0 +1,110 @@
...
+const env = require("../../config/env");
+const ValidationReasons = require("./validators/validationReasons");
+
+
+const save_folder = path.join(__dirname, "../../../public/");
+if (!fs.existsSync(save_folder)) fs.mkdirSync(save_folder);
+
+const parseError = (message) => message.toLowerCase().replace(" ", "-");
+
+const single = (field_name) => (req, res, next) => {
```
imnotteixeira, line 18: This name is unclear, what does this function do?

src/api/middleware/files.js
```diff
@@ -0,0 +1,110 @@
...
+const { MulterError } = require("multer");
+const multerConfig = require("../../config/multer");
+const { ErrorTypes } = require("./errorHandler");
+const cloudinary = require("cloudinary").v2;
+const env = require("../../config/env");
+const ValidationReasons = require("./validators/validationReasons");
+
+
+const save_folder = path.join(__dirname, "../../../public/");
+if (!fs.existsSync(save_folder)) fs.mkdirSync(save_folder);
```
imnotteixeira, line 14: I think this should be on the loader (perhaps the ./static one) instead of being here dangling and possibly be called multiple times (even though it shouldn't be a problem, it make more sense to be in a loader)

src/api/middleware/files.js
```diff
@@ -0,0 +1,110 @@
...
+                            msg: message
+                        }]
+                });
+        } else {
+            return next();
+        }
+    });
+};
+
+const save = async (req, res, next) => {
```
imnotteixeira, line 46:
```suggestion
const saveLocal = async (req, res, next) => {
```
To make a clearer distinction between cloud and this one

src/api/middleware/files.js
```diff
@@ -0,0 +1,110 @@
+const HTTPStatus = require("http-status-codes");
+const fs = require("fs");
+const path = require("path");
+const util = require("util");
+const { MulterError } = require("multer");
+const multerConfig = require("../../config/multer");
+const { ErrorTypes } = require("./errorHandler");
+const cloudinary = require("cloudinary").v2;
+const env = require("../../config/env");
```
imnotteixeira, line 9:
```suggestion
const config = require("../../config/env");
```
To follow the standard in other files

src/api/middleware/files.js
imnotteixeira: In order to access the save_folder constant in both places, you can declare it in the config (env stuff) and require it from both places

src/api/middleware/files.js
```diff
@@ -0,0 +1,110 @@
...
+
+const save_folder = path.join(__dirname, "../../../public/");
+if (!fs.existsSync(save_folder)) fs.mkdirSync(save_folder);
+
+const parseError = (message) => message.toLowerCase().replace(" ", "-");
+
+const single = (field_name) => (req, res, next) => {
+    const upload = multerConfig.single(field_name);
+    upload(req, res, (error) => {
+        if (error || !req.file) {
```
imnotteixeira, line 21: Place the `if(!req.file)` check before the upload call to make it fail even before this.

Then, you will only need an if(error) here, no more nested ifs. Also, the message variable can be `                const message = error instanceof MulterError ?
                    parseError(error.message) : error.message;`

src/services/company.js
```diff
@@ -15,6 +12,49 @@ class CompanyService {
...
+     */
+    async findAll(limit, offset) {
+
+        const totalDocCount = await Company.estimatedDocumentCount();
+
+        return {
+            totalDocCount,
+            companies:
+                [...(await Company.find({})
+                    .sort({ Name: "desc" })
```
imnotteixeira, line 31:
```suggestion
                    .sort({ name: "desc" })
```

imnotteixeira: Why is descending order the default? If we're having a default, it should be ascending :thinking:

imnotteixeira (changes requested):
I didn't fully understand why you included the "get all companies endpoint". It isn't really required, right? It was just a helpful addition for using this?

test/end-to-end/company.js
GuilhermeJSilva: All files uploaded are saved to the `public` folder but the route to access the files is `static/file_path`, maybe i should change the name of one of them

src/api/middleware/files.js
GuilhermeJSilva: The name mirrors the single function of `multer` (the package that processes `multipart/form-data`) and wraps it with error checking, the function is responsible for reading a **single** file from the request with the corresponding `filed_name`

src/api/middleware/files.js
GuilhermeJSilva: `req.file` is created by the `upload` function; checking it beforehand is impossible.

GuilhermeJSilva:
> I didn't fully understand why you included the "get all companies endpoint". It isn't really required, right? It was just a helpful addition for using this?

I did not see the endpoint to get the logged-in user's information; therefore, to check if the application was successful, I implemented the get-all companies endpoint, which I then left since it was a helpful addition. All suggestions are tackled (except for the ones which require #117), so please check the comments I made above and re-review.

src/models/Company.js
```diff
@@ -12,11 +12,20 @@ const CompanySchema = new Schema({
...
+    hasFinishedRegistration: {
+        type: Boolean,
+        default: false
+    },
+    logo: {
+        type: String,
+    }
 });
 
 // Update offers from this company on name change
```
imnotteixeira, line 31: Please make sure this hook is working. If not, either change the hook or use the findOneAndUpdate in the service `changeAttributes` function

src/config/env.js
```diff
@@ -28,4 +29,8 @@ module.exports = Object.freeze({
     // Mail
     mail_from: process.env.MAIL_FROM,
     mail_from_password: process.env.MAIL_FROM_PASSWORD,
+
+    // File upload
+    cloudinary_url: process.env.CLOUDINARY_URL,
+    upload_folder: path.join(path.join(__dirname, "../.."), process.env.UPLOAD_FOLDER || "public")
```
imnotteixeira, line 35:
```suggestion
    upload_folder: path.join(path.join(__dirname, "../.."), process.env.UPLOAD_FOLDER || "static")
```

Making the default folder name be "static" probably clears some confusion. Also, does it need to exist beforehand, or will it create if not exist?

imnotteixeira (changes requested):

src/api/middleware/files.js
GuilhermeJSilva: that's it

src/config/env.js
GuilhermeJSilva: it will create if the folder does not exist

src/models/Company.js
GuilhermeJSilva: I can confirm that it does not trigger, so I changed the function used in `changeAttributes`

src/api/middleware/company.js
```diff
@@ -32,8 +32,23 @@ const verifyMaxConcurrentOffers = async (req, res, next) => {
...
+const profileNotComplete = async (req, res, next) => {
+    const company = await (new CompanyService()).findById(req.user.company);
+    if (company.finished) {
+        return res
+            .status(HTTPStatus.FORBIDDEN)
+            .json({
+                reason: ValidationReasons.REGISTRATION_FINISHED,
+                error_code: ErrorTypes.FORBIDDEN
+            });
+    }
```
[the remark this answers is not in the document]
GuilhermeJSilva: Fixed!

src/config/multer.js
GuilhermeJSilva: This error is caught in the function `parseSingleFile`; therefore, it doesn't need to change

src/api/routes/company.js
```diff
@@ -0,0 +1,60 @@
...
+
+const fileMiddleware  = require("../middleware/files");
+const { authRequired } = require("../middleware/auth");
+
+module.exports = (app) => {
+    app.use("/company", router);
+
+    /**
+     * Creates a new Company Application
+     */
```
imnotteixeira, line 18: The problem of comments... please update :P

imnotteixeira (changes requested):

src/api/middleware/files.js
```diff
@@ -0,0 +1,97 @@
...
+    } catch (err) {
+        return next(new APIError(
+            HTTPStatus.UNPROCESSABLE_ENTITY,
+            ErrorTypes.FILE_ERROR,
+            [{
+                location: "body",
+                param: req.file.fieldname,
+                msg: ValidationReasons.FAILED_SAVE
+            }]));
+    }
```
imnotteixeira, line 90: The file is only deleted if the upload goes well, should we delete it every time? It won't be used anyway...

imnotteixeira (approved):

Issue #123, opened 2021-02-19
Update the concurrent offers validator to account for nonoverlapping offers in a time period


BrunoRosendo:
Right now, the concurrentOffersNotExceeded function, which is used to validate offers in the route and model levels, only counts all offers in a time period for a specific owner. Update this function so it accounts for nonoverlapping offers in the new offer's time period.

For example:

```
|-----------|

       (new offer)
    |---------------|

                 |-----------|
```
In this case, there are only 2 concurrent offers at any given time but, right now, the function will count 3 anyways. A possible solution would be to parse the offers returned by the getOffersInTimePeriod function to count the maximum concurrent ones at any time.

imnotteixeira:
In the current model, this problem is hard since you have a high granularity on the offer times (in the order of seconds).

To achieve this goal you would need to verify the overlap for each second, and if the offer extends for the current maximum period (6 months), there are approx. 15552000 secs which is too much to calculate.

An alternative would be to lower the granularity (e.g. the publish dates would have a "day" granularity), but that would still be 180 days in the worst scenario.

The O(n) solution for this problem is as follows (I think):
1. Fetch all offers that could collide with the new one (the ones that intersect only)
2. Order all offers (including the new one) by publishDate descending
3. iterate the time slots (in this case each day) starting with the offer that has the earliest publishDate and count how many offers are active in that day
4. If some slot has more offers than the limit, stop. This way we can make the algorithm faster in most of the cases

This has a time complexity of **n*t**, *n* being the number of intersecting offers, and *t* the time until a limit surpassing is found, with a maximum t = 180 (for day granularity and current limit of 6 months - assuming 30d/month).

With this in mind, I am open to more suggestions on how to tackle this problem, with possibly some heuristics to further optimize it (maybe storing auxiliary values on offers, at the creation time?)

BrunoRosendo:
When I wrote the issue I thought about just checking every offer to see if they actually collide with one another (even though this is O(n2), the number of offers is fairly small). But after thinking a little more I understood that this would only reduce the problem and not completely fix it, since there are still cases where it would not work. I'm just writing this in case anyone wants to implement this in the meanwhile (since I believe it is better than nothing and it's very simple), until we implement a permanent solution like the one you described.

imnotteixeira:
> I thought about just checking every offer to see if they actually collide with one another

That would only increase the computation time without actually solving anything since they can collide at different times, still. That's not worth doing.

[...]

Issue #57, opened 2019-11-26
Restrict Offer creation by Company


miguelpduarte:
Add restrictions to a Company, so that it cannot have more than X Offers active at a certain time when creating an Offer.

PR #124, opened 2021-02-20
Hotfix for concurrent offers validator when nonoverlapping offers exist in the time period

BrunoRosendo:
Right now, the concurrentOffersNotExceeded function, which is used to validate offers in the route and model levels, only counts all offers in a time period for a specific owner. This PR tries to fix this

### DIFF
--- a/src/api/middleware/validators/validatorUtils.js
+++ b/src/api/middleware/validators/validatorUtils.js
@@ -73,18 +73,16 @@ const concurrentOffersNotExceeded = (OfferModel) => async (owner, publishDate, p
     startDates.sort();
     endDates.sort();
 
+    /**
+     * This algorithm is explained in https://github.com/NIAEFEUP/nijobs-be/issues/123#issuecomment-782272539
+     */
     let counter = 0, maxConcurrent = 0;
     let startIndex = 0, endIndex = 0;
     while (startIndex < offerNumber || endIndex < offerNumber) {
-        if (startIndex < offerNumber) {
-            if (endIndex >= offerNumber || startDates[startIndex] <= endDates[endIndex]) {
-                counter++;
-                startIndex++;
-                if (counter > maxConcurrent) maxConcurrent = counter;
-            } else {
-                counter--;
-                endIndex++;
-            }
+        if (startIndex < offerNumber && (endIndex >= offerNumber || startDates[startIndex] <= endDates[endIndex])) {
+            counter++;
+            startIndex++;
+            if (counter > maxConcurrent) maxConcurrent = counter;
         } else {
             counter--;
             endIndex++;
--- a/src/api/middleware/validators/validatorUtils.js
+++ b/src/api/middleware/validators/validatorUtils.js
@@ -68,18 +68,19 @@ const concurrentOffersNotExceeded = (OfferModel) => async (owner, publishDate, p
     if (offerNumber < CompanyConstants.offers.max_concurrent) return true;
     // otherwise, let's check how many are concurrent
 
-    const startDates = offersInTimePeriod.map((offer) => offer.publishDate);
-    const endDates = offersInTimePeriod.map((offer) => offer.publishEndDate);
-    startDates.sort();
-    endDates.sort();
+    // This algorithm is explained in https://github.com/NIAEFEUP/nijobs-be/issues/123#issuecomment-782272539
 
-    /**
-     * This algorithm is explained in https://github.com/NIAEFEUP/nijobs-be/issues/123#issuecomment-782272539
-     */
+    const offersByStart = offersInTimePeriod; // we won't need this array unmodified
+    const offersByEnd = [...offersInTimePeriod];
+    offersByStart.sort((offer1, offer2) => Date.parse(offer1.publishDate) - Date.parse(offer2.publishDate));
+    offersByEnd.sort((offer1, offer2) => Date.parse(offer1.publishEndDate) - Date.parse(offer2.publishEndDate));
+
     let counter = 0, maxConcurrent = 0;
     let startIndex = 0, endIndex = 0;
     while (startIndex < offerNumber || endIndex < offerNumber) {
-        if (startIndex < offerNumber && (endIndex >= offerNumber || startDates[startIndex] <= endDates[endIndex])) {
+        if (startIndex < offerNumber &&
+            (endIndex >= offerNumber || offersByStart[startIndex].publishDate <= offersByEnd[endIndex].publishEndDate)) {
+
             counter++;
             startIndex++;
             if (counter > maxConcurrent) maxConcurrent = counter;
--- a/src/api/middleware/validators/validatorUtils.js
+++ b/src/api/middleware/validators/validatorUtils.js
@@ -53,7 +53,8 @@ const isObjectId = (id) => {
 };
 
 /**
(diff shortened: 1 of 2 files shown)


src/api/middleware/validators/validatorUtils.js
```diff
@@ -61,10 +61,36 @@ const isObjectId = (id) => {
...
+                startIndex++;
+                if (counter > maxConcurrent) maxConcurrent = counter;
+            } else {
+                counter--;
+                endIndex++;
+            }
+        } else {
+            counter--;
+            endIndex++;
+        }
```
imnotteixeira, line 91: You can merge these in a single if/else to simplify since the else is the same. Additionally, I'd put a comment with the link to #123 before this function to provide an explanation of the algorithm

```suggestion
        if (startIndex < offerNumber && (endIndex >= offerNumber || startDates[startIndex] <= endDates[endIndex])) {
                counter++;
                startIndex++;
                if (counter > maxConcurrent) maxConcurrent = counter;      
        } else {
            counter--;
            endIndex++;
        }
```

test/end-to-end/offer.js
```diff
@@ -447,6 +447,56 @@ describe("Offer endpoint tests", () => {
...
+                    offer.owner = test_company._id;
+                    offer.ownerName = test_company.name;
+                });
+
+                await Offer.create(testOffers);
+            });
+
+            afterAll(async () => {
+                await Offer.deleteMany({});
+            });
```
imnotteixeira, line 480: Remove this. It shouldn't be required

test/end-to-end/offer.js
```diff
@@ -447,6 +447,56 @@ describe("Offer endpoint tests", () => {
             });
         });
 
+        describe("Creating an offer in a time period with nonoverlapping offers", () => {
```
imnotteixeira, line 450:
```suggestion
        describe("Creating an offer in a time period with more than `max_concurrent` overlapping offers, without exceeding the limit at any point", () => {
```

imnotteixeira (changes requested):

src/api/middleware/validators/validatorUtils.js
```diff
@@ -61,10 +61,36 @@ const isObjectId = (id) => {
  */
 const concurrentOffersNotExceeded = (OfferModel) => async (owner, publishDate, publishEndDate) => {
     // We need to pass the offer model in case we're inside an Offer instance
-    const concurrentOffers = await (new CompanyService())
+    const offersInTimePeriod = await (new CompanyService())
```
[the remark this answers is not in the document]
BrunoRosendo: Correct me if I'm wrong but I don't think that's necessary. I think that, since all those offers collide with the candidate offer, the number of concurrent offers will always be what I'm currently calculating plus one (hence the `<` instead of the `<=`.
Is there any counterexample that I'm not seeing?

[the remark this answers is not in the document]
imnotteixeira: Not necessarily, you can have 5 max concurrent offers at a given time, and add a new one that does not make it 6, that's why you need to include it in the algorithm:

```
-----XXX  <- New offer
XXXXXX
XXXXX
XXXXX
XXXXX
XXXXX
```

For the first check it's ok, since you only check for total intersections, but for the actual algorithm, it neds to take into account the new offer, and for that you might want to change the first section as well, to take the new offer into account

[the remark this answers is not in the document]
BrunoRosendo: I can include it to remove all doubts but, in that examples, the date arrays wouldn't include the bottom 4 offers, since we're only fetching offers which collide with the new offer from the database (not all of the company's offers)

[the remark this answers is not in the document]
imnotteixeira: Hmm, I see, you're also using the < only in the last check, so it should work. I could not find a counter example

src/api/middleware/validators/validatorUtils.js
```diff
@@ -61,10 +61,35 @@ const isObjectId = (id) => {
...
         .getOffersInTimePeriod(owner, publishDate, publishEndDate, OfferModel);
 
-    return concurrentOffers.length < CompanyConstants.offers.max_concurrent;
+    const offerNumber = offersInTimePeriod.length;
+    if (offerNumber < CompanyConstants.offers.max_concurrent) return true;
+    // otherwise, let's check how many are concurrent
+
+    // This algorithm is explained in https://github.com/NIAEFEUP/nijobs-be/issues/123#issuecomment-782272539
```
imnotteixeira, line 71: Move this comment to the function documentation above (after L60)

src/api/middleware/validators/validatorUtils.js
```diff
@@ -61,10 +61,35 @@ const isObjectId = (id) => {
...
+
+    const offersByStart = offersInTimePeriod; // we won't need this array unmodified
+    const offersByEnd = [...offersInTimePeriod];
+    offersByStart.sort((offer1, offer2) => Date.parse(offer1.publishDate) - Date.parse(offer2.publishDate));
+    offersByEnd.sort((offer1, offer2) => Date.parse(offer1.publishEndDate) - Date.parse(offer2.publishEndDate));
```
imnotteixeira, line 76: Create a `sortOffersByDateAscending` function before this function and just pass it to the .sort() to avoid duplicated code

src/api/middleware/validators/validatorUtils.js
```diff
@@ -61,10 +61,35 @@ const isObjectId = (id) => {
...
         .getOffersInTimePeriod(owner, publishDate, publishEndDate, OfferModel);
 
-    return concurrentOffers.length < CompanyConstants.offers.max_concurrent;
+    const offerNumber = offersInTimePeriod.length;
+    if (offerNumber < CompanyConstants.offers.max_concurrent) return true;
+    // otherwise, let's check how many are concurrent
+
+    // This algorithm is explained in https://github.com/NIAEFEUP/nijobs-be/issues/123#issuecomment-782272539
+
+    const offersByStart = offersInTimePeriod; // we won't need this array unmodified
```
imnotteixeira, line 73:
```suggestion
    const offersSortedByStart = offersInTimePeriod; // we won't need this array unmodified
```

src/api/middleware/validators/validatorUtils.js
```diff
@@ -61,10 +61,35 @@ const isObjectId = (id) => {
...
+    const offersByEnd = [...offersInTimePeriod];
```
imnotteixeira, line 74:
```suggestion
    const offersSortedByEnd = [...offersInTimePeriod];
```

imnotteixeira (changes requested):
Just some small tweaks before approving. Good job!

src/api/middleware/validators/validatorUtils.js
BrunoRosendo: Although similar, the functions aren't exactly the same (since we use the publishDate in one and publishEndDate in the other). Is it worth it making a function with a custom property as an argument?

imnotteixeira: Yep! Call it `sortOffersByFieldAscending` and you can use the `obj[arg]` notation to access the value of a varying arg in the object

imnotteixeira (approved):
Nice work, this will be really useful, since it finally solves the #57 problem :rocket:

PR #125, opened 2021-02-25
Apply the .current query to both text search and non-text search

imnotteixeira:
Currently, if a text query was supplied to offer search, it would not filter by `current` only, as it did when no text query was provided. This PR fixes that.


DoStini:
Are there tests for this issue?

imnotteixeira:
> Are there tests for this issue?

There are now :sunglasses:

DoStini (dismissed):

DoStini (approved):

PR #127, opened 2021-02-27
Changed queries for job duration filtering

imnotteixeira:
Fixes the advanced search related to job duration in offers.

Now it takes into account undefined/null values and it's also possible to get offers that intersect the limit, as long as they are valid on the correct "side" (e.g. an offer with job duration 3-6 months should appear if the max duration filter is set to 5, since it can (in theory) last for 3, 4 or 5 months as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,9 @@
 # niJobs - BackEnd
 
-[![Build Status](https://img.shields.io/travis/NIAEFEUP/nijobs-be/develop.svg?style=for-the-badge)](https://travis-ci.org/NIAEFEUP/nijobs-be)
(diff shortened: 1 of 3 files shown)


GuilhermeJSilva (approved):
LGTM!

GuilhermeJSilva (approved):
LGTM! The only thing that changed from the previous review is the removal of a dangling import, was this caught by the linter? I can't check due to the force pushes

imnotteixeira:
> LGTM! The only thing that changed from the previous review is the removal of a dangling import, was this caught by the linter? I can't check due to the force pushes

The only thing I did was rebase from develop to get the latest changes, since it is passing the checks, I assume everything is ok

PR #134, opened 2021-03-12
Use HTTPS in development mode

### LINKED ISSUE #133: HTTPS development server
After some issues in testing NIAEFEUP/nijobs-fe#80, I realized we would have a problem using Chrome to test frontend changes in Netlify (or using any kind of different backend for the frontend application).

The issue only affects chrome and basically boils down to it blocking cookies that have SameSite=Lax (which is the default) and which are passed to a different origin (in this case from netlify to our custom backend running locally - for me I was using ngrok, as per the nijobs-fe readme, but the problem would be the same no matter what tool I used, since the backend wouldn't be served through netlify, so the origin would be different)

The docs (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite) state that in order for cookies to be passed between different origins, they need to have SameSite=None. That's an easy fix: one just needs to add it to the cookie options in the session middleware. However, SameSite=None, only works with **secure** cookies. This is where things get interesting...

In order to use secure cookies, we need to use HTTPS. For the frontend, our setup already allows it, by simply providing HTTPS=true as an env var. By using ngrok, and setting the API hostname to the HTTPS version, we would have an HTTPS frontend calling a different (different origin) HTTPS backend, and since we have SameSite=None and Secure=true, everything works.
(For this to really work correctly, the `proxy` option must also be set to true in the session middleware - as it is being done in production only, currently)

But this raises the problem: what about normal development in localhost? It should still work after whatever fix we apply...
Well, if we do the above, when using the normal localhost:8087 server (without ngrok), things won't work since the HTTPS frontend will be calling an HTTP backend, which is not allowed.

This leads me to the main point: we should make the dev version of the Backend serve an HTTPS API, which would allow local development, by using HTTPS on both ends, and also making it possible to communicate with Netlify (or other different host with a frontend) via HTTPS and have no Cookie problem.

In order to achieve this, I found this Post (https://medium.com/@nitinpatel_20236/how-to-create-an-https-server-on-localhost-using-express-366435d61f28) which explains how to do it, and I think we can add the necessary config to a script and even make it part of the Docker-dev setup, to make it easier. It uses a self-signed certificate, so the work should be minimal.

The only missing piece is to find out if the `proxy` option being true when there is no proxy (localhost only) will cause any problem. I don't believe it will, but if it does, we'll have to create an env option for devs to toggle it when using localhost or ngrok (or another proxy)

imnotteixeira:
Closes #133 

Created a certs/ folder to handle certs generation and usage by the application

The certgen.sh script will generate the required cert and key for https (self-signed)

This https mode will only be used in dev, since it would break tests and prod doesn't require it, since it is being handled by the apache proxy


src/index.js
```diff
@@ -9,7 +10,17 @@ const startServer = async () => {
...
     // Running the application in test mode does not start listening because parallel tests would result in EADDRINUSE
     if (process.env.NODE_ENV !== "test") {
-        app.listen(config.port, (err) => {
+
+        let server = app;
+        if (process.env.NODE_ENV !== "production") {
+            const fs = require("fs");
+            const path = require("path");
+            const key = fs.readFileSync(path.join(__dirname, "../certs/key.pem"));
+            const cert = fs.readFileSync(path.join(__dirname, "../certs/cert.pem"));
```
joaomrcsmartins, line 19: This is not a remark as much as a doubt really, but why does the certificates retrieval need to be synchronous?

imnotteixeira: It has to be synchronous, since it's needed to start the server, but it's effectively the same as using await on the Promise version of fs. Actually, I can make these 2 run concurrently by having something like:

```js
const [key, cert] = await Promise.all([
    fs.promises.readFile(path.join(__dirname, "../certs/key.pem")),
    fs.promises.readFile(path.join(__dirname, "../certs/cert.pem"))
]);
```

Thanks for the callout!

src/loaders/express.js
```diff
@@ -31,12 +31,13 @@ module.exports = (app) => {
...
-        // Set to true only in prod. Necessary due to reverse proxy setup
+        // Necessary due to reverse proxy setup, but it won't affect anything if there is no proxy.
         // Check https://github.com/expressjs/session/issues/281 for more details
-        proxy: process.env.NODE_ENV === "production",
+        proxy: true,
 
         cookie: {
-            secure: process.env.NODE_ENV === "production",
+            secure: process.env.NODE_ENV !== "test",
+            sameSite: process.env.NODE_ENV === "production" ? "lax" : "none"
```
miguelpduarte, line 40: I think we can have `SameSite=Strict` for production environments. Lax does not work when we are sending requests to the back-end cross-origin though, so we would have to have that set as None specifically (it even caused some bugs from what I've heard).

certs/.gitignore
```diff
@@ -0,0 +1,3 @@
+cert.pem
+key.pem
+keytmp.pem
```
miguelpduarte, line 3: No newline at the end of file here.

README.md
```diff
@@ -109,6 +109,13 @@ So, you can run:
 
 This approach might be the least straightforward to set up but is the most flexible as you can freely and directly interact with the runtime of the application.
 
+#### HTTPS details
+In order for Cookies to work correctly, we needed to use HTTPS in development, so that we could use local servers to test frontend changes in netlify.
```
miguelpduarte, line 113: `s/needed/need` and `s/could/can`

README.md
```diff
@@ -109,6 +109,13 @@ So, you can run:
...
+
+If you use Docker, the certs will be automatically generated and used, if you are running npm directly, you should run `./certs/certgen.sh` so that it generates the required files.
```
miguelpduarte, line 115:
```suggestion
If you use Docker, the certs will be automatically generated and used. If you are running npm directly, you should run `./certs/certgen.sh` so that it generates the required files.
```

Dockerfile
```diff
@@ -9,6 +9,12 @@ COPY package.json package-lock.json ./
 # Install Node Packages
 RUN npm install
 
+# Copying self-signed certs
+COPY certs/ certs/
+
+# Generate certs
+RUN cd certs && ./certgen.sh && cd ..
```
miguelpduarte, line 16: The final `cd ..` is unnecessary. Every `RUN` runs in a subshell and as such the current directory is not preserved unless `WORKDIR` is used.

Dockerfile
```diff
@@ -9,6 +9,12 @@ COPY package.json package-lock.json ./
...
+
```
miguelpduarte, line 17: Furthermore, does `node:14.13.1` have openssl installed by default? It is necessary to run the script.

Dockerfile
```diff
@@ -9,6 +9,12 @@ COPY package.json package-lock.json ./
 # Install Node Packages
 RUN npm install
 
+# Copying self-signed certs
+COPY certs/ certs/
```
miguelpduarte, line 13: For a fresh install, this folder will be empty save for the certgen script.

On another hand, for an install that had already generated certs, this will result in copying over (potentially old) certs.

Wouldn't it be better to just `mkdir certs` and `COPY` the script separately to avoid this?

I'm not sure that the certgen script is passing flags to openssl to overwrite files if they exist, but I think we should just avoid the potential issue altogether.

My suggestion would be something like:

```Dockerfile
RUN mkdir certs/
COPY certs/certgen.sh certs/certgen.sh
RUN cd certs && ./certgen.sh
```

miguelpduarte (changes requested):

Dockerfile
imnotteixeira: TIL

Dockerfile
imnotteixeira: I ran the server with Docker and it worked as expected, so I assume it has :P

Dockerfile
imnotteixeira: Just tested and openssl will overwrite. Nevertheless, I could do something like that. `COPY certs/certgen.sh certs/certgen.sh` should work without the mkdir.

Dockerfile
miguelpduarte: (it all makes sense if you consider that the current directory is an environment variable in a certain way!)

Dockerfile
miguelpduarte: Does that create the folder? I was not sure of that.

If so, sounds fine, and avoids copying unnecessary certs. It may just be annoying to have to go and accept the cert everytime the container is built, but since this is after the install, it should only run again if package*.json change so it shouldn't be _too_ often I guess.

src/loaders/express.js
imnotteixeira: I think that will work. Thankfully we have a staging env to test things before production :v

miguelpduarte (approved):
:shipit:

PR #135, opened 2021-03-12
Change logout endpoint to make it idempotent

### LINKED ISSUE #126: Make logout idempotent
Currently, if trying to log out when not logged in, it will return an error. however it would be easier for testing if we could just call logout to assume there is no existing session (even if there didn't exist in the first place).

Besides, the error is not necessary since clients can assert if there is an active session by checking the session cookie, and this endpoint should not be used to check if there is an active session (obviously).

To fix this, if there is no active session, simply do nothing and return a success response

joaomrcsmartins:
Closes #126.

Removed the 'authRequired' validator from the handler, so that only log out when the user is authenticated.
In the end, the session is clear and returns 200 regardless.
Added test to confirm that logout is idempotent.


test/end-to-end/auth.js
```diff
@@ -199,5 +199,13 @@ describe("Login endpoint test", () => {
 
             expect(res.status).toBe(HTTPStatus.UNAUTHORIZED);
         });
+
+        test("should return OK since the logout is idempotent", async () => {
+            const res = await test_agent
+                .delete("/auth/login")
+                .send();
+
+            expect(res.status).toBe(HTTPStatus.OK);
```
imnotteixeira, line 208: This would pass even if it were not idempotent and for some reason the agent was logged in due to some other test.

I'd wrap this test in a describe block called "clean session" or something that would define a new test_agent just for the test (thus isolating it from the agent used in other tests, and being logged out by default). You can even rename the agent variable to logged_out_agent to make it clear.

imnotteixeira (changes requested):
Please unstage the package*.json changes since they will be solved in #134

test/end-to-end/auth.js
joaomrcsmartins: I think I fixed it, but I'm just now using this framework, so please review it again.

imnotteixeira: Just created a simple scenario to confirm this would work:

```js
describe("Using logged out user", () => {
        const logged_out_agent = agent();
        const logged_in_agent = agent();

        beforeAll(async () => {
            const user = {
                email: "user@email.com",
                password: "password123",
            };

            const res = await request()
                .post("/auth/register")
                .send(withGodToken(user));

            console.log("REG RESULT", res.body);
        });

        test("should return OK since the logout is idempotent", async () => {
            await logged_in_agent
                .post("/auth/login")
                .send({
                    email: "user@email.com",
                    password: "password123",
                });

            const res = await logged_out_agent
                .get("/auth/me")
                .send();

            console.log(res.body);
            expect(res.status).toBe(HTTPStatus.OK);
        });
    });
```

And it failed, as expected, confirming that a new user (agent) is indeed used, so your implementation should work

imnotteixeira (dismissed):

miguelpduarte (approved):
Thank you for the based commit message (and extended commit message). Great git practices here! 👏 

I can only see one other test to be done which would be checking if the response is the same after logging in and logging out twice (comparing the responses between logouts).

However, this does not add to the test coverage and is something more semantic, since idempotency does not depend on the response itself, but instead on the internal state (which is being managed by passport so we don't have an easy way to test that anyway 😅 )

As such, ship it! :shipit:

miguelpduarte:
@joaomrcsmartins The problems with the audit pipeline have been fixed in #134 . Please rebase this PR so we can merge it :smile:

imnotteixeira (approved):

PR #122, opened 2021-02-10
Adding enable/disable/edit feature to offers to limit company abuse

DoStini:
Route Enable -> Set isHidden true. Can only be used by company if the offer was hidden by company
Route DIsable -> Can only be used by admins to disable an offer (isHidden = true)
Route Hide -> Can only be used by company owner to hide its offer (later can enable again)

PR #136, opened 2021-03-21
Notify companies by email when an offer is disabled by admins 

### LINKED ISSUE #131: Notify companies by email when an offer is disabled by admins
Companies should be automatically notified when an admin disables an offer they own.

To do this, create a template `src/email-templates/` for these emails, and use EmailService to send them (check application service for examples)

BrunoRosendo:
For now, I've only created the template, I'm waiting for #122 to effectively send the e-mail.
I'm creating this PR already to get feedback on the e-mail template itself.

closes #131

### DIFF
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -153,12 +153,12 @@ module.exports = (app) => {
             try {
                 const offerService = new OfferService();
 
-                await offerService.sendOfferDisabledNotification(req.params.offerId);
                 const offer = await offerService
                     .disable(
                         req.params.offerId,
                         OfferConstants.HiddenOfferReasons.ADMIN_BLOCK,
                         req.body.adminReason);
+                await offerService.sendOfferDisabledNotification(req.params.offerId);
                 return res.json(offer);
             } catch (err) {
                 return next(err);
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -151,7 +151,10 @@ module.exports = (app) => {
         validators.canDisable,
         async (req, res, next) => {
             try {
-                const offer = await (new OfferService())
+                const offerService = new OfferService();
+
+                await offerService.sendOfferDisabledNotification(req.params.offerId);
+                const offer = await offerService
                     .disable(
                         req.params.offerId,
                         OfferConstants.HiddenOfferReasons.ADMIN_BLOCK,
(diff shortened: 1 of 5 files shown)


src/email-templates/offer_disabled_notification.handlebars
```diff
@@ -0,0 +1,9 @@
+<p>Hello {{companyName}},</p>
+<br>
+<p>We took a look at your offer for {{offerTitle}} and decided to disable it.</p>
+<p>Some common causes for this are spam, abuse or a type of offer that does not align with our principles.
+    If you believe this was a mistake and want to discuss this decision, please get back at us at
+    <a href="mailto:ni@aefeup.pt">ni@aefeup.pt</a></p>
+<br>
```
imnotteixeira, line 7: I think we should mention the offer content, to keep a record of what it was when it was disabled.

We can have a `<blockquote>` or something with the content.

This will make sure that even after the offer gets edited and enabled back, we still have the emails to look for examples of "bad" offers, while we don't have a log system to show every edit made (offer history)

imnotteixeira (changes requested):
cc @andrefmrocha

andrefmrocha:
Overall the email itself looks great, but wouldn't it make more sense for a reason to have to be chosen whenever you disable an offer? Maybe that would provide more info to the company.

imnotteixeira:
> Overall the email itself looks great, but wouldn't it make more sense for a reason to have to be chosen whenever you disable an offer? Maybe that would provide more info to the company.

This follows the same principle as the company application rejection. Admins will provide a reason, but will only be visible internally, to protect ourselves from sending an unthoughtful reason to companies directly. We purposefully leave the email generic and then explain them if we have to. We will have a reason, but only internally.

andrefmrocha (dismissed):

src/api/routes/offer.js
```diff
@@ -150,7 +150,10 @@ module.exports = (app) => {
         validators.canDisable,
         async (req, res, next) => {
             try {
-                const offer = await (new OfferService())
+                const offerService = new OfferService();
+
+                await offerService.sendOfferDisabledNotification(req.params.offerId);
```
imnotteixeira, line 155: You should only do this once the offer is actually disabled, i.e., in case the next call fails, the email would still be sent, which is bad. You can place the email sending line after the disable call, and if the first one throws, the email won't be sent

src/email-templates/companyOfferDisabled.js
```diff
@@ -0,0 +1,7 @@
+module.exports = Object.freeze({
+    OFFER_DISABLED_NOTIFICATION: (companyName, offerTitle, description) => ({
+        subject: "Your NIJobs Offer",
```
imnotteixeira, line 3: This seems too vague, how about "One of your NIJobs offers was disabled"

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #138, opened 2021-03-27
Prevent non-fully-registered companies from creating offers

### LINKED ISSUE #130: Prevent non-fully-registered companies from creating offers
After #120, companies will be able to complete their registration by adding useful information to their accounts such as their logo, which is shown on offers. 

That being said, companies that are not completely registered should not be able to create offers. This goes for admins/god trying to create an offer in their behalf as well: if the company is not fully registered, the offer cannot be created

BrunoRosendo:
closes #130


src/api/middleware/validators/validationReasons.js
```diff
@@ -35,6 +35,7 @@ const ValidationReasons = Object.freeze({
     OFFER_EDIT_PERIOD_OVER: (value) => `offer-edit-period-over:${value}-hours`,
     JOB_MIN_DURATION_NOT_SPECIFIED: "job-max-duration-requires-job-min-duration",
     REGISTRATION_FINISHED: "registration-already-hasFinishedRegistration",
+    REGISTRATION_NOT_FINISHED: "registration-not-finished-yet",
```
imnotteixeira, line 38: this is fine but please fix the "REGISTRATION_FINISHED" above

change the string to `"registration-already-finished"`

imnotteixeira (changes requested):
Please unstage the package-lock, as it should not be relevant for this PR

test/end-to-end/offer.js
```diff
@@ -628,6 +629,34 @@ describe("Offer endpoint tests", () => {
...
         });
+
+        describe("Incomplete registration of the offer's company", () => {
+            beforeAll(async () => {
+                await Company.findOneAndUpdate({ name: test_company.name }, { hasFinishedRegistration: false });
+            });
+
+            afterAll(async () => {
+                await Company.findOneAndUpdate({ name: test_company.name }, { hasFinishedRegistration: true });
+            });
```
imnotteixeira, line 640: You should not reuse the same company here, what would happen if some other test (from a sibling describe) were to run at the same time? Tests cannot depend/affect one another.

src/api/routes/offer.js
```diff
@@ -69,6 +69,7 @@ module.exports = (app) => {
         when(
             (req) => !req.body?.isHidden,
             (req, res, next) => companyMiddleware.verifyMaxConcurrentOffers(req.user?.company || req.body.owner)(req, res, next)),
+        companyMiddleware.profileComplete,
```
imnotteixeira, line 72: Move this validator right after the validators.create, so that we can save some work in case companies haven't fully registered yet (it's faster to just do a boolean check every time, instead of running the verifyMaxConcurrentOffers algorithm ever time)

imnotteixeira (changes requested):

imnotteixeira (approved):

PR #141, opened 2021-03-31
Fix bug that only allowed companies to register with multiple contacts

### LINKED ISSUE #137: [Bug] Ensure that companies can complete registration with only one contact
Currently, if we make a request with a single contact, it will be parsed as a string, instead of an array, giving an error:

image

> The [ ] are not actually needed, I was just trying to debug it

This happens (I suspect) because the validator is not using the `ensureArray` utility we've created for these types of scenarios, like the filtering in offer search, for example, or the state filtering of company applications.

It should only be a matter of adding it to the validator before the `isArray` check

Also, add a test where a single contact is posted, to validate this behavior

GuilhermeJSilva:
Closes #137

### DIFF
--- a/test/end-to-end/company.js
+++ b/test/end-to-end/company.js
@@ -83,19 +83,53 @@ describe("Company application endpoint", () => {
                     .field("contacts", ["contact1", "contact2"])
                     .expect(HTTPStatus.OK);
 
-                const test_company = [... await Company.find({}).exec()][0];
+                const test_company = [... await Company.find({})][0];
                 expect([...test_company.contacts]).toEqual(["contact1", "contact2"]);
                 expect(test_company.hasFinishedRegistration).toBe(true);
                 expect(test_company.bio).toBe("A very interesting and compelling bio");
                 const filename = path.join(`${config.upload_folder}/${test_company.id}.png`);
                 expect(fs.existsSync(filename)).toBe(true);
 
-                await test_agent
+                const res = await test_agent
                     .post("/company/application/finish")
                     .attach("logo", "test/data/logo-niaefeup.png")
                     .field("bio", "A very interesting and compelling bio")
                     .field("contacts", ["contact1", "contact2"])
                     .expect(HTTPStatus.FORBIDDEN);
+
+                expect(res.body.errors).toContainEqual(
+                    ValidationReasons.REGISTRATION_FINISHED
+                );
+
+                // clean up file created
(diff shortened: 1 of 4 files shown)


test/end-to-end/company.js
```diff
@@ -100,6 +100,32 @@ describe("Company application endpoint", () => {
...
 
+            test("should finish the application with single contact", async () => {
+                await test_agent
+                    .post("/company/application/finish")
+                    .attach("logo", "test/data/logo-niaefeup.png")
+                    .field("bio", "A very interesting and compelling bio")
+                    .field("contacts", "contact1")
+                    .expect(HTTPStatus.OK);
+
+                const test_company = [... await Company.find({}).exec()][0];
```
imnotteixeira, line 111: No need for `.exec()`

```suggestion
                const test_company = [... await Company.find({})][0];
```

imnotteixeira, line 111: define test_company on the describe level (outside of the beforeEach) and here you won't need to query the db, making the test faster, as you can access test_company directly.

You should do this (also remove the .exec) on the test above

imnotteixeira (changes requested):

test/end-to-end/company.js
GuilhermeJSilva: The company is updated in each test (contacts are added, etc), therefore a new db query is necessary

test/end-to-end/company.js
```diff
@@ -100,6 +100,32 @@ describe("Company application endpoint", () => {
...
+                expect(test_company.bio).toBe("A very interesting and compelling bio");
+                const filename = path.join(__dirname, `../../${test_company.logo}`);
+                expect(fs.existsSync(filename)).toBe(true);
+
+                await test_agent
+                    .post("/company/application/finish")
+                    .attach("logo", "test/data/logo-niaefeup.png")
+                    .field("bio", "A very interesting and compelling bio")
+                    .field("contacts", ["contact1", "contact2"])
+                    .expect(HTTPStatus.FORBIDDEN);
```
imnotteixeira, line 123: Please validate the error message (and in the previous test, I hadn't noticed), please. Other than that, it can be merged

imnotteixeira (changes requested):

imnotteixeira (changes requested):
Rebase from develop and unstage package-lock, that will probably make the tests pass

PR #143, opened 2021-04-12
Feature/block company

### LINKED ISSUE #132: Block Company from posting new offers
Admins/god should be able to block a company from posting new Offers.

This block should be reversible by an admin/god only.

There should be a reason for internal consumption (stored in the company model?), so that we know why it was blocked --- it should be removed if the company is unblocked.

An email should be sent automatically when a company is blocked/unblocked

DoStini:
closes #132

### DIFF
--- a/src/api/middleware/validators/company.js
+++ b/src/api/middleware/validators/company.js
@@ -1,10 +1,8 @@
 const { body, query, param } = require("express-validator");
-const { useExpressValidators, APIError, ErrorTypes } = require("../errorHandler");
+const { useExpressValidators } = require("../errorHandler");
 const ValidationReasons = require("./validationReasons");
 const CompanyConstants = require("../../../models/constants/Company");
 const { ensureArray } = require("./validatorUtils");
-const Company = require("../../../models/Company");
-const HTTPStatus = require("http-status-codes");
 const { isObjectId } = require("./validatorUtils");
 const CompanyService = require("../../../services/company");
 
@@ -57,33 +55,18 @@ const block = useExpressValidators([
         .trim(),
 ]);
 
-const disable = useExpressValidators([
+const unblock = useExpressValidators([
     param("companyId")
         .exists().withMessage(ValidationReasons.REQUIRED).bail()
         .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
         .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND),
 ]);
 
-const canBlock = async (req, res, next) => {
-    const company = await Company.findById(req.params.companyId);
-
-    if (company.isBlocked) {
-        return next(new APIError(
-            HTTPStatus.FORBIDDEN,
-            ErrorTypes.FORBIDDEN,
-            ValidationReasons.COMPANY_ALREADY_BLOCKED
-        ));
-    }
-
-    return next();
-};
-
 module.exports = {
     finish,
     list,
     block,
-    disable,
-    canBlock,
+    unblock,
     companyExists,
     MAX_LIMIT_RESULTS,
 };
--- a/src/api/middleware/validators/company.js
+++ b/src/api/middleware/validators/company.js
@@ -6,6 +6,7 @@ const { ensureArray } = require("./validatorUtils");
 const Company = require("../../../models/Company");
 const HTTPStatus = require("http-status-codes");
 const { isObjectId } = require("./validatorUtils");
+const CompanyService = require("../../../services/company");
 
 const MAX_LIMIT_RESULTS = 100;
 
@@ -35,7 +36,7 @@ const list = useExpressValidators([
 
 const companyExists = async (companyId) => {
     try {
-        const company = await Company.findById(companyId);
+        const company = await new CompanyService().findById(companyId);
         if (!company) throw new Error(ValidationReasons.COMPANY_NOT_FOUND(companyId));
     } catch (err) {
         console.error(err);
@@ -45,7 +46,7 @@ const companyExists = async (companyId) => {
     return true;
 };
 
-const manage = useExpressValidators([
+const block = useExpressValidators([
     param("companyId")
         .exists().withMessage(ValidationReasons.REQUIRED).bail()
         .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
@@ -69,7 +70,7 @@ const canBlock = async (req, res, next) => {
 module.exports = {
     finish,
     list,
-    manage,
+    block,
     canBlock,
     companyExists,
     MAX_LIMIT_RESULTS,
--- a/src/api/middleware/validators/company.js
+++ b/src/api/middleware/validators/company.js
@@ -36,7 +36,7 @@ const list = useExpressValidators([
 
 const companyExists = async (companyId) => {
     try {
-        const company = await new CompanyService().findById(companyId);
+        const company = await new CompanyService().findById(companyId, true);
         if (!company) throw new Error(ValidationReasons.COMPANY_NOT_FOUND(companyId));
     } catch (err) {
         console.error(err);
--- a/src/api/middleware/validators/company.js
+++ b/src/api/middleware/validators/company.js
@@ -1,8 +1,11 @@
-const { body, query } = require("express-validator");
-const { useExpressValidators } = require("../errorHandler");
+const { body, query, param } = require("express-validator");
+const { useExpressValidators, APIError, ErrorTypes } = require("../errorHandler");
 const ValidationReasons = require("./validationReasons");
 const CompanyConstants = require("../../../models/constants/Company");
 const { ensureArray } = require("./validatorUtils");
+const Company = require("../../../models/Company");
+const HTTPStatus = require("http-status-codes");
+const { isObjectId } = require("./validatorUtils");
 
 const MAX_LIMIT_RESULTS = 100;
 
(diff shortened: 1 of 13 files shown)


DoStini:
Don't know if email should be in this issue or in a separate one

imnotteixeira:
> 
> 
> Don't know if email should be in this issue or in a separate one

Do it in this one, it's part of the feature

DoStini:
We should also send an email after the company gets unblocked right?

src/api/middleware/validators/company.js
```diff
@@ -28,9 +31,44 @@ const list = useExpressValidators([
         .withMessage(ValidationReasons.MIN(0)),
 ]);
 
+const companyExists = async (companyId) => {
+    try {
+        const company = await Company.findById(companyId);
```
imnotteixeira, line 36: use the company service instead

src/api/routes/company.js
```diff
@@ -57,4 +61,40 @@ module.exports = (app) => {
         }
 
     });
+
+    router.post(
```
imnotteixeira, line 65: This should be idempotent (PUT) (if the company is already blocked, do nothing and succeed (no need for canBlock check)

src/api/routes/company.js
```diff
@@ -57,4 +61,40 @@ module.exports = (app) => {
...
+        "/:companyId/block",
+        or([
+            authMiddleware.isGod,
+            authMiddleware.isAdmin
+        ],
+        { status_code: HTTPStatus.UNAUTHORIZED, error_code: ErrorTypes.FORBIDDEN, msg: ValidationReasons.INSUFFICIENT_PERMISSIONS }),
+        validators.manage,
```
imnotteixeira, line 72: why manage and not block?

src/email-templates/companyManagement.js
```diff
@@ -0,0 +1,12 @@
+module.exports = Object.freeze({
+    COMPANY_BLOCKED_NOTIFICATION: (companyName) => ({
+        subject: "Your company on NIJobs was blocked",
```
imnotteixeira, line 3:
```suggestion
        subject: "Your company account on NIJobs has been blocked",
```

src/email-templates/companyManagement.js
```diff
@@ -0,0 +1,12 @@
...
+        template: "company_blocked_notification",
+        context: { companyName },
+    }),
+    COMPANY_UNBLOCKED_NOTIFICATION: (companyName) => ({
+        subject: "Your company on NIJobs was unblocked",
```
imnotteixeira, line 8:
```suggestion
        subject: "Your company account on NIJobs has been unblocked",
```

src/email-templates/company_blocked_notification.handlebars
```diff
@@ -0,0 +1,10 @@
+<p>Hello {{companyName}},</p>
+<br>
+<p>This email is being sent to inform that your company was blocked.</p>
```
imnotteixeira, line 3:
```suggestion
<p>This email is being sent to inform that your company account has been blocked.</p>
```

src/email-templates/company_blocked_notification.handlebars
```diff
@@ -0,0 +1,10 @@
...
+<p>Some common causes for this are spam, abuse or a type of behaviour that does not align with our principles.
+    If you believe this was a mistake and want to discuss this decision, please get back at us at
```
imnotteixeira, line 5:
```suggestion
    If you believe this was a mistake and want to contest this decision, please get back at us at
```

src/email-templates/company_unblocked_notification.handlebars
```diff
@@ -0,0 +1,10 @@
+<p>Hello {{companyName}},</p>
+<br>
+<p>After further analysis, we have decided to unblock your company.</p>
+<p>You can now place and manage offers again on our platform. The offers previously place are now also available again.
+</p>
+<p>We're sorry for any inconvenience this situation has caused.
```
imnotteixeira, line 6:
```suggestion
<p>We're sorry for any inconvenience this situation may have caused.
```

src/services/company.js
```diff
@@ -45,6 +48,35 @@ class CompanyService {
...
+     */
+    block(companyId) {
+        return Company.findByIdAndUpdate(
+            companyId,
+            { isBlocked: true },
+            { new: true },
+            (err) => {
+                if (err) {
+                    throw err;
+                }
```
imnotteixeira, line 62: isn't this the default behavior?

src/services/company.js
```diff
@@ -45,6 +48,35 @@ class CompanyService {
...
+     */
+    unblock(companyId) {
+        return Company.findByIdAndUpdate(
+            companyId,
+            { isBlocked: false },
+            { new: true },
+            (err) => {
+                if (err) {
+                    throw err;
+                }
```
imnotteixeira, line 77: isn't this the default behavior?

src/api/routes/company.js
DoStini: So it could also be used in disable route, probably not the best name.

src/api/routes/company.js
DoStini: I was following the same logic as offers but I can change

src/services/company.js
DoStini: I'll add a console.error

src/api/routes/company.js
DoStini: I mean unblock

test/end-to-end/offer.js
```diff
@@ -1848,6 +1879,37 @@ describe("Offer endpoint tests", () => {
...
+                test_offer = await Offer.create(
+                    generateTestOffer({
+                        owner: blocked_test_company._id,
+                        ownerName: blocked_test_company.name,
+                    })
+                );
+
+                test("should fail to edit offer if company blocked", async () => {
+                    const res = await request()
+                        .post(`/offers/${test_offer.id}`)
```
Naapperas, line 1902: Shuldn't this `send` have be called with `/offers/edit/${test_offer.id} ?

DoStini: Interesting, yes!

src/models/Company.js
```diff
@@ -23,6 +23,10 @@ const CompanySchema = new Schema({
         type: Boolean,
         default: false
     },
+    isBlocked: {
```
imnotteixeira, line 26: I think we should have an internal `blockReason`, similar to the `adminReason` on disabled offers. Please make sure this field is hidden for anyone which is not an admin or god. #154 might help, since it's doing a similar thing, and might have some utils which can be reused here

imnotteixeira (changes requested):

src/api/routes/company.js
```diff
@@ -51,11 +55,48 @@ module.exports = (app) => {
...
+
+    router.put(
+        "/:companyId/block",
+        or([
+            authMiddleware.isGod,
+            authMiddleware.isAdmin
+        ],
+        { status_code: HTTPStatus.UNAUTHORIZED, error_code: ErrorTypes.FORBIDDEN, msg: ValidationReasons.INSUFFICIENT_PERMISSIONS }),
+        validators.block,
+        validators.canBlock,
```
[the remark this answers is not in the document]
DoStini: Now that I think about it, it shouldn't be PUT, since the company would be blocked for different reasons, does this make sense?

src/services/company.js
```diff
@@ -20,14 +23,19 @@ class CompanyService {
...
      */
-    async findAll(limit, offset) {
+    async findAll(limit, offset, showBlocked = false, showAdminReason = false) {
 
         const totalDocCount = await Company.estimatedDocumentCount();
 
+        const companyQuery = Company.find({});
+
+        if (!showBlocked) companyQuery.withoutBlocked();
+        if (!showAdminReason) companyQuery.hideAdminReason();
```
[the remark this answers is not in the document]
DoStini: I had tests covering this situation, but they're gone somehow. Will look into this. I don't think we have yet an application on where someone could access adminReason.

src/api/routes/company.js
```diff
@@ -51,11 +55,48 @@ module.exports = (app) => {
...
+        });
+
+    router.put(
+        "/:companyId/unblock",
+        or([
+            authMiddleware.isGod,
+            authMiddleware.isAdmin
+        ],
+        { status_code: HTTPStatus.UNAUTHORIZED, error_code: ErrorTypes.FORBIDDEN, msg: ValidationReasons.INSUFFICIENT_PERMISSIONS }),
+        validators.disable,
```
imnotteixeira, line 90: Maybe this should match the route action?

```suggestion
        validators.unblock,
```

imnotteixeira (changes requested):

src/api/routes/company.js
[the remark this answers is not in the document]
imnotteixeira: Close, but not quite.

A PUT request represents an idempotent request. This can have many different meanings, but you can think of it this way: a request is idempotent if the application state after 1 call is the same as the state after n calls. The trivial case is where there are no side effects, it will obviously be idempotent. However, in this case, if you repeat the same request (including the args) n times, it shouldn't have any impact on the final result.

Think of changing the arg as executing a different operation. That operation should be reproducible without changing state after the first call, therefore it should be a PUT and there should not be validation as to whether the company is already blocked or not.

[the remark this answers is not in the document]
imnotteixeira: source: https://en.wikipedia.org/wiki/Idempotence#Computer_science_examples

imnotteixeira (approved):

Issue #144, opened 2021-04-13
Creating offer with company not working


DoStini:
After #138, companies cannot create offers since the middleware used only accounts for admin accounts.
https://github.com/NIAEFEUP/nijobs-be/blob/baef2d94f0beed9b59ba0d0e98eb5540f6055085/src/api/middleware/company.js#L38
In principle just replace `req.body.owner` with `req.user.company || req.body.owner` and add a test to verify this situation.

imnotteixeira:
Actually, it should be `req.user?.company || req.body.owner` since if it is being created in god mode, there might not be a user even.

imnotteixeira:
After giving this a better look, and to avoid this being a problem in the future again, I think we should have a middleware that instantiates a `req.ownerCompany` which will be `req.user?.company || req.body.owner`. Then, the inner functions just need to use this value, and avoid duplicating this logic.

joaomrcsmartins:
Add parameter to request in `routes/auth` and refactor in the whole project

imnotteixeira:
Here's a hidden one

https://github.com/NIAEFEUP/nijobs-be/blob/1b277c87e024329ea5173b7b0e9fb9fd67fcb12a/src/services/offer.js#L231

When an offer is hidden, a god would not be able to access it by id in this case

PR #154, opened 2021-05-01
Fix/hide admin reason

### LINKED ISSUE #149: Do not return `adminReason` to non-admin/god on offer GET
This is an internal-only field, so it should not be returned if the entity fetching the offer is a visitor or a logged in company.

Naapperas:
Closes #149

### DIFF
--- a/src/api/middleware/utils.js
+++ b/src/api/middleware/utils.js
@@ -2,7 +2,6 @@ const HTTPStatus = require("http-status-codes");
 const { APIError, ErrorTypes, hideInsecureError } = require("./errorHandler");
 const ValidationReasons = require("./validators/validationReasons");
 const lodash = require("lodash");
-const config = require("../../config/env");
 
 const DEFAULT_ERROR_CODE = ErrorTypes.VALIDATION_ERROR;
 const DEFAULT_ERROR_MSG = ValidationReasons.UNKNOWN;
@@ -82,19 +81,9 @@ const when = (
     return next();
 };
 
-/**
- * Checks if a god token was sent in the request, indicating that the user might be in god mode.
- */
-const godMode = (req, _, next) => {
-    if (req.body?.god_token === config.god_token) req.godMode = true;
-
-    return next();
-};
-
 module.exports = {
     or,
     when,
-    godMode,
     DEFAULT_ERROR_CODE,
     DEFAULT_ERROR_MSG,
     DEFAULT_STATUS_CODE
--- a/src/api/middleware/utils.js
+++ b/src/api/middleware/utils.js
@@ -2,6 +2,7 @@ const HTTPStatus = require("http-status-codes");
 const { APIError, ErrorTypes, hideInsecureError } = require("./errorHandler");
 const ValidationReasons = require("./validators/validationReasons");
 const lodash = require("lodash");
+const config = require("../../config/env");
 
 const DEFAULT_ERROR_CODE = ErrorTypes.VALIDATION_ERROR;
 const DEFAULT_ERROR_MSG = ValidationReasons.UNKNOWN;
@@ -81,9 +82,19 @@ const when = (
     return next();
 };
 
+/**
+ * Checks if a god token was sent in the request, indicating that the user might be in god mode.
+ */
+const godMode = (req, _, next) => {
+    if (req.body?.god_token === config.god_token) req.godMode = true;
+
+    return next();
+};
+
 module.exports = {
     or,
     when,
+    godMode,
     DEFAULT_ERROR_CODE,
     DEFAULT_ERROR_MSG,
     DEFAULT_STATUS_CODE
--- a/src/api/routes/offer.js
+++ b/src/api/routes/offer.js
@@ -7,7 +7,7 @@ const OfferService = require("../../services/offer");
 const HTTPStatus = require("http-status-codes");
 const { ErrorTypes, APIError } = require("../middleware/errorHandler");
 const ValidationReasons = require("../middleware/validators/validationReasons");
-const { or, when, godMode } = require("../middleware/utils");
+const { or, when } = require("../middleware/utils");
 const OfferConstants = require("../../models/constants/Offer");
 
 const router = Router();
@@ -19,7 +19,7 @@ module.exports = (app) => {
      * Gets all currently active offers (without filtering, for now)
      * supports offset and limit as query params
      */
-    router.get("/", godMode, validators.get, async (req, res, next) => {
+    router.get("/", validators.get, async (req, res, next) => {
         try {
             const offers = await (new OfferService()).get(
                 {
@@ -38,7 +38,7 @@ module.exports = (app) => {
     /**
      * Gets an offer from the database with its id
     */
-    router.get("/:offerId", godMode, validators.validOfferId, async (req, res, next) => {
+    router.get("/:offerId", validators.validOfferId, async (req, res, next) => {
         try {
             const offer = await (new OfferService()).getOfferById(
                 req.params.offerId, req.user, req?.godMode || req?.user?.isAdmin);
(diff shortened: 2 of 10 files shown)


DoStini:
Now that I see this, when does this happen? When an offer is hidden it isn't shown to the unprivileged user, and the adminReason only exists when an offer is disabled (hidden)

Naapperas:
> Now that I see this, when does this happen? When an offer is hidden it isn't shown to the unprivileged user, and the adminReason only exists when an offer is disabled (hidden)

Currently when a company is logged in they can see their offers, hidden or not, using the `GET /offers/:offerId` route. However, if said offer was disabled, they should not be able to see the reason given to disable it, since it is an internal field, hence this PR. 

The logic for that was also extended for the more general `GET /offers`, even though hidden offers don't appear to non-admin/god users.

DoStini:
Oh yeah, that makes sense

src/services/offer.js
```diff
@@ -225,10 +227,15 @@ class OfferService {
...
+    async getOfferById(offerId, user, showAdminReason = false) {
+        const offerQuery = Offer.findById(offerId);
+
+        if (!showAdminReason) offerQuery.select("-adminReason");
+
+        const offer = await offerQuery;
 
-        if (offer?.isHidden && !(user?.isAdmin || offer.owner.toString() === user?.company?._id.toString())) return null;
+        // this is a good fix for now but should be changed later
+        if (offer?.isHidden && !(user?.isAdmin || showAdminReason || offer.owner.toString() === user?.company?._id.toString())) return null;
```
imnotteixeira, line 238: Why is the `showAdminReason` relevant in this line?

imnotteixeira (changes requested):

src/api/middleware/utils.js
```diff
@@ -81,9 +82,19 @@ const when = (
...
 
+/**
+ * Checks if a god token was sent in the request, indicating that the user might be in god mode.
+ */
+const godMode = (req, _, next) => {
+    if (req.body?.god_token === config.god_token) req.godMode = true;
+
+    return next();
+};
+
```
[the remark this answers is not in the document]
Naapperas: So, if I understood correctly, this logic should be inside `src/middleware/auth.js`, right? 
Also, should I change every `isAdmin, isGod` that might be inside a call to the `or` util to `hasAdminPermissions`, since those are already going to be covered by it ?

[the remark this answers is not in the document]
imnotteixeira: Yes, sorry I had the wrong link for the code.

No don't change the validation, these are different things. In this case you won't validate anything, you'll simply provide a utility variable for future code to depend on it. It can be useful for scenarios where you can be a company account with permissions vs. an admin/god and we need to change behavior based on which of these the user is.

Besides, changing the validators to use the hasAdminPermissions defeats the purpose of the creation of the `or` util, which was built to give more context to the reader about what validation is doing, and modularity.

src/api/routes/offer.js
```diff
@@ -19,12 +19,13 @@ module.exports = (app) => {
      * Gets all currently active offers (without filtering, for now)
      * supports offset and limit as query params
      */
-    router.get("/", validators.get, async (req, res, next) => {
+    router.get("/", godMode, validators.get, async (req, res, next) => {
```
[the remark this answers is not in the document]
Naapperas: Since the auth router is the first one to be loaded (at least it looks like it), is it ok to just put the `app.use` call in the `auth` route?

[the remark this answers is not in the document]
imnotteixeira: That could work, but I think it's probably safer to have another file for stuff like this. #144 will require a similar thing and you could "set the pattern" and create a file requestEnhancers.js and you load them on loaders/express.js, just before the routes are loaded

src/services/offer.js
Naapperas: If  `user?.isAdmin` was replaced with `showAdminReason`, some tests would fail because not all calls to `getOfferById` have a third parameter, thus, even though the user could be an admin, that information would never be used inside this method.
I can't think of a good solution off the top of my head, so any suggestions are welcome.

imnotteixeira: What I meant was that `showAdminReason` has nothing to do with this line. It simply controls if the offer is to be returned or not, which should be refactored to use `req.hasAdminPermissions` instead of `user?.isAdmin`.

`showAdminReason` simply alters the query above, that's fine

Naapperas: Oh ok, got it

src/services/offer.js
```diff
@@ -227,15 +227,15 @@ class OfferService {
         return constraints.length ? { "$and": constraints } : {};
     }
 
-    async getOfferById(offerId, user, showAdminReason = false) {
+    async getOfferById(offerId, req, showAdminReason = false) {
```
imnotteixeira, line 230: The function should not receive the entire request object, but only the boolean `req.hasAdminPrivileges`. You can call the arg `hasAdminPrivileges` to make it easy to understand.

This is based on the premise that service functions shouldn't need to know about requests or routing or whatever, we could easily call them aside from that (which makes it easier for testing for example)

imnotteixeira (changes requested):

src/loaders/requestEnhancers.js
```diff
@@ -0,0 +1,8 @@
+const { hasAdminPrivileges } = require("../api/middleware/auth");
+
+module.exports = (app) => {
+
+    // check if user is god or admin
```
imnotteixeira, line 5:
```suggestion
    // Populates req.hasAdminPrivileges accordingly
```

src/services/offer.js
```diff
@@ -225,10 +227,15 @@ class OfferService {
...
-        const offer = await Offer.findById(offerId);
+    async getOfferById(offerId, user, hasAdminPrivileges, showAdminReason = false) {
+        const offerQuery = Offer.findById(offerId);
+
+        if (!showAdminReason) offerQuery.select("-adminReason");
+
+        const offer = await offerQuery;
 
-        if (offer?.isHidden && !(user?.isAdmin || offer.owner.toString() === user?.company?._id.toString())) return null;
+        // this is a good fix for now but should be changed later
```
imnotteixeira, line 237:
```suggestion

```

test/unit/auth.js
```diff
@@ -0,0 +1,75 @@
...
+        });
+
+        test("req body should have hasAdminPrivileges set to 'false' if non-admin/god user is sent", async () => {
+
+            const reqBody = {};
+
+            const req = {
+                body: reqBody,
+                user: test_user
+            };
```
imnotteixeira, line 53:
```suggestion
            const req = {
                body: {},
                user: test_user
            };
```

test/unit/auth.js
```diff
@@ -0,0 +1,75 @@
...
+        });
+
+        test("req body should have hasAdminPrivileges set to 'false' if no user is sent", async () => {
+
+            const reqBody = {};
+
+            const req = {
+                body: reqBody
+            };
+
```
imnotteixeira, line 68:
```suggestion
            const req = {
                body: {},
            };
```

test/unit/auth.js
```diff
@@ -0,0 +1,75 @@
...
+        });
+
+        test("req body should have hasAdminPrivileges set to 'true' if admin user is sent", async () => {
+
+            const reqBody = {};
+
+            const req = {
+                body: reqBody,
+                user: test_user_admin
+            };
```
imnotteixeira, line 38:
```suggestion
            const req = {
                body: {},
                user: test_user_admin
            };
```

test/unit/auth.js
```diff
@@ -0,0 +1,75 @@
...
+        const test_user = {
+            email: "user@email.com",
+            password: "password123",
+        };
+
+        test("req body should have hasAdminPrivileges set to 'true' if god_token is sent", async () => {
+
+            const reqBody = withGodToken();
+
+            const req = { body: reqBody };
```
imnotteixeira, line 23:
```suggestion
            const req = { body: withGodToken() };
```

imnotteixeira (changes requested):
After applying these changes, rebase from updated master to clear conflicts and allow merge

imnotteixeira (approved):

PR #155, opened 2021-05-05
Add functions to block all offers

DoStini:


src/services/offer.js
```diff
@@ -155,6 +156,35 @@ class OfferService {
...
     }
 
+    blockByCompany(owner) {
+        return Offer.updateMany(
+            { owner, isHidden: false },
+            {
+                isHidden: true,
+                hiddenReason: HiddenOfferReasons.COMPANY_BLOCKED,
+            },
+            { new: true },
```
BrunoRosendo, line 166: What does the this line do here? I tried searching about i but didn't find much

test/end-to-end/company.js
```diff
@@ -19,7 +22,24 @@ const getCompanies = async (options) =>
...
+        jobType: "SUMMER INTERNSHIP",
+        fields: ["DEVOPS", "MACHINE LEARNING", "OTHER"],
+        technologies: ["React", "CSS"],
+        location: "Testing Street, Test City, 123",
+        isHidden: false,
+        requirements: ["The candidate must be tested", "Fluent in testJS"],
+        ...params,
+    });
+
+
```
BrunoRosendo, line 42: Maybe we should do these tests in offer.js?  We already have the tests on offers when the company is blocked/unblocked there and that way we don't need to generate offers here as well.

BrunoRosendo (changes requested):
Looks good! Just a quick question a suggestion from me

src/services/offer.js
DoStini: In some updates it would return the new item (findAndUpdate for example), but in this case it just returns information on what was changed, so I guess this isn't needed

test/end-to-end/company.js
DoStini: I think these tests should be here because they are testing the behaviour of both company block / unblock, not some getters/ offer creations. However I don't like the idea of repeating the generate function

src/services/offer.js
```diff
@@ -155,6 +156,33 @@ class OfferService {
...
+            { owner, isHidden: true, hiddenReason: HiddenOfferReasons.COMPANY_BLOCKED },
+            {
+                isHidden: false,
+                $unset: { hiddenReason: undefined, adminReason: undefined },
+            },
+            (err) => {
+                if (err) {
+                    throw err;
+                }
+            });
```
imnotteixeira, line 184: doesn't this work naturally without passing the callback? I think it behaves like an async function which will naturally reject on error, can you try it? To force an error maybe you can try passing a wrong field on the update object, something like `$foo: "bar"` (?)

imnotteixeira (changes requested):

src/services/offer.js
imnotteixeira: https://mongoosejs.com/docs/faq.html#duplicate-query as per this, I am 99% sure that calling .updateMany without the callback will execute as a promise, so you can remove this part (and in the "block" part as well

imnotteixeira (approved):

BrunoRosendo (dismissed):

BrunoRosendo (approved):

PR #156, opened 2021-06-19
Validating max image size on Company Registration Finish

### LINKED ISSUE #150: Validate max image size on Company Registration finish
Cloudinary has a limit of 10MB for each image. We should add a validation on the /company/application/finish endpoint to have a better error explanation on our API if the logo size is > 10MB

BrunoRosendo:
closes #150 

The easiest way I found to solve the problem, since we're already using multer, was to change the fileSize in the limits to 10MB and then change the error message to fit the format we've been using (with all lowercase and hyphens). Is this what was intended? Or do you think we should at least give a longer explanation in the message?

### DIFF
--- a/src/api/middleware/utils.js
+++ b/src/api/middleware/utils.js
@@ -6,6 +6,7 @@ const lodash = require("lodash");
 const DEFAULT_ERROR_CODE = ErrorTypes.VALIDATION_ERROR;
 const DEFAULT_ERROR_MSG = ValidationReasons.UNKNOWN;
 const DEFAULT_STATUS_CODE = HTTPStatus.BAD_REQUEST;
+const MAX_FILE_SIZE = 10; // MB
 
 /**
  * Combines array of middleware using OR logic. Only fails if ALL functions fail (either by throwing or calling next(error))
@@ -86,5 +87,6 @@ module.exports = {
     when,
     DEFAULT_ERROR_CODE,
     DEFAULT_ERROR_MSG,
-    DEFAULT_STATUS_CODE
+    DEFAULT_STATUS_CODE,
+    MAX_FILE_SIZE
 };
--- a/src/api/middleware/utils.js
+++ b/src/api/middleware/utils.js
@@ -6,7 +6,7 @@ const lodash = require("lodash");
 const DEFAULT_ERROR_CODE = ErrorTypes.VALIDATION_ERROR;
 const DEFAULT_ERROR_MSG = ValidationReasons.UNKNOWN;
 const DEFAULT_STATUS_CODE = HTTPStatus.BAD_REQUEST;
-const MAX_FILE_SIZE = 10; // MB
+const MAX_FILE_SIZE_MB = 10;
 
 /**
  * Combines array of middleware using OR logic. Only fails if ALL functions fail (either by throwing or calling next(error))
@@ -88,5 +88,5 @@ module.exports = {
     DEFAULT_ERROR_CODE,
     DEFAULT_ERROR_MSG,
     DEFAULT_STATUS_CODE,
-    MAX_FILE_SIZE
+    MAX_FILE_SIZE_MB
 };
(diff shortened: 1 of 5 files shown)


test/end-to-end/company.js
```diff
@@ -162,6 +162,19 @@ describe("Company application endpoint", () => {
                     });
                 });
 
+                test("should return an error when the image size is greater than the max size", async () => {
+                    const res = await test_agent
+                        .post("/company/application/finish")
+                        .attach("logo", "test/data/logo-niaefeup-10mb.png")
+                        .expect(HTTPStatus.UNPROCESSABLE_ENTITY);
```
BrunoRosendo, line 169: Is it fine to include a 10MB+ image in the repos just for testing reasons? If not, how should we test this?

BrunoRosendo:
https://github.com/NIAEFEUP/nijobs-be/blob/f7ab840b44de0c3eca67fe11586399d1ddb5aa88/test/end-to-end/company.js#L165

Also, I didn't change it but isn't this test description wrong? Shouldn't it be something like "should return an error when the image format is not supported"?

src/api/middleware/utils.js
```diff
@@ -6,6 +6,7 @@ const lodash = require("lodash");
 const DEFAULT_ERROR_CODE = ErrorTypes.VALIDATION_ERROR;
 const DEFAULT_ERROR_MSG = ValidationReasons.UNKNOWN;
 const DEFAULT_STATUS_CODE = HTTPStatus.BAD_REQUEST;
+const MAX_FILE_SIZE = 10; // MB
```
imnotteixeira, line 9: Always clarify units on constants, the comment won't be visible when using the constant in other files, you can remove it.

```suggestion
const MAX_FILE_SIZE_MB = 10;
```

src/api/middleware/validators/validationReasons.js
```diff
@@ -40,6 +40,7 @@ const ValidationReasons = Object.freeze({
     IMAGE_FORMAT: "formats-supported-png-jpeg-jpg",
     OFFER_BLOCKED_ADMIN: "offer-blocked-by-admin",
     OFFER_HIDDEN: "offer-is-hidden",
+    FILE_TOO_LARGE: (max) => `file-cant-be-larger-than-${max}mb`,
```
imnotteixeira, line 43:
```suggestion
    FILE_TOO_LARGE: (max) => `file-cant-be-larger-than-${max}MB`,
```

test/end-to-end/company.js
```diff
@@ -170,7 +170,7 @@ describe("Company application endpoint", () => {
 
                     expect(res.body.errors).toContainEqual({
                         "location": "body",
-                        "msg": "file-too-large",
+                        "msg": "file-cant-be-larger-than-10mb",
```
imnotteixeira, line 173: Import the constant from `utils` and use it here, the test should not fail if we simply change the message constant

imnotteixeira (changes requested):

test/end-to-end/company.js
imnotteixeira: Ideally, we should mock the api that multer uses to read the file, and make it return a fake file with a fake size of 10+MB, but this is easier, and it won't hurt the actual application performance, so I'm fine with it.

test/end-to-end/company.js
```diff
@@ -213,6 +214,19 @@ describe("Company application endpoint", () => {
...
 
+                test("should return an error when the image size is greater than the max size", async () => {
+                    const res = await test_agent
+                        .post("/company/application/finish")
+                        .attach("logo", "test/data/logo-niaefeup-10mb.png")
+                        .expect(HTTPStatus.UNPROCESSABLE_ENTITY);
+
+                    expect(res.body.errors).toContainEqual({
+                        "location": "body",
+                        "msg": `file-cant-be-larger-than-${MAX_FILE_SIZE_MB}MB`,
```
imnotteixeira, line 225:
```suggestion
                        "msg": ValidationReasons.FILE_TOO_LARGE(MAX_FILE_SIZE_MB),
```

imnotteixeira (changes requested):
One last thing and it's good to go!

imnotteixeira (approved):

PR #158, opened 2021-06-26
Feature/disable company account

### LINKED ISSUE #139: Allow Companies to disable their account
Companies should be able to disable their account, making them invisible to the users.

Their offers should also be hidden (not deleted!)

This should be a temporary state, allowing companies to re-enable them

A God (not admin) should be able to do it as well

Naapperas:
Closes #139


src/services/company.js
```diff
@@ -52,7 +57,50 @@ class CompanyService {
...
+                isHidden: false,
+                $unset: { hiddenReason: undefined, adminReason: undefined },
+            },
+            { new: true },
+            (err) => {
+                if (err) {
+                    throw err;
+                }
+            }
+        );
```
DoStini, line 101: Same here

DoStini (changes requested):

src/services/company.js
```diff
@@ -52,7 +57,50 @@ class CompanyService {
...
+            {
+                isHidden: true,
+                hiddenReason: OfferConstants.HiddenOfferReasons.COMPANY_DISABLED,
+            },
+            { new: true },
+            (err) => {
+                if (err) {
+                    throw err;
+                }
+            }
```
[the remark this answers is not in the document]
Naapperas: After looking at both #155 and this PR I think it's best to wait for #155 , since the logic is the same, only changing the reason to hide offers. What do you think ?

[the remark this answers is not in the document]
DoStini: #155 is now merged

src/api/middleware/company.js
```diff
@@ -94,7 +94,32 @@ const isNotBlocked = (owner) => async (req, res, next) => {
...
+        return next(new APIError(
+            HTTPStatus.FORBIDDEN,
+            ErrorTypes.FORBIDDEN,
+            ValidationReasons.COMPANY_DISABLED
+        ));
+    }
+    return next();
+};
+
+const validCompany = (companyId) => async (req, res, next) => {
```
imnotteixeira, line 112: It seems like this function name does not clearly describe what it does (in reality it checks for permissions). Isn't there any function that does this check already? Maybe cc @DoStini or @BrunoRosendo (?)

src/api/middleware/validators/company.js
```diff
@@ -55,18 +55,35 @@ const block = useExpressValidators([
...
 ]);
 
-const unblock = useExpressValidators([
+const disable = useExpressValidators([
+    param("companyId")
+        .exists().withMessage(ValidationReasons.REQUIRED).bail()
+        .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
+        .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND),
+    body("adminReason")
+        .if((reason, { req }) => req.hasAdminPrivileges).exists().withMessage(ValidationReasons.REQUIRED).bail()
```
imnotteixeira, line 64: I think the "has admin privileges" check should be done on the route level (not on the req validator). You can use an auth middleware on the route for this

src/api/middleware/validators/company.js
```diff
@@ -55,18 +55,35 @@ const block = useExpressValidators([
...
+        .isString().withMessage(ValidationReasons.STRING).bail()
+        .trim(),
+]);
+
+const unhide = useExpressValidators([
```
imnotteixeira, line 69: This should match the operation name (enable/disable, not hide/unhide)
```suggestion
const enable = useExpressValidators([
```

src/api/middleware/validators/company.js
```diff
@@ -55,18 +55,35 @@ const block = useExpressValidators([
...
 
+/* Since their logic is the same, create a common validator and export it with a more meaningfull name.
+ * Not done with block/disable because the validation chain is slightly different
+ */
 module.exports = {
     finish,
     list,
     block,
-    unblock,
+    unhide,
```
imnotteixeira, line 83: I think we should just normalize it as having two different exports for the same thing will cause confusion in the future. Just call it enable and use for unblocking and unhiding

```suggestion
    enable,
```

src/api/middleware/company.js
```diff
@@ -94,7 +94,32 @@ const isNotBlocked = (owner) => async (req, res, next) => {
...
+    const company = await (new CompanyService()).findById(companyId, true);
+
+    // only god, admin or the same company can disable a given company
+    if (!req.hasAdminPrivileges && company._id.toString() !== req.targetOwner) {
+        return next(new APIError(
+            HTTPStatus.FORBIDDEN,
+            ErrorTypes.FORBIDDEN,
+            ValidationReasons.INVALID_COMPANY
```
imnotteixeira, line 120: This error is also too generic for the same reason

src/services/offer.js
```diff
@@ -156,23 +156,40 @@ class OfferService {
         return offer;
     }
 
-    blockByCompany(owner) {
+    hideByCompany(owner, reason) {
```
imnotteixeira, line 159: Maybe we should rename to _hideByCompany, in order to hint future devs that they should use the helpers below (blockBy*, unblockBy*, etc...)

src/services/offer.js
```diff
@@ -156,23 +156,40 @@ class OfferService {
...
             { owner, isHidden: false },
             {
                 isHidden: true,
-                hiddenReason: HiddenOfferReasons.COMPANY_BLOCKED,
+                hiddenReason: reason,
             });
     }
 
-    unblockByCompany(owner) {
+    unhideByCompany(owner, reason) {
```
imnotteixeira, line 168: Same thing for _ prefix

src/api/routes/company.js
```diff
@@ -101,4 +102,45 @@ module.exports = (app) => {
...
+            }
+        });
+
+    /**
+     * Disables a previously enabled company
+     */
+    router.put("/:companyId/disable",
+        or([
+            authMiddleware.isCompany,
+            authMiddleware.isGod
```
imnotteixeira, line 133: Can't an admin do this as well?

src/services/company.js
```diff
@@ -133,6 +139,40 @@ class CompanyService {
...
+                adminReason,
+            },
+            { new: true },
+            (err) => {
+                if (err) {
+                    console.error(err);
+                    throw err;
+                }
+            });
+    }
```
imnotteixeira, line 157: You can reuse the changeAttributes. Also, take advantage of the async nature of functions.

This change applies to the enable function as well
```suggestion
    disable(companyId, adminReason) {
    try{
        return this.changeAttributes(
            companyId,
            {
                isDisabled: true,
                adminReason,
            }
            );
            } catch(err) {
                    console.error(err);
                    throw err;
                }
    }
```

src/api/routes/company.js
```diff
@@ -101,4 +102,45 @@ module.exports = (app) => {
...
+            authMiddleware.isAdmin,
+            authMiddleware.isGod
+        ], { status_code: HTTPStatus.UNAUTHORIZED, error_code: ErrorTypes.FORBIDDEN, msg: ValidationReasons.INSUFFICIENT_PERMISSIONS }),
+        validators.enable,
+        (req, res, next) => companyMiddleware.validCompany(req.params.companyId)(req, res, next),
+        async (req, res, next) => {
+            try {
+                await new OfferService().enableByCompany(req.params.companyId);
+                const company = await (new CompanyService()).enable(req.params.companyId);
+                return res.json(company);
```
imnotteixeira, line 121: I think we should send an email notifying the user about this action, similar to what is done in the block/unblock scenario

src/api/routes/company.js
```diff
@@ -101,4 +102,45 @@ module.exports = (app) => {
...
+            authMiddleware.isCompany,
+            authMiddleware.isGod
+        ], { status_code: HTTPStatus.UNAUTHORIZED, error_code: ErrorTypes.FORBIDDEN, msg: ValidationReasons.INSUFFICIENT_PERMISSIONS }),
+        validators.disable,
+        (req, res, next) => companyMiddleware.validCompany(req.params.companyId)(req, res, next),
+        async (req, res, next) => {
+            try {
+                await new OfferService().disableByCompany(req.params.companyId);
+                const company = await (new CompanyService()).disable(req.params.companyId, req.body.adminReason);
+                return res.json(company);
```
imnotteixeira, line 141: I think we should send an email notifying the user about this action, similar to what is done in the block/unblock scenario

imnotteixeira: It should state that the company is now invisible, and the offers were hidden as well

imnotteixeira (changes requested):

src/api/middleware/company.js
BrunoRosendo: Hmm, I think we only have a function that checks permissions for owning a offer or simply checking if the user is a company, we never had to verify if they have perms for modifying a company (themselves). The closest feature was the block/unblock but that's only for admins/gods.

Maybe we could move this logic to the auth middleware with another name and call it in the endpoint's [auth checks](https://github.com/NIAEFEUP/nijobs-be/blob/74fad3988e597d6980ff09403facdf02744e8c10/src/api/routes/company.js#L110)?

imnotteixeira:
Additional to the implemented changes, I think we should restrict **offer** unhiding, so that it cannot happen if the owner is disabled (not even admins or gods can't do it, to keep consistency -  an offer should never be enabled/visible if the company is disabled)

I think the `offerOwnerNotDisabled` is being used for this, but while it works for the majority of usages (offer creation, editing, hiding, etc) it is not enough for enabling offers due to the reason I mentioned earlier

src/api/middleware/company.js
imnotteixeira: https://github.com/NIAEFEUP/nijobs-be/blob/6ad5ea03cfe2a2c6fdf1f487ed48c22c7ac04581/src/api/middleware/validators/offer.js#L511

This is what I was recalling. It's not quite the same thing, as it has logic specific to the hiddenReason and does not check ownership. In this case, there's no need to generalize I think, but I still would like for this to have a better name

src/api/routes/company.js
Naapperas: If I recall correctly, the issue stated that companies and gods should be able to perform the disable operation, but mentioned nothing about admins, so I excluded them from being able to disable companies. Should I make it so admins can also disable companies?

src/api/middleware/validators/company.js
Naapperas: My line of thought was that the req validators were already checking for the existence of the "adminReason", which is not required from companies since they are not admins, and if the validator failed because we are not a privileged user there is no need to keep validating the "adminReason" . Mind if I ask why you think this should be put in a separate middleware instead?

src/api/routes/company.js
imnotteixeira: Oh, forgot about that. Ignore this suggestion then

src/api/middleware/validators/company.js
imnotteixeira: Wait... that's the thing: disabling a company does not require an adminReason, since companies can do that themselves (same logic as hiding an offer for example). Please remove this field from the validation entirely, as well as its usages on the CompanyService disable call.

Naapperas:
There is one test missing (checking if the email is sent on company enable) which I have left commented out, since i can't find the reason why it fails.

For context, I copied the same code to test the same thing on company disable, only changing the route path and variable names, and it works.

src/api/middleware/company.js
```diff
@@ -94,7 +94,32 @@ const isNotBlocked = (owner) => async (req, res, next) => {
...
+        return next(new APIError(
+            HTTPStatus.FORBIDDEN,
+            ErrorTypes.FORBIDDEN,
+            ValidationReasons.COMPANY_DISABLED
+        ));
+    }
+    return next();
+};
+
+const canModify = (companyId) => async (req, res, next) => {
```
imnotteixeira, line 112: Still too generic, since this could be related to editing. Since it isn't, let's use `canToggleCompanyVisibility`. The Company "modifier" here helps understand that we are referring to the company's visibility and not to the ability of the company to modify visibility of its offers, for example

src/api/middleware/company.js
```diff
@@ -94,7 +94,32 @@ const isNotBlocked = (owner) => async (req, res, next) => {
...
+    const company = await (new CompanyService()).findById(companyId, true);
+
+    // only god or the same company can disable a given company
+    if (!req.hasAdminPrivileges && company._id.toString() !== req.targetOwner) {
+        return next(new APIError(
+            HTTPStatus.FORBIDDEN,
+            ErrorTypes.FORBIDDEN,
+            ValidationReasons.COMPANY_NOT_MODIFIABLE
```
imnotteixeira, line 120: Please change the error message as well (not only the variable)

```suggestion
            ValidationReasons.INSUFFICIENT_PERMISSIONS_COMAPNY_VISIBILITY
```

src/services/company.js
```diff
@@ -99,40 +105,86 @@ class CompanyService {
...
-            await EmailService.sendMail({
-                to: companyAccount.email,
-                ...COMPANY_UNBLOCKED_NOTIFICATION(company.name),
-            });
+            return this.changeAttributes(
+                companyId,
+                {
+                    isDisabled: true,
+                }
+            );
```
imnotteixeira, line 157: if this fits in a line, it's better to avoid taking 6 for an easier read
```suggestion
            return this.changeAttributes(companyId, { isDisabled: true });
```

src/services/company.js
```diff
@@ -99,40 +105,86 @@ class CompanyService {
...
+
+    enable(companyId) {
+
+        try {
+            return this.changeAttributes(
+                companyId,
+                {
+                    isDisabled: false,
+                }
+            );
```
imnotteixeira, line 172: if this fits in a line, it's better to avoid taking 6 for an easier read
```suggestion
            return this.changeAttributes(companyId, { isDisabled: false });
```

src/email-templates/company_disabled_notification.handlebars
```diff
@@ -0,0 +1,9 @@
+<p>Hello {{companyName}},</p>
+<br>
+<p>This email is being sent to inform you that your company account has been disabled.</p>
+<p>As such, your offers have all been hidden, making them invisible for users to search for.</p>
```
imnotteixeira, line 4:
```suggestion
<p>As such, your offers have all been hidden, making them invisible to users.</p>
```

src/email-templates/company_disabled_notification.handlebars
```diff
@@ -0,0 +1,9 @@
...
+<p>This is a temporary state, so you or a user with admin privileges can enable your account at any time.
```
imnotteixeira, line 5:
```suggestion
<p>This is a temporary state, so you or a user with admin privileges can re-enable your account at any time. 
```

src/api/routes/offer.js
```diff
@@ -158,6 +164,11 @@ module.exports = (app) => {
         validators.disable,
         validators.isExistingOffer,
         validators.canDisable,
+        async (req, res, next) => {
+            const offer = await (new OfferService()).getOfferById(req.params.offerId, req.targetOwner, true);
+
+            return companyMiddleware.isNotDisabled(offer.owner)(req, res, next);
+        },
```
imnotteixeira, line 171: Extract this to a middleware function (`isOwnerNotDisabled`) and place it in the offer middleware file.

You can then reuse it in the enable route

src/api/routes/offer.js
```diff
@@ -192,6 +203,11 @@ module.exports = (app) => {
         validators.canBeEnabled,
         validators.canBeManaged,
         validators.offerOwnerNotBlocked,
+        async (req, res, next) => {
+            const offer = await (new OfferService()).getOfferById(req.params.offerId, req.targetOwner, true);
+
+            return companyMiddleware.isNotDisabled(offer.owner)(req, res, next);
+        },
```
imnotteixeira, line 210: `isOwnerNotDisabled` middleware

src/api/routes/offer.js
imnotteixeira: It's a good time to create the offer middleware file (under middleware/, not middleware/validators/)

imnotteixeira (changes requested):

DoStini (approved):

PR #160, opened 2021-07-03
Feature/company offers endpoint

monkin77:
Added an endpoint that returns the offers from a specific company

### DIFF
--- a/src/api/middleware/validators/company.js
+++ b/src/api/middleware/validators/company.js
@@ -72,21 +72,11 @@ const enable = useExpressValidators([
 /**
  * Checks if the param companyId is valid and corresponds to a valid company
  */
-const getCompanyOffersValidator = useExpressValidators([
+const getOffers = useExpressValidators([
     param("companyId", ValidationReasons.DEFAULT)
         .exists().withMessage(ValidationReasons.REQUIRED).bail()
         .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
-        .custom(async (companyId, { req }) => {
-            try {
-                // including blocked companies (2nd argument)
-                const company = await (new CompanyService()).findById(companyId, true, req.hasAdminPrivileges);
-                if (!company) throw new Error(ValidationReasons.COMPANY_NOT_FOUND(companyId));
-            } catch (err) {
-                console.error(err);
-                throw err;
-            }
-            return true;
-        })
+        .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND)
 ]);
 
 module.exports = {
@@ -96,6 +86,6 @@ module.exports = {
     enable,
     disable,
     companyExists,
-    getCompanyOffersValidator,
+    getOffers,
     MAX_LIMIT_RESULTS,
 };
--- a/src/api/middleware/validators/company.js
+++ b/src/api/middleware/validators/company.js
@@ -69,6 +69,12 @@ const enable = useExpressValidators([
         .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND),
 ]);
 
+const validCompanyId = useExpressValidators([
+    param("companyId", ValidationReasons.DEFAULT)
+        .exists().withMessage(ValidationReasons.REQUIRED)
+        .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID),
+]);
+
 module.exports = {
     finish,
     list,
@@ -76,5 +82,6 @@ module.exports = {
     enable,
     disable,
     companyExists,
+    validCompanyId,
     MAX_LIMIT_RESULTS,
 };
--- a/src/api/middleware/validators/company.js
+++ b/src/api/middleware/validators/company.js
@@ -75,6 +75,23 @@ const validCompanyId = useExpressValidators([
         .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID),
 ]);
 
+const isExistingCompany = useExpressValidators([
+    param("companyId", ValidationReasons.DEFAULT)
+        .exists().withMessage(ValidationReasons.REQUIRED).bail()
+        .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
+        .custom(async (companyId, { req }) => {
+            try {
+                // including blocked companies (2nd argument)
+                const company = await (new CompanyService()).findById(companyId, true, req.hasAdminPrivileges);
+                if (!company) throw new Error(ValidationReasons.COMPANY_NOT_FOUND(companyId));
+            } catch (err) {
+                console.error(err);
+                throw err;
+            }
+            return true;
+        })
+]);
+
 module.exports = {
     finish,
     list,
@@ -83,5 +100,6 @@ module.exports = {
     disable,
     companyExists,
     validCompanyId,
+    isExistingCompany,
     MAX_LIMIT_RESULTS,
 };
--- a/src/api/middleware/validators/company.js
+++ b/src/api/middleware/validators/company.js
@@ -69,13 +69,10 @@ const enable = useExpressValidators([
         .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND),
 ]);
 
-const validCompanyId = useExpressValidators([
-    param("companyId", ValidationReasons.DEFAULT)
-        .exists().withMessage(ValidationReasons.REQUIRED)
-        .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID),
-]);
-
-const isExistingCompany = useExpressValidators([
+/**
+ * Checks if the param companyId is valid and corresponds to a valid company
+ */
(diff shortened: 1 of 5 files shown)


src/services/offer.js
```diff
@@ -239,6 +239,16 @@ class OfferService {
         return offer;
     }
 
+    async getOfferByCompanyID(companyID, user, hasAdminPrivileges) {
+        let offers = await Offer.find({ owner: companyID });
+
+        offers = offers.filter((offer) =>
+            !(offer.isHidden && !(hasAdminPrivileges || offer.owner.toString() === user?.company?._id.toString()))
```
DoStini, line 246: Maybe we should have a private function with this verification, not that it's being used in two places https://github.com/NIAEFEUP/nijobs-be/blob/4335350ff2e146a7f26e6a14837863cde2fbb9e1/src/services/offer.js#L237

src/api/routes/offer.js
```diff
@@ -58,6 +58,27 @@ module.exports = (app) => {
...
+    router.get("/company/:companyID", async (req, res, next) => {
+        try {
+            const offers = await (new OfferService()).getOfferByCompanyID(req.params.companyID, req.user, req.hasAdminPrivileges);
+
+            if (!offers) {
+                return next(new APIError(
+                    HTTPStatus.NOT_FOUND,
+                    ErrorTypes.FORBIDDEN,
+                    ValidationReasons.COMPANY_OFFER_NOT_FOUND(req.params.companyID) // This is not the offerId
+                ));
```
DoStini, line 73: This will only happen if `offers` is undefined, meaning that something went wrong in the find call. 
If this error message is necessary, then you also need to verify if the array is empty

DoStini (changes requested):
Looks good, write some tests to validate this behaviour

src/services/offer.js
imnotteixeira: I was wondering if we could reuse the `hasOwnershipRights` validator, but I don't think it is possible since it relies on the express (req,res,next) structure.

https://github.com/NIAEFEUP/nijobs-be/blob/develop/src/api/middleware/auth.js#L41

Anyway, I agree we should extract the common logic and use in the offer service and here. Otherwise, we could add a function in the offer service that would do the hasOwnershipRights verification (returns true or false), and use the .filter here. And in the offer service, it would call it's own function there as well.

src/api/routes/offer.js
```diff
@@ -58,6 +58,27 @@ module.exports = (app) => {
         }
     });
 
+    /**
+     * Gets all the offers of a certain company from the db
+     */
+    router.get("/company/:companyID", async (req, res, next) => {
```
imnotteixeira, line 64: Throughout the project we are writing Id like so, I don't have a strong opinion about either version, I just want it to be consistent

```suggestion
    router.get("/company/:companyId", async (req, res, next) => {
```

src/services/offer.js
```diff
@@ -239,6 +239,16 @@ class OfferService {
         return offer;
     }
 
+    async getOfferByCompanyID(companyID, user, hasAdminPrivileges) {
```
imnotteixeira, line 242:
```suggestion
    async getOfferByCompanyId(companyId, user, hasAdminPrivileges) {
```

src/services/offer.js
```diff
@@ -239,6 +239,16 @@ class OfferService {
...
+        let offers = await Offer.find({ owner: companyID });
+
+        offers = offers.filter((offer) =>
+            !(offer.isHidden && !(hasAdminPrivileges || offer.owner.toString() === user?.company?._id.toString()))
+        );
+
+        return offers;
+    }
```
imnotteixeira, line 250:
```suggestion
        return Offer.find({ owner: companyID })
        .filter((offer) =>
            !(offer.isHidden && !(hasAdminPrivileges || offer.owner.toString() === user?.company?._id.toString()))
        );
    }
```

imnotteixeira (changes requested):

src/api/routes/offer.js
monkin77: Thinking back on it, I believe this verification is not needed anyway so I removed it. I had it since I used another route as a template.

src/services/offer.js
monkin77: I've tried this and it results in an error:
UNEXPECTED ERROR: TypeError: Offer.find(...).filter is not a function

DoStini: Maybe (await Offer.find(...)).filter(...)

test/end-to-end/offer.js
```diff
@@ -1397,6 +1397,113 @@ describe("Offer endpoint tests", () => {
...
+
+        describe("Id Validation", () => {
+            test("should fail if requested an invalid companyId", async () => {
+                const res = await request()
+                    .get("/offers/123");
+
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.OBJECT_ID);
+            });
+
+            test("should return an empty array if there ", async () => {
```
Naapperas, line 1413: I think you forgot to finish writing the test description.

test/end-to-end/offer.js
```diff
@@ -1396,6 +1396,113 @@ describe("Offer endpoint tests", () => {
         });
     });
 
+    describe("GET /offers/:companyId", () => {
```
Naapperas, line 1399: I think this line should be `describe("GET /offers/company/:companyId", () => {`, but it's not important.

test/end-to-end/offer.js
```diff
@@ -1396,6 +1396,113 @@ describe("Offer endpoint tests", () => {
...
+        beforeAll(async () => {
+            await Offer.deleteMany({});
+        });
+
+        describe("Id Validation", () => {
+            test("should fail if requested an invalid companyId", async () => {
+                const res = await request()
+                    .get("/offers/123");
```
Naapperas, line 1407: `.get("/offers/company/123");`

src/services/offer.js
imnotteixeira: Yep, `.find` returns a promise that resolves to an array

test/end-to-end/offer.js
imnotteixeira: It matters! It should match the endpoint. you'll thank it later when debugging tests

test/end-to-end/offer.js
```diff
@@ -1396,6 +1396,113 @@ describe("Offer endpoint tests", () => {
...
+                    );
+
+                // Logout
+                await test_agent
+                    .del("/auth/login")
+                    .expect(HTTPStatus.OK);
+            });
+        });
+    });
+
```
Naapperas, line 1505: Shouldn't there be a test in which one sends the god_token?

Naapperas (changes requested):

Naapperas (approved):
LGTM

src/services/offer.js
```diff
@@ -227,18 +227,29 @@ class OfferService {
         return constraints.length ? { "$and": constraints } : {};
     }
 
+    isVisibleOffer(offer, hasAdminPrivileges, user) {
+        return !(offer?.isHidden && !(hasAdminPrivileges || offer.owner.toString() === user?.company?._id.toString()));
```
imnotteixeira, line 231: It think this has the same meaning and is easier to read, and thus maintain. Please verify if it means the same... I believe it does :P

```suggestion
       return !offer?.isHidden || hasAdminPrivileges || (offer.owner.toString() === user?.company?._id.toString());
```

imnotteixeira (changes requested):

imnotteixeira (dismissed):

src/api/routes/offer.js
```diff
@@ -58,6 +59,19 @@ module.exports = (app) => {
         }
     });
 
+    /**
+     * Gets all the offers of a certain company from the db
+     */
+    router.get("/company/:companyId", companyValidators.validCompanyId, async (req, res, next) => {
```
DoStini, line 65: I'm wondering if we should verify if the company really exists and throw an error, instead of returning an empty array

imnotteixeira: I agree, even though the current behavior is not entirely wrong (An inexistent company technically has no offers :P)

DoStini (dismissed):

DoStini:
Rebase and it's ready to merge

src/api/middleware/validators/company.js
```diff
@@ -62,11 +62,36 @@ const unblock = useExpressValidators([
         .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND),
 ]);
 
+const validCompanyId = useExpressValidators([
```
imnotteixeira, line 65: this should be named after the route which uses it e.g. `getCompanyOffers`. Finally, it should be merged with the validation below (actually, you can delete this one, and only rename the "isExistingCompany", since the validation is repeated

imnotteixeira (changes requested):

DoStini (dismissed):

imnotteixeira (approved):

PR #163, opened 2021-07-24
 Correct count of offer description characters (ignore HTML tags)

### LINKED ISSUE #162: Correct count of offer description characters (ignore HTML tags)
The offer descriptions allow HTML content, however, the user won't see the tags, so it is hard to make the char count be consistent with the content they see.

A simple example such as inserting `# Title` will be converted in the editor to an H1 and will show up like this:

image

However, the offer description will be `<h1>Title</h1>`, which has 9 more chars due to the tags.

My proposal is to ignore the HTML related stuff when parsing the description and checking the count. (https://github.com/NIAEFEUP/nijobs-be/blob/develop/src/api/middleware/validators/offer.js#L110).
To do this, we need to create a custom validator, since using the usual isLength won't allow us to check for a modified input (without HTML stuff). That validator can use the HTML Node API (create an HTML node, add innerHTML and use the node.textContent.length property to get the text length (https://stackoverflow.com/questions/51267839/character-count-with-dom-elements) 

 **I noticed that there isn't actually a check on this for the edit offer validator (only a message, maybe there was a typo), so include it there as well.**

Finally, there needs to be a custom validator on the Offer model as well, that has the same behavior, instead of using the length of the content directly.

BrunoRosendo:
closes #162

I used linkeDOM to do the parsing of the HTML content, so we can get the real size consistently.
I put the logic in an auxiliary function which is used in both model and endpoint validators.

### DIFF
--- a/src/api/middleware/validators/offer.js
+++ b/src/api/middleware/validators/offer.js
@@ -2,14 +2,13 @@ const { body, query, param } = require("express-validator");
 
 const { useExpressValidators, APIError } = require("../errorHandler");
 const ValidationReasons = require("./validationReasons");
-const { valuesInSet, ensureArray } = require("./validatorUtils");
+const { valuesInSet, ensureArray, isObjectId, charLimitWithHTML } = require("./validatorUtils");
 const JobTypes = require("../../../models/constants/JobTypes");
 const { FieldTypes, MIN_FIELDS, MAX_FIELDS } = require("../../../models/constants/FieldTypes");
 const { TechnologyTypes, MIN_TECHNOLOGIES, MAX_TECHNOLOGIES } = require("../../../models/constants/TechnologyTypes");
 const OfferService = require("../../../services/offer");
 const OfferConstants = require("../../../models/constants/Offer");
 const Company = require("../../../models/Company");
-const { isObjectId } = require("../validators/validatorUtils");
 const Offer = require("../../../models/Offer");
 const { validatePublishEndDateLimit } = require("../../../models/Offer");
 const { ErrorTypes } = require("../errorHandler");
@@ -107,7 +106,7 @@ const create = useExpressValidators([
     body("description", ValidationReasons.DEFAULT)
         .exists().withMessage(ValidationReasons.REQUIRED).bail()
         .isString().withMessage(ValidationReasons.STRING)
-        .isLength({ max: OfferConstants.description.max_length }).withMessage(ValidationReasons.TOO_LONG(1500))
+        .custom(charLimitWithHTML(OfferConstants.description.max_length))
         .trim(),
 
     body("contacts", ValidationReasons.DEFAULT)
@@ -364,7 +363,7 @@ const edit = useExpressValidators([
     body("description", ValidationReasons.DEFAULT)
         .optional()
         .isString().withMessage(ValidationReasons.STRING)
-        .withMessage(ValidationReasons.TOO_LONG(OfferConstants.description.max_length))
+        .custom(charLimitWithHTML(OfferConstants.description.max_length))
         .trim(),
 
     body("contacts", ValidationReasons.DEFAULT)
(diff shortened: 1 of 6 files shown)


src/api/middleware/validators/offer.js
```diff
@@ -107,7 +106,7 @@ const create = useExpressValidators([
     body("description", ValidationReasons.DEFAULT)
         .exists().withMessage(ValidationReasons.REQUIRED).bail()
         .isString().withMessage(ValidationReasons.STRING)
-        .isLength({ max: OfferConstants.description.max_length }).withMessage(ValidationReasons.TOO_LONG(1500))
+        .custom(charLimitWithHTML(OfferConstants.description.max_length))
```
imnotteixeira, line 109:
```suggestion
        .custom(maxHTMLContentLength(OfferConstants.description.max_length))
```

src/api/middleware/validators/offer.js
```diff
@@ -364,7 +363,7 @@ const edit = useExpressValidators([
     body("description", ValidationReasons.DEFAULT)
         .optional()
         .isString().withMessage(ValidationReasons.STRING)
-        .withMessage(ValidationReasons.TOO_LONG(OfferConstants.description.max_length))
+        .custom(charLimitWithHTML(OfferConstants.description.max_length))
```
imnotteixeira, line 366:
```suggestion
        .custom(maxHTMLContentLength(OfferConstants.description.max_length))
```

src/api/middleware/validators/validatorUtils.js
```diff
@@ -95,10 +96,21 @@ const concurrentOffersNotExceeded = (OfferModel) => async (owner, publishDate, p
     return maxConcurrent < CompanyConstants.offers.max_concurrent;
 };
 
+const charLimitWithHTML = (max) => (text) => {
```
imnotteixeira, line 99:
```suggestion
const maxHTMLContentLength = (max) => (text) => {
```

src/api/middleware/validators/validatorUtils.js
```diff
@@ -95,10 +96,21 @@ const concurrentOffersNotExceeded = (OfferModel) => async (owner, publishDate, p
...
+    return true;
+};
+
 module.exports = {
     valuesInSet,
     checkDuplicatedEmail,
     ensureArray,
     isObjectId,
     concurrentOffersNotExceeded,
+    charLimitWithHTML
```
imnotteixeira, line 115:
```suggestion
    maxHTMLContentLength
```

test/end-to-end/offer.js
```diff
@@ -355,6 +355,28 @@ describe("Offer endpoint tests", () => {
...
+                    .send(withGodToken(offer_params));
+
+                expect(res.status).toBe(HTTPStatus.OK);
+
+                const created_offer_id = res.body._id;
+                const created_offer = await Offer.findById(created_offer_id);
+
+                expect(created_offer).toBeDefined();
+                expect(created_offer).toHaveProperty("description", offer_params.description);
+            });
```
imnotteixeira, line 379: Add a test for no tags as well (where the description is only text) and another one with `description: <h1>${"a".repeat(OfferConstants.description.max_length)+1}</h1>` asserting it fails

src/models/Offer.js
```diff
@@ -134,6 +138,10 @@ function validateOwnerConcurrentOffers(value) {
     return concurrentOffersNotExceeded(this.constructor)(value, this.publishDate, this.publishEndDate);
 }
 
+function validateDescription(value) {
+    return charLimitWithHTML(OfferConstants.description.max_length)(value);
```
imnotteixeira, line 142:
```suggestion
    return maxHTMLContentLength(OfferConstants.description.max_length)(value);
```

imnotteixeira (changes requested):

test/end-to-end/offer.js
BrunoRosendo: The tests without tags are already done [here](https://github.com/NIAEFEUP/nijobs-be/blob/e158b035e0efee618a351563620724ad05cc5064/test/utils/ValidatorTester.js#L236) so I'll just do the latter

imnotteixeira (approved):

PR #164, opened 2021-08-05
Regex to validate localhost URLs

### LINKED ISSUE #161: Company logo URL validation not allowing localhost addresses
The function `validImageURL` also needs to accept a localhost address, like `https://localhost:8087/static/cniu303xc.jpeg`.
https://github.com/NIAEFEUP/nijobs-be/blob/c1347347ff183dd683656e663d271e05ed67920f/src/models/modelUtils.js#L21-L25

TiagooGomess:
Closes #161


Naapperas:
Maybe we should have some tests to verify that this is working (if they are even possible).

imnotteixeira:
> Maybe we should have some tests to verify that this is working (if they are even possible).

Yes they are possible, it's just a matter of creating a unit test calling the function with different types of expected urls (cloudinary and localhost version) and some invalid ones

Finally, @TiagooGomess can you come up with a single regex that covers both cases in one? This seems a bit inefficient.

Naapperas:
> > Maybe we should have some tests to verify that this is working (if they are even possible).
> 
> Yes they are possible, it's just a matter of creating a unit test calling the function with different types of expected urls (cloudinary and localhost version) and some invalid ones
> 
> Finally, @TiagooGomess can you come up with a single regex that covers both cases in one? This seems a bit inefficient.

Are we restricting the image hosting to Cloudinary? If so, maybe we should redo the old regex to match only Cloudinary URLs.

imnotteixeira:
No, currently we use cloudinary, but the only restriction is that the URL
is valid so that we can easily change to a different provider in case we
need it

imnotteixeira (approved):

PR #168, opened 2021-08-09
Fix/edit offer same date

### LINKED ISSUE #151: Offer editing with the same dates not throwing error
Currently, if you edit an offer and put in the request body the fields `publishDate` and `publishEndDate` with exactly the same value, no error is thrown (this is due to parsing one of the dates and not the other when checking).

image

This request body causes an "OK" response, which is not the desired behaviour.

This problem also happens when creating offers, so that needs to be fixed as well.

Naapperas:
Closes #151

### DIFF
--- a/src/api/middleware/validators/offer.js
+++ b/src/api/middleware/validators/offer.js
@@ -2,7 +2,7 @@ const { body, query, param } = require("express-validator");
 
 const { useExpressValidators, APIError } = require("../errorHandler");
 const ValidationReasons = require("./validationReasons");
-const { valuesInSet, ensureArray, isObjectId, maxHTMLContentLength } = require("./validatorUtils");
+const { valuesInSet, ensureArray, isObjectId, maxHTMLContentLength, normalizeDate } = require("./validatorUtils");
 const JobTypes = require("../../../models/constants/JobTypes");
 const { FieldTypes, MIN_FIELDS, MAX_FIELDS } = require("../../../models/constants/FieldTypes");
 const { TechnologyTypes, MIN_TECHNOLOGIES, MAX_TECHNOLOGIES } = require("../../../models/constants/TechnologyTypes");
@@ -68,8 +68,6 @@ const publishEndDateLimit = (publishEndDateCandidate, { req }) => {
     // Returning truthy value to indicate no error ocurred
     return true;
 };
-
-const normalizeDate = (date, _context) => (new Date(Date.parse(date))).toISOString();
 
 const create = useExpressValidators([
     body("title", ValidationReasons.DEFAULT)
--- a/src/api/middleware/validators/offer.js
+++ b/src/api/middleware/validators/offer.js
@@ -69,6 +69,8 @@ const publishEndDateLimit = (publishEndDateCandidate, { req }) => {
     return true;
 };
 
+const normalizeDate = (date, _context) => (new Date(Date.parse(date))).toISOString();
+
 const create = useExpressValidators([
     body("title", ValidationReasons.DEFAULT)
         .exists().withMessage(ValidationReasons.REQUIRED).bail()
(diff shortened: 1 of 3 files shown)


src/api/middleware/validators/offer.js
```diff
@@ -69,6 +69,8 @@ const publishEndDateLimit = (publishEndDateCandidate, { req }) => {
     return true;
 };
 
+const normalizeDate = (date, _context) => (new Date(Date.parse(date))).toISOString();
```
imnotteixeira, line 72: Why do you need a `_context` argument? It's not being used, it seems. Also, move this to the validators helpers file (middleware/validators/validatorUtils)

test/end-to-end/offer.js
```diff
@@ -698,6 +698,42 @@ describe("Offer endpoint tests", () => {
...
+                    ownerLogo: test_company.logo,
+                    location: "Testing Street, Test City, 123",
+                    requirements: ["The candidate must be tested", "Fluent in testJS"],
+                };
+
+                const res = await request()
+                    .post("/offers/new")
+                    .send(withGodToken(offer));
+
+                // console.info(res.body);
```
imnotteixeira, line 726:
```suggestion
```

test/end-to-end/offer.js
```diff
@@ -2004,6 +2040,58 @@ describe("Offer endpoint tests", () => {
...
+                    location: "Testing Street, Test City, 123",
+                    requirements: ["The candidate must be tested", "Fluent in testJS"],
+                });
+            });
+
+            afterAll(async () => {
+
+                await Offer.deleteMany({ offer });
+
+            });
```
imnotteixeira, line 2072:
```suggestion
            afterAll(async () => {
                await Offer.deleteMany({ offer });
            });
```

test/end-to-end/offer.js
```diff
@@ -2004,6 +2040,58 @@ describe("Offer endpoint tests", () => {
...
+                const edits = {
+                    publishDate: dateAfter,
+                    publishEndDate: dateAfter,
+                };
+
+                const res = await request()
+                    .post(`/offers/edit/${offer._id}`)
+                    .send(withGodToken(edits));
+
+                // console.info(res.body);
```
imnotteixeira, line 2085:
```suggestion
```

imnotteixeira (changes requested):

src/api/middleware/validators/offer.js
Naapperas: This function is called by the validation/sanitization chain with 2 args, afaik, so I included 2 arguments in the function prototype, even though one of them is never used. I'm not sure if calling a function passing more arguments than those declared would cause problems, if not then I'll remove this argument.

imnotteixeira: JS is pretty forgiving on that, it only matters that the order is correct (if you wanted to use the second arg, you'd need to specify the first, even though you wouldn't use it, so that js knows to assign the second one to the second variable) In this case, since you only use the first arg, you only need to specify the first one.

Additionally, as a side note, if a function has a signature with 3 arguments, you may call it with 1 or 2 parameters, and the others will be `undefined`. This is how you deal with "optional" arguments. This only works for the last arguments for obvious reasons

test/end-to-end/offer.js
```diff
@@ -698,6 +698,42 @@ describe("Offer endpoint tests", () => {
             });
         });
 
+        describe("Same 'publishDate' and 'publishEndDate'", () => {
```
imnotteixeira, line 701: you can have a beforeAll block where you delete all Offers. Then you should use generateTestOffer util below

test/end-to-end/offer.js
```diff
@@ -2004,6 +2040,58 @@ describe("Offer endpoint tests", () => {
...
+
+            const dateNow = (new Date(Date.now() + (DAY_TO_MS * 5))).toISOString();
+            const dateAfter = (new Date(Date.now() + (DAY_TO_MS * 10))).toISOString();
+            let offer;
+
+            beforeAll(async () => {
+
+                offer = await Offer.create({
```
imnotteixeira, line 2051: delete previous offers, and create this one using generateTestOffer util

imnotteixeira (changes requested):

test/end-to-end/offer.js
Naapperas: In this situation, deleting offers causes tests that happen after this one to have erroneous behavior (related to #169 ). Maybe i should keep the `Offer.deleteMany({});` line but comment it out.

imnotteixeira (approved):

PR #173, opened 2021-08-14
Adds logging of all errors to the console for easier debugging

### LINKED ISSUE #32: Improve error handling
In the /register route, if the user is not succesfully created, the response will be:

https://github.com/NIAEFEUP/nijobs-be/blob/86739e1b051beffbdc70418f44ef0e35d678a9ca/src/routes/auth.js#L68-L72

The specified reason is not always true since the error can be caused by other reasons: bcrypt fails to generate salt or hash the pw, db connection issues, etc.
Its also important to distinguish a bad request (400) from server errors (500).

A possible solution is to use error types or codes. For example, Mongo provices an error code for duplicate entries (11000), so something like this can work:

```js
if (error.code === 11000) {
      return res.status(400).json({
            "success": false,
            "reason": "Duplicated username",
            "error_code": ERROR_TYPES.DB_ERROR,
        });
    }
    return res.sendStatus(500);
```

I'd also suggest to drop the JSON notation on res.json since it accepts JS objects and makes the code more readable.

### LINKED ISSUE #106: Add logging of all errors to the console for easier debugging
After having some problems in the staging env, I went to the console to try and debug the problem, and the thrown error was too simple ( it was the Account creation failure when approving and application ), and I think everywhere we have a `catch(err)`, we should always `console.error(err)` in order to facilitate the problem solving in the future. 

In the cases where we are already logging some message, simply add the err string for context as well

BrunoRosendo:
Closes #106 
Closes #32

### DIFF
--- a/src/services/application.js
+++ b/src/services/application.js
@@ -17,7 +17,7 @@ class CompanyApplicationNotFound extends Error {
     }
 }
 
-class CompanyApplicationAlreadyReiewed extends Error {
+class CompanyApplicationAlreadyReviewed extends Error {
     constructor(msg) {
         super(msg);
     }
@@ -174,7 +174,8 @@ class CompanyApplicationService {
         try {
             application.approve();
         } catch (e) {
-            throw new CompanyApplicationAlreadyReiewed(CompanyApplicationRules.CANNOT_REVIEW_TWICE.msg);
+            console.error(e);
+            throw new CompanyApplicationAlreadyReviewed(CompanyApplicationRules.CANNOT_REVIEW_TWICE.msg);
         }
 
         try {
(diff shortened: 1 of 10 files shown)


src/services/application.js
```diff
@@ -17,7 +17,7 @@ class CompanyApplicationNotFound extends Error {
     }
 }
 
-class CompanyApplicationAlreadyReiewed extends Error {
+class CompanyApplicationAlreadyReviewed extends Error {
```
BrunoRosendo, line 20: This had a typo :p

imnotteixeira:
This also closes #32

imnotteixeira (approved):

PR #178, opened 2021-08-22
Added endpoint to delete a company and its offers

### LINKED ISSUE #140: Allow Companies to delete their account
Allow a company to delete their account, and related offers.

A God (not admin) should be able to do it as well

BrunoRosendo:
closes #140

### DIFF
--- a/src/api/middleware/validators/company.js
+++ b/src/api/middleware/validators/company.js
@@ -44,11 +44,13 @@ const companyExists = async (companyId) => {
     return true;
 };
 
+const existingCompanyParamValidator = param("companyId")
+    .exists().withMessage(ValidationReasons.REQUIRED).bail()
+    .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
+    .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND);
+
 const block = useExpressValidators([
-    param("companyId")
-        .exists().withMessage(ValidationReasons.REQUIRED).bail()
-        .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
-        .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND),
+    existingCompanyParamValidator,
     body("adminReason")
         .exists().withMessage(ValidationReasons.REQUIRED).bail()
         .isString().withMessage(ValidationReasons.STRING).bail()
@@ -56,17 +58,15 @@ const block = useExpressValidators([
 ]);
 
 const disable = useExpressValidators([
-    param("companyId")
-        .exists().withMessage(ValidationReasons.REQUIRED).bail()
-        .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
-        .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND),
+    existingCompanyParamValidator,
 ]);
 
 const enable = useExpressValidators([
-    param("companyId")
-        .exists().withMessage(ValidationReasons.REQUIRED).bail()
-        .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
-        .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND),
+    existingCompanyParamValidator,
+]);
+
+const deleteCompany = useExpressValidators([
+    existingCompanyParamValidator,
 ]);
 
 /**
@@ -85,6 +85,7 @@ module.exports = {
     block,
     enable,
     disable,
+    deleteCompany,
     companyExists,
     getOffers,
     MAX_LIMIT_RESULTS,
--- a/src/api/routes/company.js
+++ b/src/api/routes/company.js
@@ -162,11 +162,11 @@ module.exports = (app) => {
         (req, res, next) => companyMiddleware.canManageAccountSettings(req.params.companyId)(req, res, next),
         async (req, res, next) => {
             try {
-                const service = new CompanyService();
+                const companyService = new CompanyService();
                 await new OfferService().deleteOffersByCompanyId(req.params.companyId);
-                const company = await service.deleteCompany(req.params.companyId);
-                const account = await new AccountService().deleteCompanyAccount(company);
-                await service.sendCompanyDeletedNotification(account.email, company.name);
+                const company = await companyService.findAndDeleteById(req.params.companyId);
+                const account = await new AccountService().findAndDeleteByCompanyId(req.params.companyId);
+                await companyService.sendCompanyDeletedNotification(account.email, company.name);
                 return res.json(company);
             } catch (err) {
                 return next(err);
--- a/src/api/routes/company.js
+++ b/src/api/routes/company.js
@@ -16,6 +16,7 @@ const router = Router();
 const fileMiddleware  = require("../middleware/files");
 const { authRequired } = require("../middleware/auth");
 const OfferService = require("../../services/offer");
+const AccountService = require("../../services/account");
 
 module.exports = (app) => {
     app.use("/company", router);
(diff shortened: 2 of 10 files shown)


src/api/middleware/validators/company.js
```diff
@@ -69,12 +69,20 @@ const enable = useExpressValidators([
         .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND),
 ]);
 
+const deleteCompany = useExpressValidators([
+    param("companyId")
+        .exists().withMessage(ValidationReasons.REQUIRED).bail()
+        .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
+        .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND),
+]);
+
```
BrunoRosendo, line 78: I continued with the logic of having one validator per endpoint but this is repeated 3x. Shouldn't we re-utilize one validator for the 3 endpoints?

src/api/routes/company.js
```diff
@@ -148,4 +148,26 @@ module.exports = (app) => {
...
+            authMiddleware.isGod
+        ], { status_code: HTTPStatus.UNAUTHORIZED, error_code: ErrorTypes.FORBIDDEN, msg: ValidationReasons.INSUFFICIENT_PERMISSIONS }),
+        validators.deleteCompany,
+        (req, res, next) => companyMiddleware.canManageAccountSettings(req.params.companyId)(req, res, next),
+        async (req, res, next) => {
+            try {
+                const service = new CompanyService();
+                await new OfferService().deleteOffersByCompanyId(req.params.companyId);
+                await service.sendCompanyDeletedNotification(req.params.companyId);
+                const company = await service.deleteCompany(req.params.companyId);
```
BrunoRosendo, line 167: This will send an email even if the company deletion fails for some reason. Should we be concerned about this?

I had to put it before because we fetch the company in the way we're sending the emails, which can't happen if it was deleted. This would require changing _sendCompanyNotification() so it is possible to pass it a company instead of a companyId

BrunoRosendo (commented):
I didn't finish but I needed your feedback on a couple of subjects. @imnotteixeira @DoStini Maybe?

src/api/routes/company.js
imnotteixeira: Yes, we should be concerned.

_sendCompanyNotification is supposed to abstract the email sending to an existing company, changing it to receive the company kind of defeats its purpose. I suggest fetching the required info and sending the email without calling _sendCompanyNotification. (i.e., sendCompanyDeletedNotification should not call _sendCompanyNotification)

From looking at its code it reminded me that the Account of the Company being deleted should be deleted as well, to prevent logins.

src/api/middleware/validators/company.js
imnotteixeira: we should have a validator per endpoint to keep the expected structure, what you can do is create an abstraction like this:

```js
const existingCompanyParamValidator = param("companyId")
        .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
        .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND)
```

and reuse it wherever it is being used. The advantage is that since you are storing the chain itself, you can add validation after this as well (e.g., `existingCompanyParamValidator.custom(other1).custom(other2)`)

imnotteixeira: I don't know which routes require this validation. If it's only on the company scope, you can create this abstraction on the middleware/company,js, otherwise use the middleware/utils.js

src/api/routes/company.js
```diff
@@ -148,4 +148,26 @@ module.exports = (app) => {
                 return next(err);
             }
         });
+
+    /**
+     * Deletes a company and all its offers
+     */
+    router.put("/:companyId/delete",
```
imnotteixeira, line 155: should be a POST since it has the offer deletion side-effect

src/services/company.js
```diff
@@ -175,6 +176,26 @@ class CompanyService {
...
 
+    async sendCompanyDeletedNotification(companyId) {
+        await this._sendCompanyNotification(companyId, COMPANY_DELETED_NOTIFICATION);
+    }
+
+    /**
+     * Deletes a company by its ID and returns it
+     * @param {*} companyId ID of the company
+     */
+    deleteCompany(companyId) {
```
imnotteixeira, line 187: I think this should be changed to deleteCompanyAndAccount for clarity, as it should delete both

src/api/middleware/validators/company.js
BrunoRosendo: I think that would create circular dependencies so I'm going to define it in validators/company.js. You can see how it is later

imnotteixeira: I actually meant validatorUtils, since it will be used by validators. It does not import anything from validators so there shouldn't be any circular dependencies. Nevertheless, if only company validators use it, put it there

src/services/company.js
BrunoRosendo: I deleted the account in the endpoint to then use it to send the email

BrunoRosendo:
The feature should be done, it's just missing the tests but I'll do them later

src/services/account.js
```diff
@@ -34,6 +34,28 @@ class AccountService {
...
+            account._id,
+            (err) => {
+                if (err) {
+                    console.error(err);
+                    throw err;
+                }
+            }
+        );
+        return account;
+    }
```
BrunoRosendo, line 58: I tried to simply call findByIdAndRemove but the function was returning null instead of the account, so I did it like this

src/services/company.js
```diff
@@ -175,6 +176,44 @@ class CompanyService {
...
+            (err) => {
+                if (err) {
+                    console.error(err);
+                    throw err;
+                }
+            }
+        );
+        return company;
+    }
+
```
BrunoRosendo, line 216: Same thing here

BrunoRosendo:
I also ran npm audit fix since the CI audit was failing

src/email-templates/company_deleted_notification.handlebars
```diff
@@ -0,0 +1,10 @@
+<p>Hello {{companyName}},</p>
+<br>
+<p>This email is being sent to inform you that your company account has been deleted.</p>
+<p>As such, your offers have all been deleted as well.</p>
+<p>This is a permanent action but always feel free to come back and send another application!</p>
```
imnotteixeira, line 5:
```suggestion
<p>This is a permanent action, but feel free to come back by submitting another application!</p>
```

src/services/company.js
```diff
@@ -175,6 +176,44 @@ class CompanyService {
...
+        await Company.findByIdAndRemove(
+            companyId,
+            (err) => {
+                if (err) {
+                    console.error(err);
+                    throw err;
+                }
+            }
+        );
+        return company;
```
imnotteixeira, line 214:
```suggestion
    try{
        const company = await Company.findById(companyId);
        await Company.findByIdAndRemove(companyId);
        return company;
        } catch(err) {
             console.error(err);
                    throw err;
        }
```

src/api/routes/company.js
```diff
@@ -148,4 +149,27 @@ module.exports = (app) => {
...
+    router.post("/:companyId/delete",
+        or([
+            authMiddleware.isCompany,
+            authMiddleware.isGod
+        ], { status_code: HTTPStatus.UNAUTHORIZED, error_code: ErrorTypes.FORBIDDEN, msg: ValidationReasons.INSUFFICIENT_PERMISSIONS }),
+        validators.deleteCompany,
+        (req, res, next) => companyMiddleware.canManageAccountSettings(req.params.companyId)(req, res, next),
+        async (req, res, next) => {
+            try {
+                const service = new CompanyService();
```
imnotteixeira, line 165:
```suggestion
                const companyService = new CompanyService();
```

imnotteixeira (changes requested):

imnotteixeira:
As mentioned in the informational email, this is a permanent and destructive action. As such, we should have a way of confirming deletion via email before actually deleting it. I'm creating an issue to track that feature afterwards (#184).

imnotteixeira (approved):

PR #187, opened 2021-09-16
Fix/parse offer fields

### LINKED ISSUE #177: Move 'targetOwner' setter to correct place
#167 introduced the `setTargetOwner` middleware that is currently being used by `requestEnhancers`. However, since the `targetOwner` field only applies to Offers, it should only be used by the offers route, and not the entire application.

In principle, just move the call to `setTargetOwner` from `requestEnhancers` to the beginning of the offers route module (also, changing the location of the middleware might be a good idea).

### LINKED ISSUE #186: It's possible for a company to create an infinite number of offers
When I tried to create some offers for a specific company, I realized that when the number of created offers surpasses 5, there is no error, even if all of the offers are visible.

Naapperas:
Closes #177 
Closes #186


DoStini:
I don't think this is the best solution. If we use isHidden in another place we will have to call once again `JSON.parse`

src/api/middleware/company.js
```diff
@@ -27,7 +27,7 @@ const verifyMaxConcurrentOffers = (owner, publishDate, publishEndDate, offerId)
 
 const verifyMaxConcurrentOffersOnCreate = (req, res, next) => {
 
-    if (req.body?.isHidden) return next();
+    if (req.body?.isHidden && JSON.parse(req.body?.isHidden)) return next();
```
DoStini, line 30: I think we should have some general sanitiser that would parse the request of the body.

DoStini (changes requested):

DoStini:
After finding [this thread](https://stackoverflow.com/questions/10005939/how-do-i-consume-the-json-post-data-in-an-express-application/10007542#10007542).
It is possible that problem is that `bodyParser` is deprecated. Used in:
https://github.com/NIAEFEUP/nijobs-be/blob/fb2ffcc8d462dc96dc46f38e73b70ec91df5c3bb/src/loaders/express.js#L46

It appears the solution is to replace `bodyParser` with `express`. [Source](https://stackoverflow.com/questions/24330014/bodyparser-is-deprecated-express-4)

BrunoRosendo:
Please check if @DoStini's solution works, that'll be the ideal!

Additionally, I think it'd be nice to have a test covering this scenario. Just make a test similar to the ones already implemented but instead of `true` use `"true"`

src/api/middleware/company.js
Naapperas: > After finding [this thread](https://stackoverflow.com/questions/10005939/how-do-i-consume-the-json-post-data-in-an-express-application/10007542#10007542), I think it is possible that problem is that `bodyParser` is deprecated. Used in:
> 
> https://github.com/NIAEFEUP/nijobs-be/blob/fb2ffcc8d462dc96dc46f38e73b70ec91df5c3bb/src/loaders/express.js#L46
> 
> It appears the solution is to replace `bodyParser` with `express`, since we are using express 4.17. [Source](https://stackoverflow.com/questions/24330014/bodyparser-is-deprecated-express-4)

Since we are no longer using `body-parser`, should I uninstall that module ?

Naapperas:
> After finding [this thread](https://stackoverflow.com/questions/10005939/how-do-i-consume-the-json-post-data-in-an-express-application/10007542#10007542), I think it is possible that problem is that `bodyParser` is deprecated. Used in:
> 
> https://github.com/NIAEFEUP/nijobs-be/blob/fb2ffcc8d462dc96dc46f38e73b70ec91df5c3bb/src/loaders/express.js#L46
> 
> It appears the solution is to replace `bodyParser` with `express`, since we are using express 4.17. [Source](https://stackoverflow.com/questions/24330014/bodyparser-is-deprecated-express-4)

Since we are no longer using `body-parser`, should I uninstall that module?

DoStini:
> > After finding [this thread](https://stackoverflow.com/questions/10005939/how-do-i-consume-the-json-post-data-in-an-express-application/10007542#10007542), I think it is possible that problem is that `bodyParser` is deprecated. Used in:
> > https://github.com/NIAEFEUP/nijobs-be/blob/fb2ffcc8d462dc96dc46f38e73b70ec91df5c3bb/src/loaders/express.js#L46
> > 
> > It appears the solution is to replace `bodyParser` with `express`, since we are using express 4.17. [Source](https://stackoverflow.com/questions/24330014/bodyparser-is-deprecated-express-4)
> 
> Since we are no longer using `body-parser`, should I uninstall that module?

Yes, if everything works well

TiagooGomess (approved):

DoStini (approved):

imnotteixeira:
From the meeting: 
* remove parseIsHidden
* add isBooleanType custom validator and replace isBoolean check, to check for boolean type instead of strings that match ["true", "false", "0", "1"]

Naapperas:
> I thought I left a comment yesterday but apparently not, my bad.
> 
> Input validation tests have been written for offer `create` and `edit` routes (the ones using the `isHidden` field) using the custom validator. There is another field using the `isBoolean` validator but I have left that be because that field is passed through the query of the request and so will be parsed by express.json() into a string.
> 
> The audit issues state that there is an issue with "No fix available", what should we do ?
> 
> (Also, I found that we have been wrongly testing offer editting in
> 
> https://github.com/NIAEFEUP/nijobs-be/blob/ca3c0e98c65297353e2bf74e5497f8cb82b22238/test/end-to-end/offer.js#L2138
> since we are testing the wrong route).

Regarding the audit thing, this check was done on my machine, which does not match what the audit CI check says

Naapperas:
> > I thought I left a comment yesterday but apparently not, my bad.
> > Input validation tests have been written for offer `create` and `edit` routes (the ones using the `isHidden` field) using the custom validator. There is another field using the `isBoolean` validator but I have left that be because that field is passed through the query of the request and so will be parsed by express.json() into a string. (Left a comment in the code explaining this decision)
> > The audit issues state that there is an issue with "No fix available", what should we do ?
> > (Also, I found that we have been wrongly testing offer editting in
> > https://github.com/NIAEFEUP/nijobs-be/blob/ca3c0e98c65297353e2bf74e5497f8cb82b22238/test/end-to-end/offer.js#L2138
> > 
> > since we are testing the wrong route).
> 
> Why didn't you change the wrong test? Does it fail if you do?
> 
> Regarding the audit issue, running `npm audit fix` locally didn't fix it?

I did not change the wrong test because it is not part of the scope of this PR and, since the `edit` route uses "URI variables" (the identifiers after the ':'), I don't know how I could correct the tests.

Regarding the audit, I tried a couple combinations of `npm audit fix` and `npm audit fix --force` and, despite the number of issues being different, I always ended up with "Some issues need review, and may require choosing a different dependency.".

imnotteixeira (dismissed):

DoStini:
Rebase @Naapperas

imnotteixeira (approved):
:rocket:

PR #188, opened 2021-09-19
Refactor code to use ES6 imports

imnotteixeira:
It's time to update the app to use the more modern syntax, as we are using  node-14 and they are becoming standard. This will also help setup the project to include the new `nijobs-commons` package


imnotteixeira:
Fixing the audit issue involves upgrading to mongoose@6. I suggest solving that later.

Naapperas (approved):

Naapperas:
> Fixing the audit issue involves upgrading to mongoose@6. I suggest solving that later.

Maybe I'm mistaken, but I had a similar issue in a different PR (#187) in the exact same package (mpath) and running `npm audit fix` solved the issue for me. I just checked and the `mongoose` version installed in my system after that audit fix is 5.13.9, so I don't think there is a need to upgrade to mongoose@6.

imnotteixeira:
> > Fixing the audit issue involves upgrading to mongoose@6. I suggest solving that later.
> 
> Maybe I'm mistaken, but I had a similar issue in a different PR (#187) in the exact same package (mpath) and running `npm audit fix` solved the issue for me. I just checked and the `mongoose` version installed in my system after that audit fix is 5.13.9, so I don't think there is a need to upgrade to mongoose@6.

If you can fix the audit issue without migrating to the next major version, great, in my case it required me to run `npm audit fix --force` since it required that bump. In any case, make sure the tests pass, and some manual testing of features wouldn't be wrong either

PR #192, opened 2021-10-17
Fixed request crash when passing jobStartDate as null

### LINKED ISSUE #191: Dates causing request crash
When a request with a null date is sent (specifically job start date), if the date is null, the request fails and doesn't time out

BrunoRosendo:
closes #191


imnotteixeira:
The audit issue is fixed by bumping mpath to >= 0.8.4. Mongoose did that on 5.13.9 https://github.com/Automattic/mongoose/blob/5.x/History.md#5139--2021-09-06

My suggestion is to bump mongoose to 5.13.11 to avoid migrating to v6 for now

BrunoRosendo:
> The audit issue is fixed by bumping mpath to >= 0.8.4. Mongoose did that on 5.13.9 https://github.com/Automattic/mongoose/blob/5.x/History.md#5139--2021-09-06
> 
> My suggestion is to bump mongoose to 5.13.11 to avoid migrating to v6 for now

I fixed almost all of the audit but there's still a npm-audit-helper vulnerability I can't fix because the package didn't bump ansi-regex yet. It doesn't seem very important though. Should I just leave it for when they bump it?

test/unit/EmailService.js
```diff
@@ -41,7 +41,7 @@ describe("EmailService", () => {
         });
 
         const MathRandom = Math.random;
-        Math.random = () => 123456789987654321;
+        Math.random = () => 123456789987654321n;
```
imnotteixeira, line 44: Why did you change this?

imnotteixeira (approved):

test/unit/EmailService.js
BrunoRosendo: Jest gave a warning because of that (after the bump). The normal int didn't have enough precision so the BigInt was needed

PR #196, opened 2021-12-05
Endpoint to verify max concurrent offers given a company and two dates

### LINKED ISSUE #194: Endpoint to verify max concurrent offers given an owner and two dates

BrunoRosendo:
closes #194


src/api/middleware/validators/company.js
```diff
@@ -77,3 +77,19 @@ export const getOffers = useExpressValidators([
         .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
         .custom(companyExists).withMessage(ValidationReasons.COMPANY_NOT_FOUND)
 ]);
+
+export const checkConcurrent = useExpressValidators([
+    existingCompanyParamValidator,
+
+    body("publishDate", ValidationReasons.DEFAULT)
+        .exists().withMessage(ValidationReasons.REQUIRED).bail()
```
imnotteixeira, line 85: I think that if `publishDate` is not passed it should default to `now`, not being required in the request

src/api/middleware/validators/company.js
```diff
@@ -77,3 +77,19 @@ export const getOffers = useExpressValidators([
...
+        .isISO8601({ strict: true }).withMessage(ValidationReasons.DATE).bail()
+        .customSanitizer(normalizeDate),
+
+    body("publishEndDate", ValidationReasons.DEFAULT)
+        .exists().withMessage(ValidationReasons.REQUIRED).bail()
```
imnotteixeira, line 90: I think that if `publisEndhDate` is not passed it should default to `publishDate + max_offer_duration`, not being required in the request. If none are passed, `publishEndDate` would be `now +  max_offer_duration`

src/api/routes/company.js
```diff
@@ -174,4 +176,29 @@ export default (app) => {
                 return next(err);
             }
         });
+
+    /**
+     * Verifies if a company has reached max concurrent offers between two dates
+     */
+    router.get("/:companyId/concurrent",
```
imnotteixeira, line 183:
```suggestion
    router.get("/:companyId/hasReachedMaxConcurrentOffersBetweenDates",
```

imnotteixeira (changes requested):

src/api/middleware/validators/company.js
```diff
@@ -77,3 +78,30 @@ export const getOffers = useExpressValidators([
...
+        .customSanitizer(normalizeDate)
+        .custom((publishEndDate, { req }) => publishEndDate > req.body.publishDate)
+        .withMessage(ValidationReasons.MUST_BE_AFTER("publishDate")),
+]);
+
+export const setDefaultValuesConcurrent = (req, res, next) => {
+    if (!req.body?.publishDate) req.body.publishDate = new Date(Date.now()).toISOString();
+
+    if (!req.body?.publishEndDate) {
+        const publishDateMs = Date.parse(req.body.publishDate);
```
imnotteixeira, line 102:
```suggestion
        const publishDateMS = Date.parse(req.body.publishDate);
```

src/api/routes/company.js
```diff
@@ -174,4 +176,30 @@ export default (app) => {
                 return next(err);
             }
         });
+
+    /**
+     * Verifies if a company has reached max concurrent offers between two dates
+     */
+    router.get("/:companyId/hasReachedMaxConcurrentOffersBetweenDates",
+        validators.setDefaultValuesConcurrent,
```
imnotteixeira, line 184: If validation only happens after this line, there's no guarantee you'll receive valid dates, which will make this function crash (#191)

src/api/middleware/validators/company.js
imnotteixeira: "If validation only happens after this line, there's no guarantee you'll receive valid dates, which will make this function crash (#191)"

This will fail if the date is a malformed string, for example

imnotteixeira (changes requested):

imnotteixeira (approved):

Issue #129, opened 2021-03-04
Improve search performance


imnotteixeira:
Currently, searching many offers depends on a `offset` + `limit` combination to simulate pagination of results. This can be slow and we might lose results in the event of addition/removal of elements in between page changes. 

I propose the change to keyset-based search.
To the one implementing this, please read this article, which explains it very well: https://medium.com/swlh/mongodb-pagination-fast-consistent-ece2a97070f3

The basic idea is to take advantage of object ids being naturally sorted, so we can use the value of the last returned object as the threshold to fetch the next page (next page will have elements whose IDs will be > than the last element from previous page)

miguelpduarte:
> By using the keyset method, we lose the ability to randomly access the pages. Simply saying, we actually can’t get page 3 without getting page 2 and getting page 2 without getting page 1. But the good thing is we don’t really care about that in most applications.

This bothers me a bit. Is the performance _that_ terrible?

miguelpduarte:
I would even dare to say: Since the data we send in each response is not all that large, we could even consider having a larger "window" in the back-end and paginate that in the front-end. May increase implementation complexity in the front-end but may help with potential performance issues if that was the core focus.

Since it seems to only be a problem after several tens of pages and our goal is to have a lot of filtering possibilities, I don't see us reaching that level of active offers quickly enough to justify losing the possibility of random page access, I think.

PR #207, opened 2022-02-17
Feature/company profile

### LINKED ISSUE #9: Company profile getting
Relevant information:
* Name
* Company logo
* Contacts infobox
* Small company bio
* X most recently published offers

toni-santos:
Closes #9

### DIFF
--- a/src/api/middleware/company.js
+++ b/src/api/middleware/company.js
@@ -85,6 +85,69 @@ export const profileComplete = async (req, res, next) => {
     return next();
 };
 
+export const restrictedAccess = (owner) => async (req, res, next) => {
+    const company = await (new CompanyService()).findById(owner, true);
+    let error = {};
+
+    if (company.isDisabled) {
+        if (req.params?.companyId === req.user.company) {
+            error = new APIError(
+                HTTPStatus.OK,
+                ErrorTypes.VALIDATION_ERROR,
+                ValidationReasons.COMPANY_DISABLED,
+                { company: company }
+            );
+        } else {
+            error = new APIError(
+                HTTPStatus.FORBIDDEN,
+                ErrorTypes.FORBIDDEN,
+                ValidationReasons.NOT_FOUND
+            );
+        }
+    }
+
+    if (company.isBlocked) {
+        if (req.params?.companyId === req.user.company) {
+            error = new APIError(
+                HTTPStatus.OK,
+                ErrorTypes.VALIDATION_ERROR,
+                ValidationReasons.COMPANY_BLOCKED,
+                { company: company }
+            );
+        } else {
+            error = new APIError(
+                HTTPStatus.FORBIDDEN,
+                ErrorTypes.FORBIDDEN,
+                ValidationReasons.NOT_FOUND
+            );
+        }
+    }
+
+
+    return next(error);
+};
+
+export const registrationStatus = (owner) => async (req, res, next) => {
+    const company = await (new CompanyService()).findById(owner, true);
+    if (!company.hasFinishedRegistration && (req.params?.companyId === req.user.company)) {
+        return next(new APIError(
+            HTTPStatus.FORBIDDEN,
+            ErrorTypes.FORBIDDEN,
+            ValidationReasons.REGISTRATION_NOT_FINISHED
+        ));
+    }
+
+    if (!company.hasFinishedRegistration && !(req.params?.companyId === req.user.company)) {
+        return next(new APIError(
+            HTTPStatus.FORBIDDEN,
+            ErrorTypes.FORBIDDEN,
+            ValidationReasons.NOT_FOUND
+        ));
+    }
+
+    return next();
+};
+
 export const isNotBlocked = (owner) => async (req, res, next) => {
     const company = await (new CompanyService()).findById(owner, true);
     if (company.isBlocked) {
--- a/src/api/middleware/company.js
+++ b/src/api/middleware/company.js
@@ -89,59 +89,39 @@ export const restrictedAccess = (owner) => async (req, res, next) => {
     const company = await (new CompanyService()).findById(owner, true);
     let error = {};
 
-    if (company.isDisabled) {
-        if (req.params?.companyId === req.user.company) {
-            error = new APIError(
-                HTTPStatus.OK,
-                ErrorTypes.VALIDATION_ERROR,
-                ValidationReasons.COMPANY_DISABLED,
-                { company: company }
-            );
-        } else {
-            error = new APIError(
-                HTTPStatus.FORBIDDEN,
-                ErrorTypes.FORBIDDEN,
-                ValidationReasons.NOT_FOUND
-            );
-        }
-    }
+    if (req.params?.companyId === req.user.company) {
+        let reason = ValidationReasons.UNKNOWN;
 
-    if (company.isBlocked) {
-        if (req.params?.companyId === req.user.company) {
-            error = new APIError(
-                HTTPStatus.OK,
-                ErrorTypes.VALIDATION_ERROR,
-                ValidationReasons.COMPANY_BLOCKED,
-                { company: company }
-            );
-        } else {
-            error = new APIError(
-                HTTPStatus.FORBIDDEN,
-                ErrorTypes.FORBIDDEN,
-                ValidationReasons.NOT_FOUND
-            );
-        }
-    }
+        if (company.isBlocked)
+            reason = ValidationReasons.COMPANY_BLOCKED;
+        else if (company.isDisabled)
+            reason = ValidationReasons.COMPANY_DISABLED;
 
+        error = new APIError(
+            HTTPStatus.OK,
+            ErrorTypes.VALIDATION_ERROR,
+            reason,
+            { company: company }
+        );
+    } else {
+        error = new APIError(
+            HTTPStatus.FORBIDDEN,
+            ErrorTypes.FORBIDDEN,
+            ValidationReasons.NOT_FOUND
+        );
+    }
 
     return next(error);
 };
 
 export const registrationStatus = (owner) => async (req, res, next) => {
     const company = await (new CompanyService()).findById(owner, true);
-    if (!company.hasFinishedRegistration && (req.params?.companyId === req.user.company)) {
-        return next(new APIError(
-            HTTPStatus.FORBIDDEN,
-            ErrorTypes.FORBIDDEN,
-            ValidationReasons.REGISTRATION_NOT_FINISHED
-        ));
-    }
 
-    if (!company.hasFinishedRegistration && !(req.params?.companyId === req.user.company)) {
+    if (!company.hasFinishedRegistration) {
         return next(new APIError(
             HTTPStatus.FORBIDDEN,
             ErrorTypes.FORBIDDEN,
-            ValidationReasons.NOT_FOUND
+            (req.params?.companyId !== req.user.company) ? ValidationReasons.NOT_FOUND : ValidationReasons.REGISTRATION_NOT_FINISHED
         ));
     }
 
--- a/src/api/middleware/company.js
+++ b/src/api/middleware/company.js
@@ -85,45 +85,41 @@ export const profileComplete = async (req, res, next) => {
     return next();
 };
 
-export const restrictedAccess = (owner) => async (req, res, next) => {
-    const company = await (new CompanyService()).findById(owner, true);
-    let error = {};
+export const canAccessProfile = (companyId) => async (req, res, next) => {
+    const company = await new CompanyService().findById(companyId, true);
 
-    if (req.params?.companyId === req.user.company) {
-        let reason = ValidationReasons.UNKNOWN;
-
-        if (company.isBlocked)
-            reason = ValidationReasons.COMPANY_BLOCKED;
-        else if (company.isDisabled)
-            reason = ValidationReasons.COMPANY_DISABLED;
-
-        error = new APIError(
-            HTTPStatus.OK,
+    const notFound = () =>
+        new APIError(
+            HTTPStatus.UNPROCESSABLE_ENTITY,
             ErrorTypes.VALIDATION_ERROR,
-            reason,
-            { company: company }
+            [
+                {
+                    value: companyId,
+                    msg: ValidationReasons.COMPANY_NOT_FOUND(companyId),
+                    param: "companyId",
+                    location: "params",
+                },
+            ]
         );
-    } else {
-        error = new APIError(
-            HTTPStatus.FORBIDDEN,
-            ErrorTypes.FORBIDDEN,
-            ValidationReasons.NOT_FOUND
+
+    const errorOrNotFound = (reason) =>
+        companyId === req.user?.company?.toString() || req.hasAdminPrivileges
+            ? new APIError(HTTPStatus.FORBIDDEN, ErrorTypes.FORBIDDEN, reason)
+            : notFound();
+
+    if (!company.hasFinishedRegistration)
+        return next(
+            errorOrNotFound(ValidationReasons.REGISTRATION_NOT_FINISHED)
         );
-    }
 
-    return next(error);
-};
+    if (req.hasAdminPrivileges)
+        return next();
 
-export const registrationStatus = (owner) => async (req, res, next) => {
-    const company = await (new CompanyService()).findById(owner, true);
+    if (company.isBlocked)
+        return next(errorOrNotFound(ValidationReasons.COMPANY_BLOCKED));
 
-    if (!company.hasFinishedRegistration) {
-        return next(new APIError(
-            HTTPStatus.FORBIDDEN,
-            ErrorTypes.FORBIDDEN,
-            (req.params?.companyId !== req.user.company) ? ValidationReasons.NOT_FOUND : ValidationReasons.REGISTRATION_NOT_FINISHED
-        ));
-    }
+    if (company.isDisabled && companyId !== req.user?.company?.toString())
+        return next(notFound());
 
     return next();
 };
--- a/src/api/middleware/validators/company.js
+++ b/src/api/middleware/validators/company.js
@@ -75,6 +75,10 @@ export const deleteCompany = useExpressValidators([
     existingCompanyParamValidator,
 ]);
 
+export const profile = useExpressValidators([
+    existingCompanyParamValidator,
+]);
+
 export const getOffers = useExpressValidators([
     existingCompanyParamValidator,
 ]);
--- a/src/api/routes/company.js
+++ b/src/api/routes/company.js
@@ -73,7 +73,7 @@ export default (app) => {
 
     });
 
-    router.get("/:companyId/profile",
+    router.get("/:companyId",
         validators.profile,
         (req, res, next) => companyMiddleware.restrictedAccess(req.params.companyId)(req, res, next),
         (req, res, next) => companyMiddleware.registrationStatus(req.params.companyId)(req, res, next),
--- a/src/api/routes/company.js
+++ b/src/api/routes/company.js
@@ -72,6 +72,15 @@ export default (app) => {
 
     });
 
+    router.get("/:companyId/profile", validators.profile, async (req, res) => {
+        const company = await new CompanyService().findById(req.params.companyId, req.hasAdminPrivileges, req.hasAdminPrivileges);
+        let offers = await new OfferService().getOffersByCompanyId(req.params.companyId, req.params.companyId, req.hasAdminPrivileges);
+        offers = offers.sort((a1, a2) =>
+            a1.publishDate > a2.publishDate
+        ).slice(0, 5);
+        return res.json({ company, offers });
+    });
+
     router.put(
         "/:companyId/block",
         or([
--- a/src/api/routes/company.js
+++ b/src/api/routes/company.js
@@ -73,15 +73,20 @@ export default (app) => {
 
     });
 
-    router.get("/:companyId/profile", validators.profile, async (req, res) => {
-        const company = await new CompanyService().findById(req.params.companyId, req.hasAdminPrivileges, req.hasAdminPrivileges);
-        const offers = (await new OfferService()
-            .getOffersByCompanyId(req.params.companyId, req.targetOwner, req.hasAdminPrivileges, {
-                sort: { publishDate: "desc" }, limit: CompanyConstants.offers.max_profile_visible
-            })
-        );
-        return res.json({ company, offers });
-    });
+    router.get("/:companyId/profile",
+        validators.profile,
+        (req, res, next) => companyMiddleware.restrictedAccess(req.params.companyId)(req, res, next),
+        (req, res, next) => companyMiddleware.registrationStatus(req.params.companyId)(req, res, next),
+        async (req, res) => {
+            const company = await new CompanyService().findById(req.params.companyId, req.hasAdminPrivileges, req.hasAdminPrivileges);
+            const offers = (await new OfferService()
+                .getOffersByCompanyId(req.params.companyId, req.targetOwner, req.hasAdminPrivileges, {
+                    sort: { publishDate: "desc" }, limit: CompanyConstants.offers.max_profile_visible
+                })
+            );
+            return res.json({ company, offers });
+        }
+    );
 
     router.put(
         "/:companyId/block",
--- a/src/api/routes/company.js
+++ b/src/api/routes/company.js
@@ -76,9 +76,10 @@ export default (app) => {
     router.get("/:companyId/profile", validators.profile, async (req, res) => {
         const company = await new CompanyService().findById(req.params.companyId, req.hasAdminPrivileges, req.hasAdminPrivileges);
         const offers = (await new OfferService()
-            .getOffersByCompanyId(req.params.companyId, req.targetOwner, req.hasAdminPrivileges))
-            .sort((a1, a2) => a1.publishDate > a2.publishDate)
-            .slice(0, CompanyConstants.offers.max_profile_visible);
+            .getOffersByCompanyId(req.params.companyId, req.targetOwner, req.hasAdminPrivileges, {
+                sort: { publishDate: "desc" }, limit: CompanyConstants.offers.max_profile_visible
+            })
+        );
         return res.json({ company, offers });
     });
 
--- a/src/api/routes/company.js
+++ b/src/api/routes/company.js
@@ -75,14 +75,23 @@ export default (app) => {
 
     router.get("/:companyId",
         validators.profile,
-        (req, res, next) => companyMiddleware.restrictedAccess(req.params.companyId)(req, res, next),
-        (req, res, next) => companyMiddleware.registrationStatus(req.params.companyId)(req, res, next),
+        (req, res, next) => companyMiddleware.canAccessProfile(req.params.companyId)(req, res, next),
         async (req, res) => {
-            const company = await new CompanyService().findById(req.params.companyId, req.hasAdminPrivileges, req.hasAdminPrivileges);
-            const offers = (await new OfferService()
-                .getOffersByCompanyId(req.params.companyId, req.targetOwner, req.hasAdminPrivileges, {
-                    sort: { publishDate: "desc" }, limit: CompanyConstants.offers.max_profile_visible
-                })
+            const company = await new CompanyService().findById(
+                req.params.companyId,
+                // Can be safely set to true, as the middleware takes
+                // care of validation for us
+                true,
+                req.hasAdminPrivileges
+            );
+            const offers = await new OfferService().getOffersByCompanyId(
+                req.params.companyId,
+                req.targetOwner,
+                req.hasAdminPrivileges,
+                {
+                    sort: { publishDate: "desc" },
+                    limit: CompanyConstants.offers.max_profile_visible,
+                }
             );
             return res.json({ company, offers });
         }
--- a/src/api/routes/company.js
+++ b/src/api/routes/company.js
@@ -10,7 +10,6 @@ import CompanyService from "../../services/company.js";
 import { ErrorTypes } from "../middleware/errorHandler.js";
 import ValidationReasons from "../middleware/validators/validationReasons.js";
 import { concurrentOffersNotExceeded } from "../middleware/validators/validatorUtils.js";
-import CompanyConstants from "../../models/constants/Company.js";
 
 import { or } from "../middleware/utils.js";
 
@@ -87,11 +86,7 @@ export default (app) => {
             const offers = await new OfferService().getOffersByCompanyId(
                 req.params.companyId,
                 req.targetOwner,
-                req.hasAdminPrivileges,
-                {
-                    sort: { publishDate: "desc" },
-                    limit: CompanyConstants.offers.max_profile_visible,
-                }
+                req.hasAdminPrivileges
             );
             return res.json({ company, offers });
         }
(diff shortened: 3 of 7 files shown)


BrunoRosendo:
Is this ready for review?

toni-santos:
I think so, I'm not sure why the audit is failing, but the goals of the issue are all met!

src/api/routes/company.js
```diff
@@ -72,6 +72,15 @@ export default (app) => {
 
     });
 
+    router.get("/:companyId/profile", validators.profile, async (req, res) => {
+        const company = await new CompanyService().findById(req.params.companyId, req.hasAdminPrivileges, req.hasAdminPrivileges);
+        let offers = await new OfferService().getOffersByCompanyId(req.params.companyId, req.params.companyId, req.hasAdminPrivileges);
```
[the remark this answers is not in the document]
BrunoRosendo: You can also create a test for this case

test/end-to-end/company.js
```diff
@@ -227,6 +227,142 @@ describe("Company endpoint", () => {
...
+                expect(res.body.errors[0]).toHaveProperty("param", "companyId");
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.COMPANY_NOT_FOUND(id));
+            });
+
+            test("should return if the offers array is empty", async () => {
+                const res = await test_agent
+                    .get(`/company/${test_company_2.id}/profile`);
+                expect(res.body).toHaveProperty("offers");
+                expect(res.body.offers).toHaveProperty("length", 0);
+            });
```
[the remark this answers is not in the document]
BrunoRosendo: You should also check if the http code is OK (success)

test/end-to-end/company.js
```diff
@@ -227,6 +227,142 @@ describe("Company endpoint", () => {
...
+
+            test("should return offers lower than X (size: 3)", async () => {
+                await test_agent
+                    .post("/auth/login")
+                    .send(test_user_admin)
+                    .expect(HTTPStatus.OK);
+
+                const offers_2 = await Promise.all(Array(3).fill(null).map((_) =>
+                    Offer.create({
+                        ...generateTestOffer({
```
BrunoRosendo, line 318: I believe you can put all the properties inside `generateTestOffer` and it'll generate the object for you, so you don't need to wrap it and use `...`

test/end-to-end/company.js
```diff
@@ -227,6 +227,142 @@ describe("Company endpoint", () => {
...
+            });
+
+            test("should return if the offers array is empty", async () => {
+                const res = await test_agent
+                    .get(`/company/${test_company_2.id}/profile`);
+                expect(res.body).toHaveProperty("offers");
+                expect(res.body.offers).toHaveProperty("length", 0);
+            });
+
+            test("should return offers lower than X (size: 3)", async () => {
```
BrunoRosendo, line 310:
```suggestion
            test("should return all offers if below the maximum limit", async () => {
```

test/end-to-end/company.js
```diff
@@ -227,6 +227,142 @@ describe("Company endpoint", () => {
...
+                expect(res.body).toHaveProperty("offers");
+                expect(res.body.offers).toHaveProperty("length", 3);
+                offers_2.forEach((val, idx) => {
+                    expect(res.body.offers[idx].owner.toString()).toEqual(val.owner.toString());
+                    expect(res.body.offers[idx].ownerName.toString()).toEqual(val.ownerName.toString());
+                    expect(res.body.offers[idx].ownerLogo.toString()).toEqual(val.ownerLogo.toString());
+                });
+            });
+
+            test("should return if offers is X (size: 5)", async () => {
```
BrunoRosendo, line 338:
```suggestion
            test("should return all offers if they match the maximum limit, async () => {
```
Don't specify the size. And use the constant defined above here as well

test/end-to-end/company.js
BrunoRosendo: Use the constant defined above for this test too

test/end-to-end/company.js
```diff
@@ -227,6 +227,142 @@ describe("Company endpoint", () => {
...
+
+                const offers = await Promise.all(Array(5).fill(null).map((_) =>
+                    Offer.create({
+                        ...generateTestOffer({
+                            "publishDate": (new Date(Date.now() - (DAY_TO_MS))).toISOString(),
+                            "publishEndDate": (new Date(Date.now() + (DAY_TO_MS))).toISOString()
+                        }),
+                        owner: test_company._id,
+                        ownerName: test_company.name,
+                        ownerLogo: test_company.logo })).sort((a1, a2) => a1.publishDate > a2.publishDate));
```
BrunoRosendo, line 352:
```suggestion
                    Offer.create(generateTestOffer({
                            "publishDate": (new Date(Date.now() - (DAY_TO_MS))).toISOString(),
                            "publishEndDate": (new Date(Date.now() + (DAY_TO_MS))).toISOString(),
                            owner: test_company._id,
                            ownerName: test_company.name,
                            ownerLogo: test_company.logo })).sort((a1, a2) => a1.publishDate > a2.publishDate)),
                        }));
```

src/api/routes/company.js
```diff
@@ -72,6 +72,15 @@ export default (app) => {
...
+        offers = offers.sort((a1, a2) =>
+            a1.publishDate > a2.publishDate
+        ).slice(0, 5);
```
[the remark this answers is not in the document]
BrunoRosendo: Please don't hardcode the maximum offers. You can create a constant in models/Company, for example, and use it everywhere you need

test/end-to-end/company.js
BrunoRosendo: You can also create a test for when there are more offers than the limit and verify if the most recent ones are returned

BrunoRosendo (changes requested):
This is already looking good, nice job!

Something related to this but for another issue:
@imnotteixeira @DoStini Will the company profile also have infinite scrolling or will the offers just be limited? We already have an endpoint to return all the company offer, we could adapt that to return with a offset/limit (or using the same strategy as in the future new search). What do you think?

imnotteixeira:
> This is already looking good, nice job!
> 
> Something related to this but for another issue: @imnotteixeira @DoStini Will the company profile also have infinite scrolling or will the offers just be limited? We already have an endpoint to return all the company offer, we could adapt that to return with a offset/limit (or using the same strategy as in the future new search). What do you think?

Let's not insist on the offset/limit strategy, especially since we are implementing #129 with cursors.

I'd say we should not have infinite scrolling, but rather an explicit "load more" button which the user would click, but for the backend it would work the same way so it doesn't matter.

src/api/routes/company.js
BrunoRosendo, line 77:
```suggestion
        let offers = await new OfferService().getOffersByCompanyId(req.params.companyId, req.params.companyId, req.hasAdminPrivileges).current();
```
I think this is enough to only show current offers but please test it. This approach means that not even the owner sees unpublished offers on his profile but I think that makes sense, since we already have a `My Offers` page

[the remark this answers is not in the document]
toni-santos: I grasped the issue of using req.params.companyId instead of req.targetOwner, however I'm not sure what I should test for this case.

src/api/routes/company.js
```diff
@@ -72,6 +72,15 @@ export default (app) => {
 
     });
 
+    router.get("/:companyId/profile", validators.profile, async (req, res) => {
+        const company = await new CompanyService().findById(req.params.companyId, req.hasAdminPrivileges, req.hasAdminPrivileges);
+        const offers = (await new OfferService()
+            .getOffersByCompanyId(req.params.companyId, req.targetOwner, req.hasAdminPrivileges))
+            .sort((a1, a2) => a1.publishDate > a2.publishDate)
+            .slice(0, 5);
```
BrunoRosendo, line 80: You're missing the constant here

test/end-to-end/company.js
```diff
@@ -227,6 +227,182 @@ describe("Company endpoint", () => {
...
+                            ownerLogo: test_company_2.logo
+                        })
+                    )).sort((a1, a2) => a1.publishDate > a2.publishDate));
+
+
+                const res = await test_agent
+                    .get(`/company/${test_company_2._id}/profile`)
+                    .expect(HTTPStatus.OK);
+                expect(res.body).toHaveProperty("offers");
+                expect(res.body.offers.length).toBeLessThan(CompanyConstants.offers.max_profile_visible);
```
BrunoRosendo, line 338: Why not `toEqual` (max - 1) instead of `toBeLessThan`?

BrunoRosendo (changes requested):

src/api/routes/company.js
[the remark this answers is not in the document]
BrunoRosendo: You could create a hidden offer, log in as a random user (not the owner) and verify that the user does not see it when requesting the profile. You can take a look at these tests: https://github.com/NIAEFEUP/nijobs-be/blob/d4dfefa378063d3172c05810977876b0216a3420/test/end-to-end/offer.js#L2850

src/api/middleware/validators/company.js
```diff
@@ -75,6 +75,10 @@ export const deleteCompany = useExpressValidators([
     existingCompanyParamValidator,
 ]);
 
+export const profile = useExpressValidators([
+    existingCompanyParamValidator,
+]);
```
imnotteixeira, line 80: This means that when a company does not exists we will return a HTTP 422 error, instead of HTTP 404 (NOT FOUND) https://github.com/NIAEFEUP/nijobs-be/blob/d4dfefa378063d3172c05810977876b0216a3420/src/api/middleware/errorHandler.js#L57

We should not have this check on the validators, but on the route code itself, and throw the error with the correct status there (similar to what is done [here](https://github.com/NIAEFEUP/nijobs-be/blob/ca3c0e98c65297353e2bf74e5497f8cb82b22238/src/api/middleware/files.js#L30-L39))

src/api/routes/company.js
imnotteixeira, line 80: You should sort and limit on the query, not after getting stuff from DB

test/end-to-end/company.js
```diff
@@ -227,6 +227,182 @@ describe("Company endpoint", () => {
...
+                    .expect(HTTPStatus.UNPROCESSABLE_ENTITY);
+                expect(res.body).toHaveProperty("error_code", ErrorTypes.VALIDATION_ERROR);
+                expect(res.body).toHaveProperty("errors");
+                expect(res.body.errors[0]).toHaveProperty("param", "companyId");
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.COMPANY_NOT_FOUND(id));
+            });
+
+            /* test("", async () => {
+
+            });*/
```
imnotteixeira, line 306: remove

imnotteixeira (changes requested):

src/api/middleware/validators/company.js
Naapperas: I understand returning 404 instead of 422, but by this line of thought shouldn't every other route that has a `companyId` request parameter have the check on the route code and not on the validators ?

src/api/routes/company.js
toni-santos: I tried looking into doing this, however, `getOffersByCompanyId` currently does not return the mongoose query, and therefore I'm unable to use Mongoose to sort and limit it. I kept it as it was in this last commit, as I'm unsure if I should change the function or just keep it as it is.

src/api/routes/company.js
toni-santos: I believe this is related to what I pointed out in another review comment, but I was unable to do this since `getOffersByCompanyId` is currently not returning the query, hence `current()` will not be a usable function.

src/api/routes/company.js
Naapperas: Another solution could be to refactor the current method so that it returns a query instead of the results (correct me if I'm wrong but most of, if not all, service methods return queries), this way we would only have the "base" code written once (fetching offers based on the company that owns them) and then the route itself would do more processing in order to get the desired result. What do you think ?

BrunoRosendo: I agree it'd be best, if possible

toni-santos:
I believe the changes that were discussed previously have all been implemented, taking into account the restrictions and difficulties imposed by how the mongoose queries are currently working. Please alert me to any changes that should be made!

src/api/routes/offer.js
```diff
@@ -69,7 +69,8 @@ export default (app) => {
      */
     router.get("/company/:companyId", companyValidators.getOffers, async (req, res, next) => {
         try {
-            const offers = await (new OfferService()).getOffersByCompanyId(req.params.companyId, req.targetOwner, req.hasAdminPrivileges);
+            const offers = await (new OfferService())
+                .getOffersByCompanyId(req.params.companyId, req.targetOwner, req.hasAdminPrivileges, {});
```
DoStini, line 73: Shouldnt we have some kind of pagination in here? I believe we discussed in the meeting the user should be able to view all of the company offers by clicking somethign like "More offers". What kind of route parameters would be needed to implement this using the new keyset system? @BrunoRosendo 

But we should do this in a different issue

test/end-to-end/company.js
```diff
@@ -227,6 +228,229 @@ describe("Company endpoint", () => {
...
+                    })
+                );
+
+                await (new OfferService()).disable(disabled_test_offer._id, OfferConstants.HiddenOfferReasons.ADMIN_BLOCK);
+
+                const res = await test_agent
+                    .get(`/company/${disabled_company._id}/profile`)
+                    .expect(HTTPStatus.OK);
+                expect(res.body).toHaveProperty("offers");
+                expect(res.body.offers.length).toEqual(0);
```
DoStini, line 451: This name is missleading. The company is not disabled, one of its offers is. Also, what happens if we try to access a disabled or blocked company's page?

src/api/routes/offer.js
DoStini: Also, if this doesn't work how it used to work (calling the function without that empty object`{}`, you should add a default value to the function so that you don't have to use an empty object everywhere you need this

DoStini (changes requested):

src/api/routes/offer.js
BrunoRosendo: @DoStini It'd be a lot more simpler, since we wouldn't have search in the company's page. We'd just alter the company offers endpoint to use the same logic as in #212

imnotteixeira: What if we reused the offer search function, and passing the company we want to filter on (like if the user passed in that filter anyway). We would get the "pagination" for free after #212

test/end-to-end/company.js
```diff
@@ -450,6 +466,51 @@ describe("Company endpoint", () => {
                 expect(res.body).toHaveProperty("offers");
                 expect(res.body.offers.length).toEqual(0);
             });
+
+            test("should return if a disabled company returns 403", async () => {
```
Naapperas, line 470: Change the test description to "Should fail to retrieve details if company is disabled"

test/end-to-end/company.js
```diff
@@ -450,6 +466,51 @@ describe("Company endpoint", () => {
...
+                        generateTestOffer({
+                            "publishDate": (new Date(Date.now() - (DAY_TO_MS))).toISOString(),
+                            "publishEndDate": (new Date(Date.now() + (DAY_TO_MS))).toISOString(),
+                            owner: disabled_company._id,
+                            ownerName: disabled_company.name,
+                            ownerLogo: disabled_company.logo
+                        })
+                    )).sort((a1, a2) => a1.publishDate > a2.publishDate));
+
+                await test_agent
```
Naapperas, line 487: You should get the response object given by this method chain in order to perform further assertion: comparing the real error message with the expected one, etc...

test/end-to-end/company.js
```diff
@@ -450,6 +466,51 @@ describe("Company endpoint", () => {
...
+                        generateTestOffer({
+                            "publishDate": (new Date(Date.now() - (DAY_TO_MS))).toISOString(),
+                            "publishEndDate": (new Date(Date.now() + (DAY_TO_MS))).toISOString(),
+                            owner: blocked_company._id,
+                            ownerName: blocked_company.name,
+                            ownerLogo: blocked_company.logo
+                        })
+                    )).sort((a1, a2) => a1.publishDate > a2.publishDate));
+
+                await test_agent
```
Naapperas, line 509: Same as mentioned above.

test/end-to-end/company.js
```diff
@@ -450,6 +466,51 @@ describe("Company endpoint", () => {
...
+                            ownerLogo: disabled_company.logo
+                        })
+                    )).sort((a1, a2) => a1.publishDate > a2.publishDate));
+
+                await test_agent
+                    .get(`/company/${disabled_company._id}/profile`)
+                    .expect(HTTPStatus.FORBIDDEN);
+            });
+
+            test("should return if a blocked company returns 403", async () => {
```
Naapperas, line 492: Same as mentioned above but regarding blocked companies.

Naapperas (changes requested):

test/end-to-end/company.js
```diff
@@ -227,6 +228,290 @@ describe("Company endpoint", () => {
...
+                            ownerName: test_company_3.name,
+                            ownerLogo: test_company_3.logo
+                        })
+                    )).sort((a1, a2) => a1.publishDate > a2.publishDate));
+
+                const res = await test_agent
+                    .get(`/company/${test_company_3._id}/profile`)
+                    .expect(HTTPStatus.OK);
+                expect(res.body).toHaveProperty("offers");
+                expect(res.body.offers.length).toEqual(CompanyConstants.offers.max_profile_visible);
```
Naapperas, line 415: We should also assert that the company information returned in the response is correct, not just the offers.

test/end-to-end/company.js
```diff
@@ -227,6 +228,290 @@ describe("Company endpoint", () => {
...
+                for (let idx = 0; idx < CompanyConstants.offers.max_concurrent; idx++) {
+                    const val = offers[idx];
+
+                    expect(res.body.offers[idx].owner.toString()).toEqual(val.owner.toString());
+                    expect(res.body.offers[idx].ownerName.toString()).toEqual(val.ownerName.toString());
+                    expect(res.body.offers[idx].ownerLogo.toString()).toEqual(val.ownerLogo.toString());
+                }
+            });
+
+            test("should return if a random user is unable to see hidden offers", async () => {
```
Naapperas, line 425: Change this test description to something like "Should return no hidden offers when viewing company profile as unprivileged user"

test/end-to-end/company.js
```diff
@@ -227,6 +228,290 @@ describe("Company endpoint", () => {
...
+                    .get(`/company/${id}/profile`)
+                    .send({ adminReason })
+                    .expect(HTTPStatus.UNPROCESSABLE_ENTITY);
+                expect(res.body).toHaveProperty("error_code", ErrorTypes.VALIDATION_ERROR);
+                expect(res.body).toHaveProperty("errors");
+                expect(res.body.errors[0]).toHaveProperty("param", "companyId");
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.COMPANY_NOT_FOUND(id));
+            });
+
+            test("should succeed if the company has no offers", async () => {
```
Naapperas, line 327: This test description may be misleading: one might think that having no offers => success

test/end-to-end/company.js
```diff
@@ -227,6 +228,290 @@ describe("Company endpoint", () => {
...
+                const res = await test_agent
+                    .get(`/company/${test_company_2.id}/profile`)
+                    .expect(HTTPStatus.OK);
+                expect(res.body).toHaveProperty("offers");
+                expect(res.body.offers).toHaveProperty("length", 0);
+            });
+
+            test("should return all offers if below the maximum limit", async () => {
```
Naapperas, line 335: "Should return all company offers when company has less offers than the limit" or something similar.

test/end-to-end/company.js
```diff
@@ -227,6 +228,290 @@ describe("Company endpoint", () => {
...
+                expect(res.body).toHaveProperty("offers");
+                expect(res.body.offers.length).toEqual(CompanyConstants.offers.max_profile_visible - 1);
+                offers.forEach((val, idx) => {
+                    expect(res.body.offers[idx].owner.toString()).toEqual(val.owner.toString());
+                    expect(res.body.offers[idx].ownerName.toString()).toEqual(val.ownerName.toString());
+                    expect(res.body.offers[idx].ownerLogo.toString()).toEqual(val.ownerLogo.toString());
+                });
+            });
+
+            test("should return all offers if they match the maximum limit", async () => {
```
Naapperas, line 365: Same as above

test/end-to-end/company.js
```diff
@@ -227,6 +228,290 @@ describe("Company endpoint", () => {
...
+                expect(res.body).toHaveProperty("offers");
+                expect(res.body.offers.length).toEqual(CompanyConstants.offers.max_profile_visible);
+                offers.forEach((val, idx) => {
+                    expect(res.body.offers[idx].owner.toString()).toEqual(val.owner.toString());
+                    expect(res.body.offers[idx].ownerName.toString()).toEqual(val.ownerName.toString());
+                    expect(res.body.offers[idx].ownerLogo.toString()).toEqual(val.ownerLogo.toString());
+                });
+            });
+
+            test("should return the maximum number of offers to be displayed if there are more offers than it", async () => {
```
Naapperas, line 394: Same as above

Naapperas (changes requested):

src/api/middleware/company.js
```diff
@@ -85,6 +85,69 @@ export const profileComplete = async (req, res, next) => {
...
+                { company: company }
+            );
+        } else {
+            error = new APIError(
+                HTTPStatus.FORBIDDEN,
+                ErrorTypes.FORBIDDEN,
+                ValidationReasons.NOT_FOUND
+            );
+        }
+    }
```
Naapperas, line 124: There is a lot of code that is repeated, I would suggest doing something like this:

```suggestion
    if (req.params?.companyId === req.user.company) {
        error = new APIError(
            HTTPStatus.OK,
            ErrorTypes.VALIDATION_ERROR,
            company.isBlocked ? ValidationReasons.COMPANY_BLOCKED : company.isDisabled ? ValidationReasons.COMPANY_DISABLED : ValidationReasons.UNKNONW, // this here can be any other validation reason, put UNKOWN as an example
            { company: company }
        );
    } else {
        error = new APIError(
            HTTPStatus.FORBIDDEN,
            ErrorTypes.FORBIDDEN,
            ValidationReasons.NOT_FOUND
        );
    }
```

This will give priority to the blocked information so we should think more about how to handle this.

Naapperas (changes requested):

src/api/middleware/company.js
```diff
@@ -85,6 +85,69 @@ export const profileComplete = async (req, res, next) => {
...
+        ));
+    }
+
+    if (!company.hasFinishedRegistration && !(req.params?.companyId === req.user.company)) {
+        return next(new APIError(
+            HTTPStatus.FORBIDDEN,
+            ErrorTypes.FORBIDDEN,
+            ValidationReasons.NOT_FOUND
+        ));
+    }
```
Naapperas, line 146:
```suggestion
    if (!company.hasFinishedRegistration) {
        return next(new APIError(
            HTTPStatus.FORBIDDEN,
            ErrorTypes.FORBIDDEN,
            (req.params?.companyId === req.user.company) ? ValidationReasons.REGISTRATION_NOT_FINISHED : ValidationReasons.NOT_FOUND
        ));
    }
```

Naapperas (changes requested):

Naapperas:
Also, be aware of the file conflicts due to this branch not being "up-to-date"

src/api/middleware/company.js
BrunoRosendo: The chain is a bit confusing, at least put brackets to differentiate a bit:
```js
company.isBlocked ? ValidationReasons.COMPANY_BLOCKED : 
  (company.isDisabled ? ValidationReasons.COMPANY_DISABLED : ValidationReasons.UNKNONW)
```

BrunoRosendo:
The assignee will change

src/api/routes/company.js
```diff
@@ -72,6 +73,21 @@ export default (app) => {
 
     });
 
+    router.get("/:companyId/profile",
```
Naapperas, line 76:
```suggestion
    router.get("/:companyId",
```

Naapperas (changes requested):

ttoino:
The tests should now align with the implementation goals, right now the following restrictions are in place:

- When the company is disabled, blocked or has not finished registration, the same error as when the company does not exist is returned;
- When the company is disabled and the user is the owner, admin or god it succeeds;
- When the company is blocked and the user is the owner an error is returned;
- When the company is blocked and the user is an admin or god it succeeds;
- When the company has not finished registration and the user is the owner, admin or god an error is returned (not sure about this one).

If any of these don't make sense please let me know.

Naapperas:
> The tests should now align with the implementation goals, right now the following restrictions are in place:
> 
>     * When the company is disabled, blocked or has not finished registration, the same error as when the company does not exist is returned;
> 
>     * When the company is disabled and the user is the owner, admin or god it succeeds;
> 
>     * When the company is blocked and the user is the owner an error is returned;
> 
>     * When the company is blocked and the user is an admin or god it succeeds;
> 
>     * When the company has not finished registration and the user is the owner, admin or god an error is returned (not sure about this one).
> 
> 
> If any of these don't make sense please let me know.

I think that not having finnished its registratio should not be a restriction on company fetching. What do you think @BrunoRosendo @DoStini ?

BrunoRosendo:
> > The tests should now align with the implementation goals, right now the following restrictions are in place:
> > 
> >     * When the company is disabled, blocked or has not finished registration, the same error as when the company does not exist is returned;
> > 
> >     * When the company is disabled and the user is the owner, admin or god it succeeds;
> > 
> >     * When the company is blocked and the user is the owner an error is returned;
> > 
> >     * When the company is blocked and the user is an admin or god it succeeds;
> > 
> >     * When the company has not finished registration and the user is the owner, admin or god an error is returned (not sure about this one).
> > 
> > 
> > If any of these don't make sense please let me know.
> 
> I think that not having finnished its registratio should not be a restriction on company fetching. What do you think @BrunoRosendo @DoStini ?
> 
> 

What information would we have in that case? Not sure if it's useful for a company profile but, either way, we should have the same restriction as in the "get all companies" endpoint

DoStini:
Im my opinion, if a company didnt finish their registration, is like of they do not exist in the platform at all. So I think we should 404 Not Found for hidden companies and not finished

test/end-to-end/company.js
```diff
@@ -226,7 +226,571 @@ describe("Company endpoint", () => {
...
+                        expect.objectContaining({
+                            param: "companyId",
+                            msg: ValidationReasons.COMPANY_NOT_FOUND(
+                                test_company_with_unfinished_registration._id
+                            ),
+                        }),
+                    ])
+                );
+            });
+        });
```
Naapperas, line 562: Bump on a comment made above, this should return a 404.

Also, the 'getAllCompanies' route returns companies that have not finished their registration, which is the opposite oh what this route is doing, what should be changed ?

Naapperas (changes requested):
Besides the comments I made, this looks good to me. Just be careful to rebase this branch with develop.

ttoino:
I decided to leave the implementation as is and created a new issue to deal with the inconsistent http statuses (#282).

src/api/middleware/company.js
```diff
@@ -85,6 +85,45 @@ export const profileComplete = async (req, res, next) => {
...
+            HTTPStatus.UNPROCESSABLE_ENTITY,
+            ErrorTypes.VALIDATION_ERROR,
+            [
+                {
+                    value: companyId,
+                    msg: ValidationReasons.COMPANY_NOT_FOUND(companyId),
+                    param: "companyId",
+                    location: "params",
+                },
+            ]
```
Naapperas, line 102: @ttoino can you explain why you are setting the error directly here? I recall you telling me that already but I've forgotten it.

ttoino: I wanted the endpoint to return the same error as when the company does not exist, right now the only way is to do that (I also hate it), but after #282 we could just return a 404.

src/api/middleware/company.js
```diff
@@ -85,6 +85,45 @@ export const profileComplete = async (req, res, next) => {
     return next();
 };
 
+export const canAccessProfile = (companyId) => async (req, res, next) => {
+    const company = await new CompanyService().findById(companyId, true);
+
+    const notFound = () =>
+        new APIError(
+            HTTPStatus.UNPROCESSABLE_ENTITY,
```
Naapperas, line 93: Shouldn't this be "NOT_FOUND" ?

ttoino: Same reason as above, after #282 it would be NOT_FOUND.

src/api/middleware/company.js
BrunoRosendo: I find this a bit confusing but it could be improved in #282

src/api/routes/company.js
```diff
@@ -72,6 +73,30 @@ export default (app) => {
...
+                req.hasAdminPrivileges
+            );
+            const offers = await new OfferService().getOffersByCompanyId(
+                req.params.companyId,
+                req.targetOwner,
+                req.hasAdminPrivileges,
+                {
+                    sort: { publishDate: "desc" },
+                    limit: CompanyConstants.offers.max_profile_visible,
+                }
```
BrunoRosendo, line 94: We should perhaps do this similarly to what's done in offer service's get, that it, pass sort and limit as explicit parameters. What do you think?

BrunoRosendo (changes requested):

src/api/routes/company.js
Naapperas: Changing that would just be a matter of consistency, right?

BrunoRosendo: yes, behaviour is the same but, currently, you're depending on Mongoose's implementation of filters

Naapperas: Oh yeah, you're totally right.

Naapperas:
@ttoino #278 was merged, if you rebase it should clear most of the audit issues.

test/end-to-end/company.js
```diff
@@ -228,7 +228,571 @@ describe("Company endpoint", () => {
...
+                    "company._id",
+                    test_company_without_offers._id.toString()
+                );
+            });
+
+            test("should return all offers when below limit", async () => {
+                const offers = await createTestOffers(
+                    CompanyConstants.offers.max_profile_visible - 1,
+                    test_company_with_offers_below_limit
+                );
```
Naapperas, line 419: I do not feel too confortable with this style of creating offers but it is not technically wrong, in my opinion.

Naapperas (dismissed):
Other than the pointed out code, this looks good to me.

src/api/routes/company.js
BrunoRosendo: I think we're just missing this, then I'll approve

BrunoRosendo (dismissed):

Naapperas (dismissed):

DoStini (dismissed):

Naapperas (approved):

BrunoRosendo (approved):

PR #208, opened 2022-02-18
Fix vacancies without minimum value

### LINKED ISSUE #206: Vacancies minimum value
Add a validator to the `vacancies` field to be a minimum of 1

BrunoRosendo:
closes #206


test/utils/ValidatorTester.js
```diff
@@ -332,6 +332,26 @@ const ValidatorTester = (requestEndpoint) => (location) => (field_name) => ({
             });
         });
     },
+
+    mustBeGreaterThanOrEqualTo: (min) => {
```
DoStini, line 336: Don't we have any case where we test the max one?

BrunoRosendo: I don't think so. We don't have any case for that one, so I don't think we should add the code. If necessary, It's very easy to create it by looking at this one

src/models/Offer.js
```diff
@@ -62,7 +62,7 @@ const OfferSchema = new Schema({
     },
 
     isPaid: { type: Boolean },
-    vacancies: { type: Number },
+    vacancies: { type: Number, min: OfferConstants.vacancies.min },
```
imnotteixeira, line 65: Does this allow an empty value as well?

package.json
```diff
@@ -48,7 +48,7 @@
     "morgan": "^1.9.1",
     "multer": "^1.4.2",
     "nodemailer": "^6.7.2",
-    "nodemailer-express-handlebars": "^4.0.0",
+    "nodemailer-express-handlebars": "^5.0.0",
```
imnotteixeira, line 51: we did it, nice :P

src/models/Offer.js
BrunoRosendo: I took this form the docs:
`Validators are not run on undefined values. The only exception is the [required validator](https://mongoosejs.com/docs/api.html#schematype_SchemaType-required).` 

[here](https://mongoosejs.com/docs/schematypes.html#number-validators)

src/models/constants/Offer.js
```diff
@@ -12,6 +12,9 @@ export default Object.freeze({
     requirements: {
         min_length: 1,
     },
+    vacancies: {
+        min: 1,
```
imnotteixeira, line 16: It should probably be allowed to be 0. Thinking for the future if we ever update offers' vacancies, they can end up with 0 vacancies. With the current code, the DB wouldn't let that.

imnotteixeira (changes requested):

src/models/constants/Offer.js
BrunoRosendo: Should I update only the DB or the validators as well?

imnotteixeira: DB at least, validators if you want

imnotteixeira (approved):

PR #209, opened 2022-02-19
Removed offer grace period

### LINKED ISSUE #204: Let Offers be edited regardless of grace period
The idea behind having this check was for offers to not be edited after a long period and mislead students. Considering companies can just delete them and recreate the edited version, this makes no sense, so we can just let them have it.

CarlosMealha:
closes #204

### DIFF
--- a/src/api/middleware/validators/offer.js
+++ b/src/api/middleware/validators/offer.js
@@ -12,8 +12,6 @@ import OfferConstants from "../../../models/constants/Offer.js";
 import Company from "../../../models/Company.js";
 import Offer, { validatePublishEndDateLimit } from "../../../models/Offer.js";
 import {
-    HOUR_IN_MS,
-    OFFER_EDIT_GRACE_PERIOD_HOURS,
     MONTH_IN_MS,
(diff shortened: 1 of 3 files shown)


imnotteixeira:
Looks good, try running `npm audit fix` to try to fix the audit problem (make sure you are running with node 14)

BrunoRosendo:
I've tried to fix the audit but nodemailer-express-handlebars hasn't been updated yet (vulnerability in its only dependency). Should we just wait? @imnotteixeira

imnotteixeira:
Let's see if this is solved this week. I pinged the package owner https://github.com/yads/nodemailer-express-handlebars/pull/52#issuecomment-1046219399

Couldn't find an email, and seems like he's not active on twitter, so this is our best shot.

If this isn't solved, we can try bumping our node and npm versions, since in npmv8 we have an `overrides` option to override transitive dependency versions, to solve this.

DoStini:
I don't think the author of that published the most recent update yet

BrunoRosendo:
He updated in GitHub but not yet in npm :/

BrunoRosendo (approved):

PR #211, opened 2022-02-25
Added location router

### LINKED ISSUE #199: Location Service Endpoint
Currently, we are using Geodb Cities API, which is very incomplete. For example, Aveiro is not available.

We should implement our own endpoint to search for possible locations

Naapperas:
Closes #199

### DIFF
--- a/src/api/middleware/validators/location.js
+++ b/src/api/middleware/validators/location.js
@@ -8,6 +8,7 @@ import LocationConstants from "../../../models/constants/Location.js";
 export const search = useExpressValidators([
     query("searchTerm", ValidationReasons.DEFAULT)
         .isString().withMessage(ValidationReasons.STRING).bail()
-        .isLength({ min: 3 }).withMessage(ValidationReasons.LOCATION_SEARCH_TERM_MIN_LENGTH(LocationConstants.searchTerm.minLength)),
+        .isLength({ min: LocationConstants.searchTerm.minLength })
+        .withMessage(ValidationReasons.TOO_SHORT(LocationConstants.searchTerm.minLength)),
 
 ]);
--- a/src/api/middleware/validators/location.js
+++ b/src/api/middleware/validators/location.js
@@ -3,6 +3,19 @@ import { query } from "express-validator";
 import { useExpressValidators } from "../errorHandler.js";
 import ValidationReasons from "./validationReasons.js";
 
+const mustHaveLongitude = (latitude, { req }) => {
+
+    const { longitude } = req.query;
+
+    return !!longitude;
+};
+
+const mustHaveLatitude = (longitude, { req }) => {
+
+    const { latitude } = req.query;
+
+    return !!latitude;
+};
 
 export const search = useExpressValidators([
     query("city", ValidationReasons.DEFAULT)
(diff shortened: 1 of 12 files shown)


src/api/middleware/validators/location.js
```diff
@@ -0,0 +1,13 @@
...
+
+import { useExpressValidators } from "../errorHandler.js";
+import ValidationReasons from "./validationReasons.js";
+
+import LocationConstants from "../../../models/constants/Location.js";
+
+export const search = useExpressValidators([
+    query("searchTerm", ValidationReasons.DEFAULT)
+        .isString().withMessage(ValidationReasons.STRING).bail()
+        .isLength({ min: 3 }).withMessage(ValidationReasons.LOCATION_SEARCH_TERM_MIN_LENGTH(LocationConstants.searchTerm.minLength)),
```
imnotteixeira, line 11:
```suggestion
        .isLength({ min: LocationConstants.searchTerm.minLength }).withMessage(ValidationReasons.LOCATION_SEARCH_TERM_MIN_LENGTH(LocationConstants.searchTerm.minLength)),
```

imnotteixeira: There is a `TOO_SHORT` validation reason that should be reused. THe error object will already tell you which field it is, no need to create more and more reasons.

src/loaders/mongoose.js
```diff
@@ -10,25 +10,46 @@ export default async () => {
...
         await createDefaultAdmin();
     } catch (err) {
         console.error("Mongoose: failure in initial connection to the DB (aborting, will not retry)", err);
         process.exit(44);
     }
 
-    mongoose.connection.on("error", (err) => {
-        console.error("Mongoose connection error:", err);
+    mongoose.connections.forEach((connection) => {
+        connection.on("error", (err) => { // default mongoose connection used for the main app DB
```
imnotteixeira, line 50: Is the comment accurate? Shouldn't this be iterating all the connections? Inc. the location one?

src/models/Location.js
```diff
@@ -0,0 +1,42 @@
...
+    }
+});
+
+LocationSchema.index(
+    { citySearch: "text", countrySearch: "text" },
+    { name: "Search index", weights: { citySearch: 10, countrySearch: 5 } }
+);
+
+// this works because mongoose caches commands until a valid connection can be established or a timeout error is raised
+const Location = mongoose.connection.useDb(config.location_db_name).model("Location", LocationSchema);
```
imnotteixeira, line 40: We need to pass a `{useCache: true}` to the useDb options

imnotteixeira (changes requested):

src/loaders/mongoose.js
Naapperas: I just copied it over, forgot to change/remove it

Naapperas: Also, before this I explicitely called the on("error") on both the connection and connections[1] objects, so this is in fact iterating both connections.

Naapperas:
After discussing some issues regarding the implementations of these features, it has been concluded that this behavior is best achieved in another application instead of being built into NIJobs, so this PR will be closed after those features have been implemented elsewhere.

PR #212, opened 2022-03-01
Changed search system to keyset-based

### LINKED ISSUE #129: Improve search performance

BrunoRosendo:
closes #129 

The new package base64url is capable of encoding a token in a way that's safe to write in a URL


Naapperas (dismissed):

src/api/middleware/validators/offer.js
```diff
@@ -452,10 +454,30 @@ export const setDefaultValuesCreate = (req, res, next) => {
 };
 
 export const get = useExpressValidators([
-    query("offset", ValidationReasons.DEFAULT)
+    query("lastOfferId", ValidationReasons.DEFAULT)
```
imnotteixeira, line 457: Let´s make this generic. In the future we might not want to make the id come from offer ids.

`queryToken` is just a suggestion, you can use something else

```suggestion
    query("queryToken", ValidationReasons.DEFAULT)
```

src/models/Offer.js
```diff
@@ -146,22 +146,24 @@ function validateDescription(value) {
 /**
  * Currently active Offers (publish date was before Date.now and end date is after Date.now)
  */
+OfferSchema.statics.currentCondition = () => ({
```
imnotteixeira, line 149:
```suggestion
OfferSchema.statics.filterCurrent = () => ({
```

src/models/Offer.js
```diff
@@ -146,22 +146,24 @@ function validateDescription(value) {
...
-            $gt: new Date(Date.now()),
-        },
-    });
+    return this.where(this.model.currentCondition());
 };
 
 /**
  * Currently active and non-hidden Offers
  */
+OfferSchema.statics.withoutHiddenCondition = () => ({ isHidden: false });
```
imnotteixeira, line 164:
```suggestion
OfferSchema.statics.filterNonHidden = () => ({ isHidden: false });
```

src/api/middleware/validators/offer.js
```diff
@@ -452,10 +454,30 @@ export const setDefaultValuesCreate = (req, res, next) => {
...
+                const { value, ...filters } = req.query;
+
+                if (!await offerService.doesOfferMatchFilters(offerId, filters)) {
+                    throw new Error(ValidationReasons.OFFER_NOT_MATCHING_CRITERIA);
+                }
+
+                if (!value) return true;
+
+                const score = await offerService.getTextSearchScoreById(offerId, value);
+                if (score <= 0) throw new Error(ValidationReasons.OFFER_NOT_MATCHING_CRITERIA);
```
imnotteixeira, line 473: Why are we throwing when something does not match?

Say we searched for the first batch of offers. The last one was modified, and no longer matches the filters. We won't be able to show more results in frontend because of this, which shouldn't happen. Why does the filter matching impact the next search? Even if some result no longer matches the filtering, that's not a problem as it was returned as part of a previous state. We shouldn't worry about it.

src/api/middleware/validators/offer.js
```diff
@@ -452,10 +454,30 @@ export const setDefaultValuesCreate = (req, res, next) => {
...
-        .toInt(),
+        .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
+        .custom(existingOfferId)
+        .custom(async (offerId, { req }) => {
+            try {
+                const offerService = new OfferService();
+                const { value, ...filters } = req.query;
+
+                if (!await offerService.doesOfferMatchFilters(offerId, filters)) {
+                    throw new Error(ValidationReasons.OFFER_NOT_MATCHING_CRITERIA);
```
imnotteixeira, line 467: same thing as below

src/api/middleware/validators/offer.js
```diff
@@ -452,10 +454,30 @@ export const setDefaultValuesCreate = (req, res, next) => {
...
 
 export const get = useExpressValidators([
-    query("offset", ValidationReasons.DEFAULT)
+    query("lastOfferId", ValidationReasons.DEFAULT)
         .optional()
-        .isInt({ min: 0 }).withMessage(ValidationReasons.INT)
-        .toInt(),
+        .custom(isObjectId).withMessage(ValidationReasons.OBJECT_ID).bail()
+        .custom(existingOfferId)
+        .custom(async (offerId, { req }) => {
```
imnotteixeira, line 461: if this validator is only for the `OFFER_NOT_MATCHING_CRITERIA` check, then we can probably remove it

src/services/offer.js
```diff
@@ -192,29 +195,64 @@ class OfferService {
...
+
+            offers = Offer.aggregate([
+                { $match: { $text: { $search: value } } },
+                { $addFields: {
+                    score: { $meta: "textScore" },
+                    adminReason: { $cond: [showAdminReason, "$adminReason", "$$REMOVE"] }
+                } },
+                { $match: { "$or": [
+                    { score: { "$lt": lastOfferScore } },
+                    { score: lastOfferScore, _id: { "$gt": ObjectId(lastOfferId) } }
```
imnotteixeira, line 221: What if there are more offers now that have a bigger score than the last score? They wouldn't be returned

imnotteixeira: Also, why do we care about the score? Can't we just sort by the id?

imnotteixeira (changes requested):

src/services/offer.js
BrunoRosendo: If we do that, the offers best matching the search won't be at the top, which isn't ideal

src/api/middleware/validators/offer.js
BrunoRosendo: I did this to validate a poorly made request. If the offer does not match the full text search, we won't get a positive score and we can't get more offers. In the case the offer is edited, we wouldn't get more offers even if the we remove the validator, right?

BrunoRosendo: I thought it'd be best for the FE to know what happened instead of just not receiving anything else

src/services/offer.js
BrunoRosendo: Also, do we want new offers to be returned in the infinite scroll? That's not the approach I was following. In this article, they prioritize consistency over new items: https://medium.com/swlh/mongodb-pagination-fast-consistent-ece2a97070f3

src/api/middleware/validators/offer.js
imnotteixeira: I understand. But still, I think we can optimize a bit: let's send an encoded string (our `queryToken`) that contains information about both the last offer's id and the score. This way we don't have to query the db every time just to get the score. It also prevents the problem I described: if we know the score threshold and the last id, we can continue the search even if that last offer no longer exists or is not part of the results anymore.

So yes, we sen the score to the frontend, but it doesn't know about it, it's just a random string that you just send back to continue fetching stuff. The queryToken could literally be "this is the id: askdsald, and this is the last score: score" and the frontend wouldn't care, it would just send it back the next time.

For encoding/decoding, it's probably simple enough to use base64 of a stringified json with id and score, but feel free to think about other (possibly smaller) options

src/services/offer.js
imnotteixeira: Ok, leave it as is then

src/api/middleware/validators/offer.js
BrunoRosendo: @imnotteixeira I just pushed the new changes

src/services/offer.js
```diff
@@ -192,29 +196,69 @@ class OfferService {
...
-            .sort(value ? { score: { "$meta": "textScore" } } : undefined)
-            .skip(offset)
+        const results = await offers
+            .sort(value ? { score: { "$meta": "textScore" }, _id: 1 } : { _id: 1 })
             .limit(limit)
         ;
 
-        return showAdminReason ? offersQuery : offersQuery.select("-adminReason");
-
+        return results.map(this.buildQueryToken);
```
imnotteixeira, line 259: no need to have the token per offer, simply return an object instead with two fields: the search `results` and the `queryToken`. Notify whoever is working on the frontend to update the way they receive the search results (accessing the results array, instead of the offers directly)

imnotteixeira (changes requested):

imnotteixeira:
@BrunoRosendo please update the route function, the comment is outdated here

https://github.com/NIAEFEUP/nijobs-be/blob/develop/src/api/routes/offer.js#L22-L25

also, we should rename the variable, since the new .get function no longer returns offers only (also the token now)

https://github.com/NIAEFEUP/nijobs-be/blob/develop/src/api/routes/offer.js#L28

package.json
```diff
@@ -32,6 +32,7 @@
   "dependencies": {
     "@babel/plugin-proposal-optional-chaining": "^7.12.7",
     "babel": "^6.23.0",
+    "base64url": "^3.0.1",
```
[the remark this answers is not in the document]
BrunoRosendo: Base64 is not safe for URLs (and the token is passed in the query by the FE). The package implements an altered version of the algorithm which is URL safe. You can read more [here](https://en.wikipedia.org/wiki/Base64#RFC_4648)

src/services/offer.js
```diff
@@ -262,6 +306,33 @@ class OfferService {
         return constraints.length ? { "$and": constraints } : {};
     }
 
+    /**
+     * Builds a query token, by taking the offer's id and FTS score (if present) and decoding in safe base64
```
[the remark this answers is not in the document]
BrunoRosendo: Answered above

package.json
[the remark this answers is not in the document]
imnotteixeira: Ah, got it

src/services/offer.js
BrunoRosendo: @TiagooGomess Can you take care of this in https://github.com/NIAEFEUP/nijobs-fe/pull/172 ?

TiagooGomess: Sure

TiagooGomess: It's already done in https://github.com/NIAEFEUP/nijobs-fe/pull/172. This change facilitates my work by some orders of magnitude.

src/services/offer.js
```diff
@@ -192,29 +196,72 @@ class OfferService {
...
+     * queryToken: Id of the last fetched offer
      * limit: How many offers to show
      * jobType: Array of jobTypes allowed
      */
-    get({ value = "", offset = 0, limit = OfferService.MAX_OFFERS_PER_QUERY, showHidden = false, showAdminReason = false, ...filters }) {
-
-        const offers = (value ? Offer.find(
-            { "$and": [this._buildFilterQuery(filters), { "$text": { "$search": value } }] }, { score: { "$meta": "textScore" } }
-        ) : Offer.find(this._buildFilterQuery(filters))).current();
+    async get({ value = "", queryToken = null, limit = OfferService.MAX_OFFERS_PER_QUERY,
```
imnotteixeira, line 207: Can we support a different `sort by` value? I'm thinking we might want to sort results other than by `relevance`

src/services/offer.js
```diff
@@ -192,29 +196,72 @@ class OfferService {
...
+            offers.current();
+            if (!showHidden) offers.withoutHidden();
+            if (!showAdminReason) offers.select("-adminReason");
+        }
 
-        const offersQuery = offers
-            .sort(value ? { score: { "$meta": "textScore" } } : undefined)
-            .skip(offset)
+        const results = await offers
+            .sort(value ? { score: { "$meta": "textScore" }, _id: 1 } : { _id: 1 })
```
imnotteixeira, line 255: Is there any way can we support a different `sort by` value? I'm thinking we might want to sort results other than by `relevance`

BrunoRosendo: I think we can sort by whatever field we want, we just need an extra prop in the query and then we can choose different properties for the sort

I can create an issue for that, as it seems like a different matter

imnotteixeira: But we'd have to send the sort by in the queryToken as well right? Otherwise we wouldn't know when loading more, or it could possibly be different

BrunoRosendo: I was thinking we could let FE choose the prop it wants to sort for (query) and then use it to encode/decode the queryToken

imnotteixeira: That's weird, what does it mean "use it to encode/decode"? If the backend doesn't know what it is, hiw will it decode it?

Also, changing the sortby creates a new search, so we cannot search, change the sortby and send the first queryToken, it's like using different query filters.

So the sort by would neet to be part of the queryToken, and be used by the backend and the FE doesn't know anything about it, except that it will send it on first search requests, and every time it changes, issue a new (different) search request

BrunoRosendo: Well yes, changing the sortby would need a new search. I was saying something like this:

FE: sortBy: `field`
BE:
- Sorts the results by `field` and the last element has `field: X`
- Constructs the queryToken as `{id: 123, sortField: X}`

And then the flow would be the same as currently. But the FE would need a new search to change the sortBy

imnotteixeira: Exactly, it's only a matter of making the sortby generic. Currently it's score, but it could be other fields

BrunoRosendo: Agreed. But I think we should release with infinite scrolling as it is and implement this in a separate PR

imnotteixeira: Sure, just wanted to make sure this could be easily extended, and it seems like it is.

Please create the issue for visibility

TiagooGomess:
@BrunoRosendo https://github.com/NIAEFEUP/nijobs-fe/pull/172 is ready, is this also ready?

BrunoRosendo:
> @BrunoRosendo https://github.com/NIAEFEUP/nijobs-fe/pull/172 is ready, is this also ready? When we merge these, we need to do it at the same time.

Yes, let's merge them!

BrunoRosendo:
@imnotteixeira Could you please re-review? I can't request it, I'm not sure why

src/services/offer.js
imnotteixeira, line 207: This function is getting too big (around 40 lines). Can we extract its contents into smaller functions to make it more readable?

src/services/offer.js
```diff
@@ -262,6 +309,30 @@ class OfferService {
...
     }
 
+    /**
+     * Encodes a query token, by taking the offer's id and FTS score (if present) and decoding in safe url base64
+     * @param {*} offer
+     */
+    encodeQueryToken(offer) {
+        return base64url.encode(JSON.stringify({
+            id: offer._id,
+            score: offer.score || offer._doc?.score
```
imnotteixeira, line 319: Why the need for this OR? You shouldn't be accessing `_doc`

imnotteixeira (changes requested):

src/services/offer.js
BrunoRosendo: If I remember correctly, the query returned by `find()` includes the score attribute. However, `aggregate()`returns an instance of the `Aggregate` class, in which I could only access the score by checking `_doc`

BrunoRosendo: And I couldn't filter by textScore in a normal `find()`

src/services/offer.js
BrunoRosendo: @imnotteixeira Refactored

src/services/offer.js
imnotteixeira: the encode function should not depend on internal implementations, it should receive the `id` and `score` directly, and build the token

src/services/offer.js
```diff
@@ -212,44 +212,10 @@ class OfferService {
...
-                ] }, { score: { "$meta": "textScore" } }
-
-                ) : Offer.find(this._buildFilterQuery(filters)));
-            }
-
-            offers.current();
-            if (!showHidden) offers.withoutHidden();
-            if (!showAdminReason) offers.select("-adminReason");
-        }
+        const offers = lastOfferId && value ?
```
imnotteixeira, line 215: Thinking about this, shouldn't the query token include all the search information (i.e. value and filters)? I don't think we should require the client to send the value again if they are already sending a query token to "continue the last search", whatever it is, and which includes certain filters. What if the filters changed now? It does not make sense to treat it as a follow-up search, as the score we use to get next results will have a different meaning.

src/services/offer.js
```diff
@@ -262,6 +228,59 @@ class OfferService {
...
+
+        return offers;
+    }
+
+    /**
+     * Builds a search aggregation with aggregate().
+     * Only use this when loading more offers and searching with full-text search at the same time.
+     * Otherwise, use _buildSearchQuery().
+     */
+    _buildSearchAggregation(lastOfferId, lastOfferScore, value, showAdminReason, showHidden, filters) {
```
imnotteixeira, line 267: The function name should not be tied to the implementation, but rather to what it does. Something like `_buildSearchContinuationQuery`. Also update the comment above, whoever reading this want to know what the function does, not what it uses (as they can simply look at L268 for that)

src/services/offer.js
```diff
@@ -262,6 +228,59 @@ class OfferService {
             { results };
     }
 
+    /**
+     * Builds a search query with a regular find().
```
imnotteixeira, line 232: Same comment for this documentation. Explain what the function does and when to use, not how it does it.

Maybe it would be valuable to name it something like `_buildInitialSearchQuery` to more clearly distinguish between the one used on the first search, and the subsequent ones

imnotteixeira (changes requested):

src/services/offer.js
BrunoRosendo: Why would the filters change? If they changed, then that has to be a different search. Including the filters in the token feels redundant to me and doesn't really help the FE, since they already saving the filter information and passing it to the endpoint. We'd need to refactor BE and FE.

But what do you think, @TiagooGomess, as for the FE implementation?

imnotteixeira: You are developing an API (which may have multiple clients and you may not control all of them). It does not make sense to continue a search and change the filters, therefore the token should contain all the information to continue a given search. 

This does not affect the FE, it will just send back what you give it. Even if it sends some filters, whenever there's a queryToken they should be ignored (same for search value)

If anything, this could even simplify the FE, but no changes are really required there

BrunoRosendo: @imnotteixeira I implemented this, please review

src/services/offer.js
```diff
@@ -192,29 +196,108 @@ class OfferService {
 
     /**
      * Fetches offers according to specified options
+     * Learn more about keyset search here: https://github.com/NIAEFEUP/nijobs-be/issues/129
+     *
      * @param {*} options
      * value: Text to use in full-text-search
-     * offset: Point to start looking (and limiting)
+     * queryToken: Id of the last fetched offer
```
imnotteixeira, line 203: Update comment

src/services/offer.js
```diff
@@ -192,29 +196,108 @@ class OfferService {
...
+        const offers = (value ? Offer.find({ "$and": [
+            this._buildFilterQuery(filters),
+            { "$text": { "$search": value } }
+        ] }, { score: { "$meta": "textScore" } }
+
+        ) : Offer.find(this._buildFilterQuery(filters)));
+
+        offers.current();
+        if (!showHidden) offers.withoutHidden();
+        if (!showAdminReason) offers.select("-adminReason");
```
imnotteixeira, line 261: Are these filters being applied?

Shouldn't the .current/.withoutHidden/.select be applied to the returned object? I'm not sure the query is being affected by all those filters (depends on whether the different functions mutate the offers object (query) in-place or not)

src/services/offer.js
```diff
@@ -192,29 +196,108 @@ class OfferService {
...
+            ]);
+        } else {
+            offers = Offer.find({ "$and": [
+                this._buildFilterQuery(filters),
+                { _id: { "$gt": ObjectId(lastOfferId) } }
+            ] });
+
+            offers.current();
+            if (!showHidden) offers.withoutHidden();
+            if (!showAdminReason) offers.select("-adminReason");
```
imnotteixeira, line 295: Same here for using the same offers object every time

e.g. `offers = offers.current()`, etc...

imnotteixeira (changes requested):

src/services/offer.js
BrunoRosendo: They are being applied. The behaviour is also tested.
By the way, do you think those 3 repeated lines are enough to extract them to another function? It seemed overkill but I'm not sure

src/services/offer.js
BrunoRosendo: Same answer as above

src/services/offer.js
BrunoRosendo: @imnotteixeira re-review please (the button still isn't working :/)

TiagooGomess (approved):

PR #226, opened 2022-04-09
One click apply

### LINKED ISSUE #217: One click apply button

BrunoRosendo:
closes #217

### DIFF
--- a/src/models/modelUtils.js
+++ b/src/models/modelUtils.js
@@ -26,8 +26,8 @@ export const validImageURL = (val) => {
 };
 
 export const validApplyURL = (val) => {
-    const httpRegex = /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)?/;
-    const emailRegex = /mailto:(\S+@\S+)/;
+    const httpRegex = /^https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)?$/;
+    const emailRegex = /^mailto:(\S+@\S+)$/;
 
     return httpRegex.test(val) || emailRegex.test(val);
 };
--- a/src/models/modelUtils.js
+++ b/src/models/modelUtils.js
@@ -24,3 +24,10 @@ export const validImageURL = (val) => {
 
     return regex.test(val);
 };
+
+export const validApplyURL = (val) => {
+    const httpRegex = /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)?/;
+    const emailRegex = /mailto:(\S+@\S+)/;
+
+    return httpRegex.test(val) || emailRegex.test(val);
+};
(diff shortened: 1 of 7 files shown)


test/end-to-end/offer.js
```diff
@@ -866,6 +866,91 @@ describe("Offer endpoint tests", () => {
...
+                const res = await request()
+                    .post("/offers/new")
+                    .send(withGodToken(offer_params))
+                    .expect(HTTPStatus.UNPROCESSABLE_ENTITY);
+
+                expect(res.body.errors[0]).toHaveProperty("param", "applyURL");
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.BAD_APPLY_URL);
+            });
+
+            test("should fail if applyURL is a URL with an unsupported protocol", async () => {
```
DoStini, line 893: Add a test where you use a `javascript:` link

DoStini, line 893: Add a test where you use a `javascript:` link

src/models/modelUtils.js
```diff
@@ -24,3 +24,10 @@ export const validImageURL = (val) => {
 
     return regex.test(val);
 };
+
+export const validApplyURL = (val) => {
+    const httpRegex = /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)?/;
+    const emailRegex = /mailto:(\S+@\S+)/;
+
+    return httpRegex.test(val) || emailRegex.test(val);
```
DoStini, line 32: This is not enough as it seems.

image

You can use something like this and bypass this security check:

image

Add another additional test that checks this behaviour

DoStini (changes requested):

src/models/modelUtils.js
miguelpduarte: I think it's missing the start of line basically (`^https`). All it should check is to start with "http", "https" or "mailto", and I believe that the browser/React should be able to handle the rest (for example `"` escaping should not be possible even if the regex does not filter this out, I think).

miguelpduarte: For example updating the regex to `httpRegex = /^https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)?$/;` already adresses this issue.

miguelpduarte: For example, I was doing some testing and I think that something as simple as `httpRegex = /^https?:\/\/[a-zA-Z0-9#&=/.]{1,343}$/;` will probably work for our use-case. Given that the aim is simply to filter out potential XSS attempts, just focus on that specifically, the rest is not relevant in this case.

BrunoRosendo: I agree, that makes sense. We'll probably also have a confirmation in the FE letting the user know where he's getting redirected to

DoStini: Yes, the frontend will have a prompt informing the user that he will be redirected

BrunoRosendo: @miguelpduarte Now that I'm testing that regex, shouldn't we be only more inclusive? That one excludes query params like `?a=1&b=2` and other valid characters that were present in the first expression. I'll try to simplify the regex without making more strict

BrunoRosendo: I've got one more question. Should we allow the companies to submit URLs without the http/https? e.g. 'www.google.com`

DoStini (approved):

DoStini (approved):

Issue #222, opened 2022-04-05
Editing publish date error


DoStini:
When requesting to edit with this publish date

![image](https://user-images.githubusercontent.com/23351316/161853153-9bd77530-80cd-4057-8e89-89d87dd46bf9.png)

This is the result

![image](https://user-images.githubusercontent.com/23351316/161853078-345a8ed7-5544-4c6d-ba8b-9bf9168c0561.png)

The problem is here, as if the offer is actually in the past, it will never be editable. I don't think we need this validator anymore

https://github.com/NIAEFEUP/nijobs-be/blob/bbebccb87000553c392a96814661891165163ce6/src/api/middleware/validators/offer.js#L429-L438

This can also be removed

https://github.com/NIAEFEUP/nijobs-be/blob/bbebccb87000553c392a96814661891165163ce6/src/api/middleware/validators/validationReasons.js#L37

PR #229, opened 2022-04-09
Edit company details

dfneves19:


BrunoRosendo:
The solution looks good :)

However, please use the `closes` keyword in the description to indicate which issue this is referring to (in this case, `closes #222`. Also, update the title to better explain what your solution does, "Edit company details" seems like you're implementing that feature, which isn't the case. You could use something like "Removed isEditable validator".

Finally, watch out for your commit messages. You have a commit named "branch" and, in the other one, you should explain what you did instead of saying something abstract like "fixed X" (like in the title)
You can fix these commits using the interactive rebase. You can ask for help if you need it

DoStini:
@BrunoRosendo I think this was already merged in a previous commit, the file diffs are from a previous pr
I suggest @dfneves19 and @kiko-serra to go to develop, pull and then create a new branch called `feature/edit-company-details` for example

BrunoRosendo:
> @BrunoRosendo I think this was already merged in a previous commit, the file diffs are from a previous pr
> I suggest @dfneves19 and @kiko-serra to go to develop, pull and then create a new branch called `feature/edit-company-details` for example

Oh ok, just close this PR then. And take note of the closes keyword @dfneves19, as it's very useful

PR #230, opened 2022-04-09
NIJobs Launch

DoStini:


.dockerignore
```diff
@@ -0,0 +1,14 @@
+# Items that don't need to be in a Docker image.
+# Anything not used by the build system should go here.
+Dockerfile
+docker-compose*.yml
+.dockerignore
+.gitignore
+README.md
+# Ignore .git and .cache folders
+.git
```
monkin77, line 9: Está feio

BrunoRosendo (approved):
After carefully checking every change, I'm pretty sure this is good enough for the first version of NIJobs ;)

monkin77 (approved):

PR #239, opened 2022-05-29
Upgrade multer version

DoStini:


BrunoRosendo (approved):
We should probably merge to master too, since the build is failing right now

PR #242, opened 2022-07-26
Simplified http Regex

### LINKED ISSUE #241: Replace applyURL's http regex with a simpler version
Right now, [this regex expression](https://github.com/NIAEFEUP/nijobs-be/blob/6992a1a8cd7ec8ba249d97245054ed81645a24f4/src/models/modelUtils.js#L29) is unreadable and too complicated. The idea was to only allow valid URLs, but that's overkill in terms of security.

Instead, we can replace it with a simple RE that checks if the URL begins with http or https.

CarlosMealha:
Hey guys, I think it's everything okay with this regex, but if you have any suggestion feel free to give it

Closes #241

### DIFF
--- a/src/models/modelUtils.js
+++ b/src/models/modelUtils.js
@@ -26,7 +26,7 @@ export const validImageURL = (val) => {
 };
 
 export const validApplyURL = (val) => {
-    const httpRegex = /^https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)?$/;
+    const httpRegex = /^https?:\S+$/;
     const emailRegex = /^mailto:(\S+@\S+)$/;
 
     return httpRegex.test(val) || emailRegex.test(val);
--- a/src/models/modelUtils.js
+++ b/src/models/modelUtils.js
@@ -26,7 +26,7 @@ export const validImageURL = (val) => {
 };
 
 export const validApplyURL = (val) => {
-    const httpRegex = /^https?:\S+$/;
+    const httpRegex = /^https?:\/\/\S+\.\S+$/;
     const emailRegex = /^mailto:(\S+@\S+)$/;
 
     return httpRegex.test(val) || emailRegex.test(val);
--- a/test/end-to-end/offer.js
+++ b/test/end-to-end/offer.js
@@ -980,6 +980,22 @@ describe("Offer endpoint tests", () => {
 
                 expect(res.body).toHaveProperty("applyURL", applyURL);
             });
+
+            test("should fail if applyURL is an invalid HTTPS URL", async () => {
+                const applyURL = "https://invalid";
+                const offer_params = generateTestOffer({
+                    applyURL,
+                    owner: test_company._id,
+                });
+
+                const res = await request()
+                    .post("/offers/new")
+                    .send(withGodToken(offer_params))
+                    .expect(HTTPStatus.UNPROCESSABLE_ENTITY);
+
+                expect(res.body.errors[0]).toHaveProperty("param", "applyURL");
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.BAD_APPLY_URL);
+            });
         });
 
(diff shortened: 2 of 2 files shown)


BrunoRosendo:
@CarlosMealha Did you test the behaviour with React like it was asked in the issue? If so, could you elaborate on how that went?

DoStini:
rebase

BrunoRosendo:
https://github.com/NIAEFEUP/nijobs-fe/pull/249#discussion_r939547747

From this discussion, I think we should add a bit more complexity and verify if there's a dot as well. So, something like http(s)://[CHARS].[CHARS]

CarlosMealha:
> [NIAEFEUP/nijobs-fe#249 (comment)](https://github.com/NIAEFEUP/nijobs-fe/pull/249#discussion_r939547747)
> 
> From this discussion, I think we should add a bit more complexity and verify if there's a dot as well. So, something like http(s)://[CHARS].[CHARS]

Okay, I will change it and give you my feedback

CarlosMealha:
So I tested this regex with different props and it looks everything okay, the page didn't changed.

Naapperas:
Once this and https://github.com/NIAEFEUP/nijobs-fe/pull/249 are merged, we can move this regex to [nijobs-commos](https://github.com/NIAEFEUP/nijobs-commons)

Naapperas:
@CarlosMealha update the regex so we can merge this PR

CarlosMealha:
> @CarlosMealha update the regex so we can merge this PR

I already updated the regex, do you want me to do anything else?

BrunoRosendo:
> [NIAEFEUP/nijobs-fe#249 (comment)](https://github.com/NIAEFEUP/nijobs-fe/pull/249#discussion_r939547747)
> 
> From this discussion, I think we should add a bit more complexity and verify if there's a dot as well. So, something like http(s)://[CHARS].[CHARS]

You're missing the change we talked about in this comment. Adding .S+ probably does the trick :)
We should also add a test case for that: the success case already works so add a test where a URL like https://invalid should fail

Naapperas:
On another note, once #245 gets merged, rebase so the audit run does not fail, then we can merge this

CarlosMealha:
> On another note, once #245 gets merged, rebase so the audit run does not fail, then we can merge this

Okay guys, I'll do it

CarlosMealha:
I changed the regex and I'm not 100% sure if this is what you expected:
image

I tested the expression online and it works, but my test is not passing, and I don't know why
image
image

Naapperas:
> I changed the regex and I'm not 100% sure if this is what you expected: image
> 
> Also tested the expression online and it works, but it's not passing, and I don't know why image image

That's almost it: you are just missing the `//` part of the URL. Other than that, I don't think URLs with multiple successive `.` are valid, you just need to remove the `+` in front of the `\.` and that should be it.

BrunoRosendo:
I didn't test it but the offer_params property should be applyURL instead of apply

src/models/modelUtils.js
```diff
@@ -26,7 +26,7 @@ export const validImageURL = (val) => {
 };
 
 export const validApplyURL = (val) => {
-    const httpRegex = /^https?:\/\/(www\.)?[-a-zA-Z0-9@:%._+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_+.~#?&//=]*)?$/;
+    const httpRegex = /(^https?:\/\/)\S+\.\S+$/;
```
Naapperas, line 29:
```suggestion
    const httpRegex = /^https?:\/\/\S+\.\S+$/;
```
I don't see how a group is needed here but I might be missing something, if that's the case just mark this as resolved.

Naapperas (changes requested):

src/models/modelUtils.js
CarlosMealha: I just grouped to organize it

BrunoRosendo: @CarlosMealha Just remove those brackets and then we can merge this

BrunoRosendo (approved):

Naapperas (approved):

PR #244, opened 2022-08-16
Allow requests with cors specified in a list of regexes or urls and setup heroku

### LINKED ISSUE #145: Add CORS config help to README
Add some explanation to what should be placed in the `ACCESS_CONTROL_ALLOW_ORIGIN` variable

DoStini:
We are trying to give a more interesting experience for frontend previews.
In heroku deploys (which we will be using also for backend previews), we will allow any request from one of our frontend previews

closes #145


DoStini:
@imnotteixeira @miguelpduarte Do you see any problems with this approach? Do you think there is a different way of doing dynamic cors?

Cors with "*" didn't work due to the use of `credentials="include"` in frontend requests

I've also checked and netlify doesn't allow the creation of  links with this regex

image

src/config/env.js
```diff
@@ -30,6 +30,10 @@ export default Object.freeze({
     admin_password: process.env.ADMIN_PASSWORD,
     access_control_allow_origin: process.env.ACCESS_CONTROL_ALLOW_ORIGIN || "*",
 
+    // Deploy previews
+    is_heroku: process.env.IS_HEROKU,
```
imnotteixeira, line 34: This variable is not needed, simply have an allowlist of allowed domains to query the api

imnotteixeira (changes requested):

src/config/env.js
imnotteixeira: Also, the name would be misleading, this should not be coupled with any deployment service. What if we use vercel, netlify, digital ocean/ something else?

imnotteixeira:
> @imnotteixeira @miguelpduarte Do you see any problems with this approach? Do you think there is a different way of doing dynamic cors?
> 
> Cors with "*" didn't work due to the use of `credentials="include"` in frontend requests
> 
> I've also checked and netlify doesn't allow the creation of links with this regex
> 
> image

As you've noticed, allowing `*` does not work due to `credentials=include`, as well as not being a good practice.

The way to go is what you've done here: have an allowlist of domains, and only return the header if it matches the allowlisted domain

src/config/env.js
DoStini: All right then

DoStini: I believe this is a good implementation to what we want.

We will be using this regex in beta deploys (or heroku): `"https:\/\/deploy-preview-\d+--nijobs\.netlify\.app"`

imnotteixeira: I don't think a regex is the way to go here. It solves the very specific problem we have now, sure, but what if there's a new provider we may want to use as well as netlify, what about local frontend builds? No one should have to know/understand this regex in order to allowlist something.

I think the ideal solution would be to make this variable a JSON string representing an array of regexes. It allows the solution you have now, but people can still add more domains easily, without having to think about regexes (exact strings are still valid regexes anyway)

In the end, the config object would parse the JSON into a JS array of Regexes, which can be evaluated against the requester's origin

DoStini: I think this might be the way to go. I left some examples in the readme and in the env file

README.md
```diff
@@ -40,6 +39,14 @@ To start developing, copy `.env` to `.env.local` (by running `cp .env .env.local
...
 Then, you can override the variable's values, according to their explanation in `.env`.
 
+#### CORS Setup
+
+In order to allow requests from multiple origins besides the one defined by `ACCESS_CONTROL_ALLOW_ORIGIN`, an array of URL's or regexes can be defined by `ACCESS_CONTROL_ALLOW_ORIGIN_REGEX_LIST`. It is crucial that all URL's have no trailing `/`. Example:
+
+```
+ACCESS_CONTROL_ALLOW_ORIGIN=https://localhost:3000
+ACCESS_CONTROL_ALLOW_ORIGIN_REGEX_LIST=["https:\\/\\/deploy-preview-\\d+--nijobs\\.netlify\\.app", "https://nijobs.netlify.app"]
+```
```
Naapperas, line 49: Couldn't this be used to close #145 ?

DoStini: Yup

.env
```diff
@@ -32,8 +32,12 @@ TEST_LOG_REQUESTS=false
 # Default admin credentials - ALSO OVERRIDE IN PRODUCTION
 ADMIN_EMAIL=ni@aefeup.pt
 ADMIN_PASSWORD=n1j0bs_ftw.12345
+
 #CORS allowed origin - OVERRIDE IN PRODUCTION
 ACCESS_CONTROL_ALLOW_ORIGIN=
```
imnotteixeira, line 37: Remove this var and just use the new one (remember to edit the config/env.js)

.env
```diff
@@ -32,8 +32,12 @@ TEST_LOG_REQUESTS=false
 # Default admin credentials - ALSO OVERRIDE IN PRODUCTION
 ADMIN_EMAIL=ni@aefeup.pt
 ADMIN_PASSWORD=n1j0bs_ftw.12345
+
 #CORS allowed origin - OVERRIDE IN PRODUCTION
 ACCESS_CONTROL_ALLOW_ORIGIN= 
+# List of regexes or url's specifying allowed origins. Example:
+# ACCESS_CONTROL_ALLOW_ORIGIN_REGEX_LIST=["https:\\/\\/deploy-preview-\\d+--nijobs\\.netlify\\.app", "https://nijobs.netlify.app"]
+ACCESS_CONTROL_ALLOW_ORIGIN_REGEX_LIST=
```
imnotteixeira, line 40:
```suggestion
ACCESS_CONTROL_ALLOW_ORIGINS=
```

README.md
```diff
@@ -40,6 +39,14 @@ To start developing, copy `.env` to `.env.local` (by running `cp .env .env.local
 
 Then, you can override the variable's values, according to their explanation in `.env`.
 
+#### CORS Setup
+
+In order to allow requests from multiple origins besides the one defined by `ACCESS_CONTROL_ALLOW_ORIGIN`, an array of URL's or regexes can be defined by `ACCESS_CONTROL_ALLOW_ORIGIN_REGEX_LIST`. It is crucial that all URL's have no trailing `/`. Example:
+
+```
+ACCESS_CONTROL_ALLOW_ORIGIN=https://localhost:3000
```
imnotteixeira, line 47: No need for the first one

src/loaders/express.js
```diff
@@ -76,9 +76,23 @@ export default (app) => {
...
-        // Allow connections for all origins
+    app.use((req, res, next) => {
+        // Allow connections for default specified connections
         res.setHeader("Access-Control-Allow-Origin", config.access_control_allow_origin);
+
+        // Allow requests from connections specified by the allow list regexes
+        const origin = req.header("origin")?.toLowerCase();
+        if (origin) {
+            const originAllowed = config.access_control_allow_origin_regex_list.find((allowOrigin) => {
+                const matches = origin.match(RegExp(allowOrigin, "g"));
```
imnotteixeira, line 87: There's probably a built-in way to just check if a particular string matches a regex, no need to get the matches, maybe .test or .matches or something, can't remember.

imnotteixeira (changes requested):

.env
DoStini: @BrunoRosendo We'll have to update .env in our server

BrunoRosendo:
@DoStini Can this be implemented with another service? It'd be very useful to have previews for the backend

.env
imnotteixeira: Why do we need both settings? We can just use `ACCESS_CONTROL_ALLOW_ORIGIN_REGEX_LIST` now.

documentation/docs/intro/getting-started.md
```diff
@@ -34,6 +34,16 @@ To start developing, copy `.env` to `.env.local` (by running `cp .env .env.local
...
 
+Whenever the file is updated, the server needs to be restarted for changes to have effect.
+
+
+### Cors Setup
+
+In order to allow requests from multiple clients, an array of URL's or regexes can be defined by `ACCESS_CONTROL_ALLOW_ORIGINS`. This is an important step regarding connecting the frontend and the backend of the project. It is crucial that all URL's have no trailing `/`. Example:
+
+```
+ACCESS_CONTROL_ALLOW_ORIGINS=["https:\/\/localhost:3000", "https:\\/\\/deploy-preview-\\d+--nijobs\\.netlify\\.app", "https://nijobs.netlify.app"]
```
imnotteixeira, line 45: Maybe you'll want to rename to `ACCESS_CONTROL_ALLOW_ORIGIN_REGEX_LIST`

src/loaders/express.js
```diff
@@ -49,7 +49,7 @@ export default (app) => {
 
     // Set the API call rate limit only on production
     if (process.env.NODE_ENV === "production") {
-        const api_rate_limiter = new RateLimit({
+        const api_rate_limiter = RateLimit({
```
imnotteixeira, line 52: Why this change?

src/loaders/express.js
```diff
@@ -76,9 +76,23 @@ export default (app) => {
     }
 
     // Adding headers (CORS)
-    app.use((_, res, next) => {
-        // Allow connections for all origins
-        res.setHeader("Access-Control-Allow-Origin", config.access_control_allow_origin);
+    app.use((req, res, next) => {
+        // Allow connections for default specified connections
```
imnotteixeira: Comment seems wrong

src/loaders/express.js
```diff
@@ -76,9 +76,23 @@ export default (app) => {
...
+        res.setHeader("Access-Control-Allow-Origin", "null");
+
+        // Allow requests from connections specified by the allow list regexes
+        const origin = req.header("origin")?.toLowerCase();
```
imnotteixeira: Why convert to lower case? Might cause some misunderstandings with the regex provided in the config file

src/loaders/express.js
```diff
@@ -91,7 +105,7 @@ export default (app) => {
     });
 
     // Health check endpoint
-    app.get("/", (_, res) => res.status(HTTPStatus.OK).json({ "online": true }));
+    app.get("/", (_, res) => res.status(HTTPStatus.OK).send("OK"));
```
imnotteixeira: Why?

imnotteixeira (changes requested):

.env
DoStini: We don't, forgot to remove it from there

src/loaders/express.js
DoStini: It isn't working correctly and crashed the app in production mode. There was probably an update to this library

documentation/docs/intro/getting-started.md
DoStini: You previously requested to change to `origins`

imnotteixeira: Sure, I meant that it needs to match whatever is on .env

src/loaders/express.js
DoStini: I think Render required the health check endpoint to return this, I will try again with the previous version

documentation/docs/intro/getting-started.md
DoStini: It is the same now

imnotteixeira (approved):

Issue #169, opened 2021-08-09
"Normalize" tests


Naapperas:
Right new there are tests that manage (create/delete) the resources they need and there are others that use what is available from prior tests. This can cause conflicts that should not happen, since the tests, or test groups (mainly tests inside a `describe`) should be independent from each other, and not have an influence on tests that appear after.

So the test suites should be refactored in order to have each test/test group use only the resources they need, creating them and deleting them, without leaving "residues" for other tests.

Naapperas:
I had problems regarding this when working on #168: in the edit offer tests, I had to change the date of creation of the offers because the dates previously used were causing conflicts with another test later on

BrunoRosendo:
We need to prioritize this issue very soon.
We should also split the test files, maybe by route. The test files are way too big right now. Small discussion [here](https://github.com/NIAEFEUP/nijobs-be/pull/238#discussion_r923813377)

Process-ing:
While working on issue #261, I faced some `offer` publish dates in the offer and company tests were outdated (see more details in the [issue](https://github.com/NIAEFEUP/nijobs-be/issues/261)). It's most likely that other dates are also outdated, leading to unwanted behavior in the tests, so those should be checked.

BrunoRosendo:
We can consider separating tests into different files and also have some listener that purges all the database's information. This could happen in each test or group

PR #245, opened 2022-08-17
Bumped mongoose to v6.5.2

BrunoRosendo:


Naapperas:
Some tests will need to be [rewritten](https://mongoosejs.com/docs/migrating_to_6.html#model-exists-returns-a-lean-document-instead-of-boolean)

BrunoRosendo:
> 

That's fine, I've fixed those. I'm going a bit crazy with some other tests that are mysteriously failing

Naapperas:
> > 
> 
> That's fine, I've fixed those. I'm going a bit crazy with some other tests that are mysteriously failing

Push and I can also try and look into it

Naapperas:
Remove the `.idea` and `documentation` folders

src/services/offer.js
```diff
@@ -152,14 +142,9 @@ class OfferService {
             query,
             {
                 isHidden: false,
-                $unset: { hiddenReason: undefined, adminReason: undefined }, // Removing property from document.
+                $unset: { hiddenReason: "", adminReason: "" }, // Removing property from document.
```
Naapperas, line 145: Shouldn't these be set to null instead of an empty string?

Naapperas: Since the purpose is to have them removed from the document itself, and not just "nullify" their value

BrunoRosendo: https://www.mongodb.com/docs/manual/reference/operator/update/unset/#mongodb-update-up.-unset

The specified value doesn't impact the operation, "" is used in the examples. I'm just not sure as to why `undefined` stopped working, I'm yet to find the change on that

Naapperas: https://mongoosejs.com/docs/migrating_to_6.html#removed-omitundefined

BrunoRosendo: Doesn't seem like it. That's $set instead of $unset, and the values weren't being removed, they were keeping their old values for some reason

Anyway, now I'm trying to fix the remaining 10 tests but have yet to find the cause, something to do with dependencies with previous tests... We really need #169

Naapperas: Gonna bump the priority on that one then

Naapperas: Also, just to clarify on the subject, unless I misunderstood something, the values were keeping their old value because when using `undefined` those keys were being removed from the update payload and this not update anything in actuality.

Not that it matters much if the current version works as intended unless there are some new "best practices" regarding this matter.

BrunoRosendo: Yes I think you're completely right!

Naapperas:
Fix the tests only as much as needed, any further refactoring on the tests should come only when #169 is dealt with.

BrunoRosendo:
For the record, this afterAll() seems to be deleting the wrong offer, used in tests below. I'll continue investigating tomorrow
image

BrunoRosendo:
> For the record, this afterAll() seems to be deleting the wrong offer, used in tests below. I'll continue investigating tomorrow image

This query was wrong. Before, it didn't delete anything but now it treats it as an empty object and deletes every offer

BrunoRosendo:
@Naapperas Everything should be done now :relieved:

Naapperas (dismissed):

Naapperas (approved):

PR #251, opened 2022-09-13
Fix/hide hidden offers

### LINKED ISSUE #250: Any user can see a hidden offer
Currently, we're setting a `req.targetOwner` in each request related to offers or companies. We do it this way, as can be seen [here](https://github.com/NIAEFEUP/nijobs-be/blob/6da1c15376767e493f9fe38358c8fbe436993de6/src/api/middleware/offer.js#L21):
- If the user is logged in as a company, use their id
- Otherwise, check if the request has the `owner` parameter and use that

The problem is that we set the `targetOwner` even if the user is not logged in and doesn't have any special permissions (admin or god). Because of this, if we try to access a hidden offer in the `offer/:offerId` endpoint, a regular user can see it as long as they know the company's ID (which is easy to discover with the company endpoints), like this:
```
GET /offers/:offerId

{
  "owner": ":ownerId"
}
```

I see two ways we can fix this:
- Change the `setTargetOwner` to only use the `owner` parameter when the user has admin privileges (including god). We need to make sure this doesn't break other functionalities using this function. I still think this is the preferred way.
- Instead of passing `targetOwner` to the offer service, we could pass `req.user?.company?._id`, to make sure we only allow the user if they have admin privileges or if they're logged in as the owner. However, I don't really like this approach because we should use `targetOwner` whenever possible, it should have the same permissions across the whole website

limwa:
Closes #250.

These changes introduce a check to make it so that the target owner in any request can only be set if the user has admin privileges. As such, regular users will be unable to see hidden offers.

### DIFF
--- a/src/api/middleware/offer.js
+++ b/src/api/middleware/offer.js
@@ -19,7 +19,8 @@ export const isOwnerNotBlocked = async (req, res, next) => {
 };
 
 export const setTargetOwner = (req, res, next) => {
+    const wantedTargetOwner = req.hasAdminPrivileges && req.body.owner;
+    req.targetOwner = req.user?.company?._id.toString() || wantedTargetOwner || undefined;
 
-    req.targetOwner = req.user?.company?._id.toString() || req.body.owner;
     return next();
 };
--- a/src/api/middleware/offer.js
+++ b/src/api/middleware/offer.js
@@ -19,7 +19,7 @@ export const isOwnerNotBlocked = async (req, res, next) => {
 };
 
 export const setTargetOwner = (req, res, next) => {
-    const wantedTargetOwner = req.hasAdminPrivileges && req.body.owner;
+    const adminTargetOwner = req.hasAdminPrivileges && req.body.owner;
     req.targetOwner = req.user?.company?._id.toString() || wantedTargetOwner || undefined;
 
     return next();
--- a/src/api/middleware/offer.js
+++ b/src/api/middleware/offer.js
@@ -20,7 +20,7 @@ export const isOwnerNotBlocked = async (req, res, next) => {
 
 export const setTargetOwner = (req, res, next) => {
     const adminTargetOwner = req.hasAdminPrivileges && req.body.owner;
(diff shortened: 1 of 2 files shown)


Naapperas (dismissed):

limwa:
Ready to merge!

src/api/middleware/offer.js
```diff
@@ -19,7 +19,8 @@ export const isOwnerNotBlocked = async (req, res, next) => {
 };
 
 export const setTargetOwner = (req, res, next) => {
+    const wantedTargetOwner = req.hasAdminPrivileges && req.body.owner;
+    req.targetOwner = req.user?.company?._id.toString() || wantedTargetOwner || undefined;
```
BrunoRosendo, line 23: Why is the `|| undefined` necessary? Is this different from what we had before?

src/api/middleware/offer.js
```diff
@@ -19,7 +19,8 @@ export const isOwnerNotBlocked = async (req, res, next) => {
 };
 
 export const setTargetOwner = (req, res, next) => {
+    const wantedTargetOwner = req.hasAdminPrivileges && req.body.owner;
```
BrunoRosendo, line 22:
```suggestion
    const adminTargetOwner = req.hasAdminPrivileges && req.body.owner;
```

BrunoRosendo (changes requested):

src/api/middleware/offer.js
limwa: Yes. Please note that before the condition was `req.user?.company?._id.toString() || req.body.owner`. Therefore, since the company ID and the req.body.owner can only be either strings or undefined, the resulting value is a string or undefined.

With the new fix, if the company ID is undefined and the user doesn't have admin privileges, the resulting value would be `false` instead of `undefined`.

The `|| undefined` is there to make sure that the behavior, in the case of non-authorization, stays exactly the same.

BrunoRosendo (approved):

PR #262, opened 2022-10-17
Allowing for null values when editing optional offer fields

### LINKED ISSUE #260: Unable to remove optional offer fields
This discussion started in https://github.com/NIAEFEUP/nijobs-fe/issues/250 and we discovered the backend also needs some changing regarding this. Basically, some of the offer optional fields should be removable (e.g. `applyURL` and `vacancies`). The companies might want to remove them from their offers and.

Currently, setting these values as `null` in the `/offers/edit` endpoint returns the following error:
image

This type of operation should be supported for the optional fields that make sense (without default values for example)

BrunoRosendo:
closes #260

### DIFF
--- a/test/end-to-end/offer.js
+++ b/test/end-to-end/offer.js
@@ -2754,7 +2754,7 @@ describe("Offer endpoint tests", () => {
                 });
             });
 
-            describe("testing dates editing", () => {
+            describe("testing publish dates editing", () => {
                 test("should fail if publishDate after offer publishEndDate", async () => {
                     const res = await test_agent
                         .post(`/offers/edit/${future_test_offer._id.toString()}`)
@@ -2883,6 +2883,15 @@ describe("Offer endpoint tests", () => {
                         .expect(HTTPStatus.OK);
                     expect(res.body).toHaveProperty("applyURL", applyURL);
                 });
+
+                test("Should edit if applyURL is specified as null", async () => {
+                    const applyURL = null;
+                    const res = await test_agent
+                        .post(`/offers/edit/${future_test_offer._id.toString()}`)
+                        .send(withGodToken({ applyURL }))
+                        .expect(HTTPStatus.OK);
+                    expect(res.body).toHaveProperty("applyURL", applyURL);
+                });
             });
 
             describe("testing other validations", () => {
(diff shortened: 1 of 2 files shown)


BrunoRosendo:
This PR does not include jobMinDuration and jobMaxDuration. These fields are weird to me, they're optional on the BE but mandatory on the FE form. Besides from that, the validations gets weird when editing because those fields and mandatory when the other is present

Naapperas:
> This PR does not include jobMinDuration and jobMaxDuration. These fields are weird to me, they're optional on the BE but mandatory on the FE form. Besides that, the validations get weird when editing because those fields and mandatory when the other is present

After discussing this with @DoStini we concluded it is better to have both fields mandatory on the BE: it simplifies things in the long run. For the case when the `jobMaxDuration` equals 12, I think we should interpret 12 as being "12 or more months" to reflect the UI that is shown to the user on the FE.

test/end-to-end/offer.js
```diff
@@ -2883,6 +2883,15 @@ describe("Offer endpoint tests", () => {
...
                     expect(res.body).toHaveProperty("applyURL", applyURL);
                 });
+
+                test("Should edit if applyURL is specified as null", async () => {
+                    const applyURL = null;
+                    const res = await test_agent
+                        .post(`/offers/edit/${future_test_offer._id.toString()}`)
+                        .send(withGodToken({ applyURL }))
+                        .expect(HTTPStatus.OK);
+                    expect(res.body).toHaveProperty("applyURL", applyURL);
```
Naapperas, line 2893: Just out of curiosity: why is this test different from the others?

Naapperas (approved):

AvilaAndre (approved):
I have little to no experience in Express but it looks good to me.

test/end-to-end/offer.js
BrunoRosendo: The `applyURL` tests had that structure so I just followed it :shrug: Nothing that really matters

BrunoRosendo:
> > This PR does not include jobMinDuration and jobMaxDuration. These fields are weird to me, they're optional on the BE but mandatory on the FE form. Besides that, the validations get weird when editing because those fields and mandatory when the other is present
> 
> After discussing this with @DoStini we concluded it is better to have both fields mandatory on the BE: it simplifies things in the long run. For the case when the `jobMaxDuration` equals 12, I think we should interpret 12 as being "12 or more months" to reflect the UI that is shown to the user on the FE.

Sounds good, I opened https://github.com/NIAEFEUP/nijobs-be/issues/263
Merging this now

Issue #246, opened 2022-08-17
Uniformize services to either return a Promise or a Query


BrunoRosendo:
Currently, some of the services return a Promise while others return a Query. We should make this consistent across all the services. It probably makes more sense to return a Promise in all of them, to make sure we don't depend on the service's implementation.

Here are some examples:
[Service returning a Promise](https://github.com/NIAEFEUP/nijobs-be/blob/73c899fd61b0e0a6dd512e5be134106554573f0c/src/services/offer.js#L22), notice the use of async/await
[Service returning a Query](https://github.com/NIAEFEUP/nijobs-be/blob/73c899fd61b0e0a6dd512e5be134106554573f0c/src/services/company.js#L68), notice that it just returns the query

Naapperas:
This was not an issue since Mongoose queries are `thenable` (see [this](https://masteringjs.io/tutorials/fundamentals/thenable) link) and are "awaited" in the route. Nonetheless, we should not mix things up.

PR #270, opened 2022-11-21
Add README badge for code coverage

### LINKED ISSUE #248: Add README badge for code coverage
NIJobs has a pretty good test coverage of all its features but we're not showcasing it to everyone. It'd be nice to have a badge showing how much coverage we have.

One could look into [Codecov](https://about.codecov.io/) or a similar tool. We should be able to easily integrate it with our Github actions. CI. You can look at multiple projects using this, but here's an example: https://github.com/codecov/example-javascript.

It's also possible to integrate this tool with pull requests and it'll tell us how the coverage has increased or decreased. We can consider using this if we think is worth it.

ttoino:
Closes #248.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,6 +31,7 @@ services:
     volumes:
       - ./src:/usr/src/nijobs-be/src
       - ./test:/usr/src/nijobs-be/test
+      - ./coverage:/usr/src/app/coverage
     depends_on:
       - mongo
   mongo:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -29,8 +29,8 @@ services:
     environment:
       - DB_HOST=mongo
     volumes:
-      - ./src:/usr/src/nijobs-be/src
-      - ./test:/usr/src/nijobs-be/test
+      - ./src:/usr/src/app/src
+      - ./test:/usr/src/app/test
       - ./coverage:/usr/src/app/coverage
     depends_on:
       - mongo
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,9 @@ jobs:
     - name: Launch the stack and run the tests
       run: docker-compose up --exit-code-from test test
 
+    - name: Upload code coverage to Codecov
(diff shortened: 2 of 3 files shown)


ttoino:
Code coverage reports can now be found [here](https://app.codecov.io/gh/NIAEFEUP/nijobs-be) (they will only show up for this pull request until it is merged).

Anyone from the organization should be able to login to Codecov with Github and access all the repository settings. I've left them all as the default, but we should discuss if we want any of them changed. [Here](https://docs.codecov.com/docs/common-recipe-list) are some common configurations and [here](https://docs.codecov.com/docs/codecovyml-reference) is the config file reference.

Naapperas (approved):
LGTM, there is a PR that fixes the audit issues, this will be merged after that

docker-compose.yml
```diff
@@ -31,6 +31,7 @@ services:
     volumes:
       - ./src:/usr/src/nijobs-be/src
       - ./test:/usr/src/nijobs-be/test
+      - ./coverage:/usr/src/app/coverage
```
BrunoRosendo, line 34: I'm curious, any reason to use the `app/` folder instead of `nijobs-be` like the other volumes?

BrunoRosendo (commented):
Good job! This is already approved, I just had a small question

docker-compose.yml
ttoino: The [dockerfile](https://github.com/NIAEFEUP/nijobs-be/blob/c272e3bbbda9937b4ec7c2c4d8ac6bbb811684c7/Dockerfile-test#L4) is using the `app` folder as the workdir, not sure why the other volumes use `nijobs-be` :thinking:

BrunoRosendo: You're right :upside_down_face: Seems like those volumes aren't even used... Can you fix this? The `web-dev` service is correct, this is the incorrect one

BrunoRosendo (dismissed):
Nice! We just need to wait for #246 to be merged now

Naapperas (approved):
LGTM

Issue #221, opened 2022-03-29
Make the search system sortable by other fields


BrunoRosendo:
In order to do this, we need to make the current system receive a field in the query and use to sort the results and construct the queryToken.

You can read more about it in [this discussion](https://github.com/NIAEFEUP/nijobs-be/pull/212#discussion_r836838250)

PR #276, opened 2022-12-06
247 default order offers by most recent

### LINKED ISSUE #247: Default order offers by most recent

dsantosferreira:
Closes #247 

Offers are now ordered with the following precedence: score > publish date (by most recent) > id. I ran the test below using Postman to check if it was working as intended (using the query "Julia"):

orderPublishDateTest

### DIFF
--- a/test/end-to-end/offer.js
+++ b/test/end-to-end/offer.js
@@ -1517,10 +1517,8 @@ describe("Offer endpoint tests", () => {
                                     expect(res.body.results[i].publishDate)
                                         .toBe(res.body.results[i + 1].publishDate);
 
-                                    // eslint-disable-next-line no-undef
-                                    expect(BigInt(`0x${res.body.results[i]._id}`))
-                                        // eslint-disable-next-line no-undef
-                                        .toBeLessThan(BigInt(`0x${res.body.results[i + 1]._id}`));
+                                    expect(res.body.results[i]._id < res.body.results[i + 1]._id)
+                                        .toBeTruthy();
                                 }
                             }
                         });
@@ -1567,10 +1565,8 @@ describe("Offer endpoint tests", () => {
 
                             expect(res2.status).toBe(HTTPStatus.OK);
                             expect(res2.body.results).toHaveLength(1);
-                            // eslint-disable-next-line no-undef
-                            expect(BigInt(`0x${res2.body.results[0]._id}`))
-                                // eslint-disable-next-line no-undef
-                                .toBeGreaterThan(BigInt(`0x${res1.body.results[1]._id}`));
+                            expect(res2.body.results[0]._id > res1.body.results[1]._id)
+                                .toBeTruthy();
                         });
                     });
                 });
@@ -2176,10 +2172,8 @@ describe("Offer endpoint tests", () => {
                                         expect(res.body.results[i].publishDate)
                                             .toBe(res.body.results[i + 1].publishDate);
 
-                                        // eslint-disable-next-line no-undef
-                                        expect(BigInt(`0x${res.body.results[i]._id}`))
-                                            // eslint-disable-next-line no-undef
-                                            .toBeLessThan(BigInt(`0x${res.body.results[i + 1]._id}`));
+                                        expect(res.body.results[i]._id < res.body.results[i + 1]._id)
+                                            .toBeTruthy();
                                     }
                                 }
                             }
@@ -2255,10 +2249,8 @@ describe("Offer endpoint tests", () => {
                             expect(res2.status).toBe(HTTPStatus.OK);
                             expect(res2.body.results).toHaveLength(1);
 
-                            // eslint-disable-next-line no-undef
-                            expect(BigInt(`0x${res2.body.results[0]._id}`))
-                                // eslint-disable-next-line no-undef
-                                .toBeGreaterThan(BigInt(`0x${res1.body.results[3]._id}`));
+                            expect(res2.body.results[0]._id > res1.body.results[3]._id)
+                                .toBeTruthy();
                         });
 
                         test("Should succeed if there are no more offers after the last one", async () => {
--- a/test/end-to-end/offer.js
+++ b/test/end-to-end/offer.js
@@ -1069,7 +1069,7 @@ describe("Offer endpoint tests", () => {
 
             describe("queryToken validation", () => {
                 test("should fail if queryToken does not contain a valid id", async () => {
-                    const queryToken = (new OfferService()).encodeQueryToken("123");
+                    const queryToken = (new OfferService()).encodeQueryToken("123", 10, mockCurrentDate, "test");
 
                     const res = await request()
                         .get("/offers")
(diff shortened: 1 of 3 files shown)


BrunoRosendo:
@dsantosferreira Run `npm audit fix` and see if the audit can be fixed automatically with it. I think it can

Naapperas:
> > @dsantosferreira Run `npm audit fix` and see if the audit can be fixed automatically with it. I think it can
> 
> @BrunoRosendo Ran that command locally and it didn't find any vulnerabilities after fixing. However, it now has conflicts with branch "develop". What should i do?

#278 fixes audit issues, maybe we can merge this PR after that one.

BrunoRosendo:
@dsantosferreira You should remove the last commit. Check git rebase -i or git reset for example

It seems the audit is fixed in another PR so you can wait for it to be merged and rebase.

A note for the future, you can test the audit by running npm audit --production before pushing

BrunoRosendo:
@dsantosferreira What's missing in this PR? Removing last commit and rebasing?

BrunoRosendo (changes requested):
I'm pretty sure this change breaks the pagination of the search. First of all, please read [this PR](https://github.com/NIAEFEUP/nijobs-be/pull/212) and [this article](https://medium.com/swlh/mongodb-pagination-fast-consistent-ece2a97070f3) that explains the implementation we use in NIJobs.

We use a key-set based search, meaning we use IDs (or any field we use for sorting, like `score` or now `publishDate`). Why use IDs in the first place? Because they're inherently sorted by creation date. However, our system has a `publishDate` that might differ in order, and this PR will fix that.

That's where the problem occurs. For example, the user fetches 3 offers with id 1, 2 and 3, but they are actually now sorted by [3, 2, 1]. Rightly so, the system will register the last ID as 1 and create the `queryToken`. When the user fetches the next page, the system will retrieve that ID and return offers after ID 1 (instead of 3), meaning that offers 2 and 3 will be back in the second page :fearful: 

Why isn't this tested? Because the pagination tests only use pages with size 1, so order does not matter. PLEASE, create tests for this scenario when you fix it!

Here's a summary of what should be done in order to fix this issue:
- Include the `publishDate` of the last offer in the `queryToken`. Look at `decodeQueryToken` and `encodeQueryToken` (after #221 it will be any sortable field)
- In `_buildSearchContinuationQuery`, change the logic a little bit:
  - When a `value` exists, retrieve offers before publish date if the score is the same. If both are the same, use the current ID system (do not remove this part).
  - When a `value` is not used, retrieve offers before the last publish date. If they're the same, use the ID once again

I think that should fix the issue. Once again, do not forget tests for these scenarios

Naapperas (changes requested):
LGTM, only thing missing is testing this behavior correctly

Naapperas:
@dsantosferreira #278 was merged, if you rebase it should clear most of the audit issues.

BrunoRosendo:
@dsantosferreira There is a test failing and the PR has some commits from previous merges, can you fix that? Tell us if you need help

dsantosferreira:
> @dsantosferreira There is a test failing and the PR has some commits from previous merges, can you fix that? Tell us if you need help

@BrunoRosendo Regarding the commits from previous merges, I have already talked with Nuno and already know how to fix it. Unfortunately i won't be able to do it today but the issue should be fixed tomorrow night. As for the test that is failing I have no idea why that is happening... The test that fails comes from the "creating a new offer" endpoint which has nothing to do with the code I developed. While I was writing my tests I ran the test suites multiple times and noticed that some of the tests fail in an inconsistent way. I ran the test suites thrice just now and got these results:


### First try
test1Results
test1Suite

### Second try
test2Success

### Third try
test3Results
test3Suite

Any idea on how to solve this?

test/end-to-end/offer.js
```diff
@@ -1417,6 +1454,125 @@ describe("Offer endpoint tests", () => {
...
+                                    queryToken: res1.body.queryToken,
+                                    limit: 1
+                                });
+
+                            expect(res2.status).toBe(HTTPStatus.OK);
+                            expect(res2.body.results).toHaveLength(1);
+                            // eslint-disable-next-line no-undef
+                            expect(BigInt(`0x${res2.body.results[0]._id}`))
+                                // eslint-disable-next-line no-undef
+                                .toBeGreaterThan(BigInt(`0x${res1.body.results[1]._id}`));
```
BrunoRosendo, line 1573: I'm a bit confused by these comparisons, why do they need to be converted into BigInts?

BrunoRosendo (dismissed):
looks good, I just have a small question so I'm approving

test/end-to-end/offer.js
dsantosferreira: Before converting to BigInts I tried converting to Numbers. When doing so, there was a loss of precision when doing the conversion. The numerical value of the id capped at approximately 3.0907e+28 (between 2^94 and 2^95), and ids can go up to 2^96-1, since each id is a 12-byte value. Hence the BigInt conversion to have the necessary precision.
However, I've put some more thought into the matter and realized that I could be comparing the strings directly, since digits come before letters in lexicographical order and each id is guaranteed to be 12 bytes long. The snippet would look like this:

```suggestion
                            expect(`0x${res.body.results[i]._id}` < `0x${res.body.results[i + 1]._id}`)
                                            .toBeTruthy();
```

dsantosferreira: Should i submit these changes or keep it as it is?

Naapperas: Doesn't the "greaterThan" function (I don't remember the exact name) work with strings?

dsantosferreira: No, it only works with numeric values (Number and BigInt)

Naapperas: Does the change you suggested work?

dsantosferreira: Yes

BrunoRosendo: I like that snippet a lot more. Also, you can remove 0x and get rid of string interpolation.

dsantosferreira: The snippet's version is implemented in all tests that used the BigInt conversion

BrunoRosendo (dismissed):

Naapperas (dismissed):
LGTM

BrunoRosendo:
@dsantosferreira This just needs a rebase and then we can merge

dsantosferreira:
> @dsantosferreira This just needs a rebase and then we can merge

Done.

Naapperas:
@dsantosferreira this was failing the build tests: probably due to them being "bad-ish". When that happens you can re-run the GH workflow and see if you get them all passing, which allows for the PR to be merged.

Naapperas (approved):

dsantosferreira:
> @dsantosferreira this was failing the build tests: probably due to them being "bad-ish". When that happens you can re-run the GH workflow and see if you get them all passing, which allows for the PR to be merged.

Ok got it! Should I merge this PR?

Naapperas:
> > @dsantosferreira this was failing the build tests: probably due to them being "bad-ish". When that happens you can re-run the GH workflow and see if you get them all passing, which allows for the PR to be merged.
> 
> Ok got it! Should I merge this PR?

Absolutely, it was already approved. Good job :)

PR #288, opened 2022-12-18
Make search sortable by other fields

### LINKED ISSUE #221: Make the search system sortable by other fields

BrunoRosendo:
closes #221

### DIFF
--- a/test/end-to-end/offer.js
+++ b/test/end-to-end/offer.js
@@ -2695,7 +2695,7 @@ describe("Offer endpoint tests", () => {
                         description: "Ability to have an incredible job",
                         jobType: "OTHER",
                         location: "Aveiro",
-                        vacancies: 2,
+                        vacancies: 1,
                         ownerName: "Awesome Company",
                     });
                 });
@@ -2853,8 +2853,8 @@ describe("Offer endpoint tests", () => {
 
                     expect(res.status).toBe(HTTPStatus.OK);
                     expect(res.body?.results).toHaveLength(2);
-                    expect(res.body.results[0].title).toBe(test_offer.title);
-                    expect(res.body.results[1].title).toBe("Amazing offer");
+                    expect(res.body.results[0].title).toBe("Amazing offer");
+                    expect(res.body.results[1].title).toBe(test_offer.title);
                 });
 
                 test("should sort by vacancies descending", async () => {
--- a/test/end-to-end/offer.js
+++ b/test/end-to-end/offer.js
@@ -1026,6 +1026,16 @@ describe("Offer endpoint tests", () => {
                 const FieldValidatorTester = QueryValidatorTester("technologies");
                 FieldValidatorTester.mustHaveValuesInRange(TechnologyConstants.TechnologyTypes, TechnologyConstants.MIN_TECHNOLOGIES + 1);
             });
+
+            describe("sortBy", () => {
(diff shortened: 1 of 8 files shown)


BrunoRosendo:
The code should be done but I'll be waiting for #276 to finish so I can test some more and rebase

Naapperas:
Doesn't this need to be rebased? I see some old commits present in the PR...

BrunoRosendo:
> Doesn't this need to be rebased? I see some old commits present in the PR...

It does, I'll rebase when the other PR is merged. I also want to add some tests after that

dsantosferreira (changes requested):
Looks good to me! In addition to looking at the code I also messed around with a couple of examples in Postman and it seemed to be working well. Only things that seem to be missing are testing situations involving multiple pages (which seems to be what @BrunoRosendo plans on doing) and rebasing. Great job!

src/services/company.js
```diff
@@ -53,51 +53,55 @@ class CompanyService {
...
+     * @param {*} showAdminReason weather to show the admin reason given to hide this company, defaults to false
      */
-    findById(company_id, showHidden = false, showAdminReason = false) {
+    async findById(company_id, showHidden = false, showAdminReason = false) {
         const query = Company.findById(company_id);
         if (!showHidden) query.withoutBlocked().withoutDisabled();
         if (!showAdminReason) query.hideAdminReason();
-        return query;
+        const company = await query;
+        return  company;
```
DoStini: Is this necessary? I think calling `await findById` has the same effect as it was previously

src/services/company.js
```diff
@@ -53,51 +53,55 @@ class CompanyService {
...
-        return Company.findByIdAndUpdate(
+    async block(companyId, adminReason) {
+        const company =  await Company.findByIdAndUpdate(
             companyId,
             {
                 isBlocked: true,
                 adminReason
             },
             { new: true });
+        return company;
```
DoStini: same question for all the changes of this type

BrunoRosendo: This was because of #246, there seems to be a difference between returning a promise or a query

test/end-to-end/offer.js
```diff
@@ -40,6 +40,7 @@ describe("Offer endpoint tests", () => {
         isHidden: false,
         isArchived: false,
         requirements: ["The candidate must be tested", "Fluent in testJS"],
+        vacancies: 2,
```
dsantosferreira, line 43: The number of vacancies of the two offers used in tests below should be different (they both have 2 vacancies at the moment).
 https://github.com/NIAEFEUP/nijobs-be/blob/0977f3cb447acc0aa47b5e0a3b0adc040fc75bf2/test/end-to-end/offer.js#L2849-L2869 
https://github.com/NIAEFEUP/nijobs-be/blob/0977f3cb447acc0aa47b5e0a3b0adc040fc75bf2/test/end-to-end/offer.js#L2690-L2700

dsantosferreira (commented):
Everything looks good! After changing the vacancies value of one of the test offers this should be ready to be merged.

dsantosferreira (approved):
Looks good to me! Great job!

PR #306, opened 2023-02-11
224 confirmation email when companies register

### LINKED ISSUE #224: Confirmation Email when companies are accepted
In order to prevent fake users using official company's emails, we should implement a confirmation email with a token. You can read more about tokens in #184

FranciscoCardoso913:
# Confirmation email when companies register
### Important:

Although this is a back-end issue it also has changes in the Front-End! See the changes made in the Front-End here: https://github.com/NIAEFEUP/nijobs-fe/pull/309

Issue: https://github.com/NIAEFEUP/nijobs-be/issues/224

### Discussed solution:
    
    -When created, applications should have a state UNVERIFIED by default instead of the current PENDING state;

    -After creating an application an email with a link should be sent to the company so they can validate their application, the link should expire in 10 minutes;

    -A company should not be able to create an application with an email associated with an UNVERIFIED application whose link is yet to expire;

    -If the link of an application expires and the company tries to create another application the previous application of that company should be deleted;

    -The link should send the user to a page where the application is validated, when validating the application its state should change to pending and an email should be sent to the company.

### Back-End changes:

    -The application model now has a new attribute called "isVerified" which indicates if the application has already been validated or not by the company, the attribute is set to false when the application is created;

	-There is a new application state, an application is UNVERIFIED when the isVerified is false, and PENDING when isVerified is true and both of the approvedAt and the rejectedAt are null;

 	-Before creating a new application the Back-End now checks if an application with the same email as the new application had been created in less than 10 minutes from the time the request for a new application is sent, if so the Back-End returns an error to the Front-End, otherwise, the Back-End deletes the old application( if it exists) and creates a new one;

	-If the Front-End requests it the Back-End validates the application, returning an error if for some reason the validation is illegal (token has expired, the token is not valid, the application is already validated, etc...) ;

	-Emails have been changed! Now when a new application is created a single email with a link is sent to the company, only after the application is validated that two other emails are sent, one of them sent to the NIJobs team informing that a company has created a new application and the other sent to the company to inform that everything went well and that we will review their request.

	-The function verifyAndDecodeToken now returns a different error when the token expires and when the token is invalid, unlike before when the error returned was the same;

### Emails new order:

	-The first email which is sent to the company when the application is created:
Screenshot from 2023-02-11 23-06-16

	-The second email sent to the NiJobs team when the application is validated.
Screenshot from 2023-02-11 23-07-30

	
	-The third email which is sent to the company after the application is validated.
Screenshot from 2023-02-11 23-08-37

### What is yet to be done:
	
	-Tests;

### Doubts:

	-Should we change any of the email's text/order?

	-Did we use the API errors correctly or should we have done something differently?
	
	-Is everything in the correct place or is there anything that is more adequate somewhere else?

    -Should the UNVERIFIED applications, which can no longer be verified, be deleted automatically from time to time or should we let it be like it is now (it's only deleted when a new application with the same email is created  )?


.env
```diff
@@ -20,6 +20,9 @@ JWT_SECRET=Lendas contam que o Rui foi membro do IEEE.
 # Frontend Password Recovery Base Route
 PASSWORD_RECOVERY_LINK=https://localhost:3000/recover
 
+# Froantend Application Confirmation Base Route
```
DoStini:
```suggestion
# Frontend Application Confirmation Base Route
```

.env
```diff
@@ -20,6 +20,9 @@ JWT_SECRET=Lendas contam que o Rui foi membro do IEEE.
...
+APPLICATION_CONFIRMATION_LINK=https://localhost:8087/apply/company/validate/
```
DoStini: Should be port 3000

src/lib/token.js
```diff
@@ -1,15 +1,22 @@
...
     try {
         return jwt.verify(token, secret, { algorithm: "HS256" });
-    } catch (err) {
-        return null;
+    } catch (jwtErr) {
+        if (jwtErr.name === "TokenExpiredError") {
+            return next(new APIError(HTTPStatus.FORBIDDEN, ErrorTypes.FORBIDDEN, ValidationReasons.EXPIRED_TOKEN));
+        } else {
+            return next(new APIError(HTTPStatus.FORBIDDEN, ErrorTypes.FORBIDDEN, ValidationReasons.INVALID_TOKEN));
+        }
```
DoStini: I don't think it is necessary to differenciate between expired and invalid. I'd just stick with invalid. Also, needs to be updated from master @Naapperas

src/models/CompanyApplication.js
```diff
@@ -33,6 +36,9 @@ export const CompanyApplicationRules = Object.freeze({
     CANNOT_REVIEW_TWICE: {
         msg: "company-application-already-reviewed",
     },
+    APPLICATION_RECENTLY_CREATED: {
+        msg: "company-application-created-recently",
```
DoStini:
```suggestion
        msg: "company-application-recently-created",
```

src/models/CompanyApplication.js
```diff
@@ -150,6 +161,13 @@ export const isRejectable = (application) => {
     return true;
 };
 
+
+CompanyApplicationSchema.methods.companyValidation = function() {
+    if (this.isVerified) throw new APIError(HTTPStatus.FORBIDDEN, ErrorTypes.FORBIDDEN, ValidationReasons.ALREADY_VALIDATED);
+    this.isVerified = true;
+    return this.save({ validateModifiedOnly: true });
+};
```
DoStini: I don't think we should be throwing an APIError at the model level of our platform. Following the other patterns, it should simply be an error, if I'm not mistaken

src/services/application.js
```diff
@@ -216,6 +217,36 @@ class CompanyApplicationService {
             throw new CompanyApplicationAlreadyReviewed(CompanyApplicationRules.CANNOT_REVIEW_TWICE.msg);
         }
     }
+
+    buildConfirmationLink(id) {
+        const token = generateToken({ _id: id }, config.jwt_secret, RECOVERY_LINK_EXPIRATION);
```
DoStini: Should not be recovery link but confirmation

src/models/constants/ApplicationStatus.js
```diff
@@ -1,7 +1,10 @@
 const ApplicationStatus = Object.freeze({
+    "UNVERIFIED": "UNVERIFIED",
     "PENDING": "PENDING",
     "APPROVED": "APPROVED",
     "REJECTED": "REJECTED",
 });
 
 export default ApplicationStatus;
+
+export const RECOVERY_LINK_EXPIRATION = 600;
```
DoStini: What is this being used for? Also, wrong naming as this is used for confirmation

DoStini: Should be `CONFIRMATION_LINK_EXPIRATION` or something like that

src/api/middleware/auth.js
```diff
@@ -78,15 +78,15 @@ export const hasAdminPrivileges = async (req, res, next) => {
...
-    });
+        storeInLocals(req, {
+            token: decoded,
+        });
 
-    return next();
+        return next();
+    } catch (err) {
+        return next(err);
+    }
```
DoStini: Why wrapped in try catch?

src/api/middleware/application.js
```diff
@@ -0,0 +1,19 @@
...
+    if (cursor !== null && Date.now() - cursor.submittedAt < RECOVERY_LINK_EXPIRATION * 1000) {
+        throw new APIError(HTTPStatus.FORBIDDEN, ErrorTypes.FORBIDDEN, CompanyApplicationRules.APPLICATION_RECENTLY_CREATED.msg);
+    }
+    return true;
+};
+
+export const deleteApplications = async (email) => {
+    await CompanyApplication.deleteMany({ email: email, isVerified: false }).catch(function(error) {
+        console.error(error);
+        throw (error); // Failure
```
DoStini:
```suggestion
        throw error;
```

src/api/middleware/application.js
```diff
@@ -0,0 +1,19 @@
+import CompanyApplication, { CompanyApplicationRules } from "../../models/CompanyApplication.js";
+import { APIError, ErrorTypes } from "./errorHandler.js";
+import { StatusCodes as HTTPStatus } from "http-status-codes/build/cjs/status-codes.js";
+import { RECOVERY_LINK_EXPIRATION } from "../../models/constants/ApplicationStatus.js";
+
+export const exceededCreationTimeLimit = async (email) => {
+    const cursor = await CompanyApplication.findOne({ email, isVerified: false }).exec();
```
DoStini:
```suggestion
    const application = await CompanyApplication.findOne({ email, isVerified: false }).exec();
```

src/api/middleware/application.js
```diff
@@ -0,0 +1,19 @@
...
+    if (cursor !== null && Date.now() - cursor.submittedAt < RECOVERY_LINK_EXPIRATION * 1000) {
```
DoStini: The logic here is quite confusing. I suggest reversing

```suggestion
    if (application !== null && Date.now() < application.submittedAt + RECOVERY_LINK_EXPIRATION * SECOND_IN_MS) {
```

DoStini: Isn't it possible to use this without the submittedAt field?

src/api/routes/application.js
```diff
@@ -12,14 +14,27 @@ export default (app) => {
...
+            const applicationService = new ApplicationService();
             // This is safe since the service is destructuring the passed object and the fields have been validated
-            const application = await (new ApplicationService()).create(req.body);
+            const application = await applicationService.create(req.body);
             return res.json(application);
         } catch (err) {
             console.error(err);
             return next(err);
         }
     });
```
DoStini: Validators are not called inside the route method, they are an argument, take a look at this example:

https://github.com/NIAEFEUP/nijobs-be/blob/2ddec2edbcc344c99c8c24592bc654e937e7ecc0/src/api/routes/offer.js#L115-L138

You might have to call as `(req) => applicationMiddleware.deleteApplications(req.body.email)`

DoStini (changes requested):

src/lib/token.js
FranciscoCardoso913: The ideia was to make it easier to understand what went wrong, otherwise it seems like the same

src/models/constants/ApplicationStatus.js
FranciscoCardoso913: Yes the naming is completely wrong( copy paste problem) it's used as a constant of the expiration time of the token

src/models/CompanyApplication.js
Naapperas: You're right: the way this method is implemented ties our data layer with the route layer, which can pose more work should we want to change either one of them.

src/api/middleware/application.js
Naapperas: Isn't this use of `exec` wrong? Also, `findOne` should, when awaited, return the specified object, not a cursor, if i recall correctly.

src/api/middleware/application.js
Naapperas: This should be wrapped in a `try/catch` instead of using the built-in `catch` method of promises.

src/api/middleware/application.js
Naapperas: Perhaps you could inverse the verification and check for errors or incorrect state, signaling it with a corresponding error. I don't know if I made myself clear.

src/api/routes/application.js
Naapperas: Also, why place the service in a variable that is only used once?

FranciscoCardoso913:
# New workflow

### Discussed solution

 - When creating a new application companies will need to verify their account by clicking on a link sent by email.
 - After that, the company will be able to finish its registration. In addition to that, an email will be sent to the NiJobs team to inform us that a new company wants to register so that we can evaluate its application
 - After finishing their registration, companies will be able to create offers, however, these offers will only be made public when the company's application is accepted (its status will be "pending")
 - When the company is accepted by us it will have full access to the NiJobs website.

src/api/middleware/auth.js
dsantosferreira: We were not aware that calling next with an error would skip all the handlers that can't handle such error. We are removing the try catch.

src/api/middleware/application.js
FranciscoCardoso913: I thought that when awaiting a promise you should use the mongoose exec because it returns a better stack trace.
https://mongoosejs.com/docs/promises.html

src/lib/token.js
Naapperas: What is your reasoning behind this request? @DoStini

src/api/middleware/application.js
Naapperas: @DoStini what do you mean by this?

src/services/company.js
```diff
@@ -189,6 +191,14 @@ class CompanyService {
...
     }
 
+    async releaseOffers(companyId) {
+        try {
+            await Offer.updateMany({ company : companyId }, { $set: { "isPending" : false } });
+        } catch (err) {
+            console.error(err);
+            throw err;
+        }
+    }
```
Naapperas: I get where you are coming from with the name of this method, but I think that there should be a better one. Perhaps not and I'm just being picky...

On another note, you have got the wrong attribute for the owning company: instead of `company` it should be `owner`

Naapperas (changes requested):

Naapperas:
@FranciscoCardoso913 @dsantosferreira Updates on this?

src/api/routes/company.js
```diff
@@ -258,23 +258,22 @@ export default (app) => {
     );
 
     /**
-     * Get the application of a company
+     * Get the state of a company with its id
      */
-    router.get("/:companyId/application",
+    router.get("/:companyId/state",
```
Naapperas: Why did this route change? Don't we need to see the full applications when logged-in as an admin?

test/end-to-end/company/:companyId/state.js
Naapperas, line 1: Good job on using the new test format, really helps a ton :)

Naapperas (changes requested):
Regarding the check annotations, GH does not have the ability to do an E2E security audit, from the looks of it. Most of the times the security issues that it signals are already being mitigated my validators/middleware. We should still see if these are valid just in case something slips through

src/api/routes/company.js
FranciscoCardoso913: We changed what was being returned by the route, instead of returning the whole application it is now only returning the state as we don't need the rest of the application's information.

Naapperas: Actually, it could be better to return the whole application: makes this specific endpoint future-proof. I'll leave it up to you whether this changes or not. Eitherway good job.

src/api/middleware/application.js
```diff
@@ -0,0 +1,13 @@
+import CompanyApplication, { CompanyApplicationRules } from "../../models/CompanyApplication.js";
+import { APIError, ErrorTypes } from "./errorHandler.js";
+import { StatusCodes as HTTPStatus } from "http-status-codes/build/cjs/status-codes.js";
+import { VALIDATION_LINK_EXPIRATION } from "../../models/constants/ApplicationStatus.js";
+import { SECOND_IN_MS } from "../../models/constants/TimeConstants.js";
+
+export const exceededCreationTimeLimit = async (req, res, next) => {
```
DoStini, line 7: I'm not sure about this function name, because it fails if an application exists, not when is exceeded. So I would call it something like `applicationDoesNotExist` or something similar. What do you think?

src/api/middleware/auth.js
```diff
@@ -78,15 +78,26 @@ export const hasAdminPrivileges = async (req, res, next) => {
...
+            token: decoded,
+        });
 
-    storeInLocals(req, {
-        token: decoded,
-    });
+        return next();
+    } catch (jwtErr) {
+        if (jwtErr.name === "TokenExpiredError") {
+            return next(new APIError(HTTPStatus.GONE, ErrorTypes.FORBIDDEN, ValidationReasons.EXPIRED_TOKEN));
```
DoStini: Why gone? This should just be forbidden, no? 

_"The HyperText Transfer Protocol (HTTP) 410 Gone client error response code indicates that access to the target resource is no longer available at the origin server and that this condition is likely to be permanent."_

src/api/routes/application.js
```diff
@@ -11,15 +14,36 @@ export default (app) => {
...
      */
-    router.post("/", validators.create, async (req, res, next) => {
-
+    router.post("/", validators.create, applicationMiddleware.exceededCreationTimeLimit, async (req, res, next) => {
         try {
+            const applicationService = new ApplicationService();
+            await applicationService.deleteApplications(req.body.email);
             // This is safe since the service is destructuring the passed object and the fields have been validated
-            const application = await (new ApplicationService()).create(req.body);
+            const application = await applicationService.create(req.body);
```
DoStini: We don't need to delete the application, just "update or create"
https://stackoverflow.com/a/72043808

src/email-templates/approval_notification.handlebars
```diff
@@ -1,9 +1,9 @@
-<h1>Glad to have you on board, {{companyName}}!</h1>
-<p>We decided to accept your application, and you are just one step away from being able to use NIJobs!</p>    
-<p>Now, you can log in to <a href="https://ni.fe.up.pt/nijobs">NIJobs</a> and complete your registration.
-Once you finish, you'll be able to place some work offers!</p>
+<h1>We are delighted to have you on board, {{companyName}}!</h1>
+<p>We are pleased to inform you that your application has been accepted, and all the offers
+    you have created so far are now visible to the public!</p>
+<p>Going forward, any offer you create will be immediately visible to the public audience.</p>
```
DoStini: We are pleased to inform you that your application has been accepted, and all the offers you have created are now accessible to the public!
From now on, any offer you create will be immediately available to the public audience.

src/email-templates/approval_notification.handlebars
```diff
@@ -1,9 +1,9 @@
...
 <br>
-<p>If you need anything else, don't hesitate to contact us!</p>
+<p>If you need anything else, don't hesitate to contact us at nijobs@aefeup.pt!</p>
```
DoStini: If you have any further questions

src/email-templates/confirm-application.handlebars
```diff
@@ -0,0 +1,7 @@
+<h1>Confirm your NIJobs application</h1>
+<p>We have received your application successfully!</p>
+<p>To complete the process, please click on the following <a href="{{link}}" target="_blank">link</a>. Kindly note that the link will expire in 10 minutes.</p>
```
DoStini: hyperlinks as actions are a good design practise

```suggestion
<p>To complete the process, please <a href="{{link}}" target="_blank">click on this link</a>. Kindly note that the link will expire in 10 minutes.</p>
```

src/email-templates/new_company_application_company.handlebars
```diff
@@ -1,8 +1,11 @@
-<h1>We have successfully received your application!</h1>
-<p>We will now review your application, and in case you're approved, you will receive another email with further instructions in order to complete your registration.</p>    
-<p>Your Application ID is {{applicationId}} and you registered {{companyName}}</p>
+<h1>Your application has been validated!</h1>
+<p>We will now review your application to determine the trustworthiness of your company.
+    In the meantime, you can complete your registration by logging into your account and begin creating offers.
+    Please note that your offers will remain hidden from the public until your company is approved.</p>
+<p>Your Application ID is {{applicationId}} and you registered {{companyName}}.</p>
```
DoStini: Don't say "trustworhiness", I think you can just say that we are verifying or reviewing the application.

src/lib/token.js
```diff
@@ -6,10 +6,4 @@ export const generateToken = (data, secret, expiresInSeconds) => jwt.sign(
...
 );
 
-export const verifyAndDecodeToken = (token, secret) => {
-    try {
-        return jwt.verify(token, secret, { algorithm: "HS256" });
-    } catch (err) {
-        return null;
-    }
-};
+export const verifyAndDecodeToken = (token, secret) => jwt.verify(token, secret, { algorithm: "HS256" });
```
DoStini, line 9: Was this updated in the recover password routes?

src/services/application.js
```diff
@@ -28,6 +33,17 @@ export class CompanyApplicationEmailAlreadyInUse extends Error {
...
+    constructor(msg) {
+        super(msg);
+    }
+}
+export class CompanyApplicationUnverified extends Error {
+    constructor(msg) {
+        super(msg);
+    }
+}
+
```
DoStini, line 46: It seems to me that these classes are not adding any additional information. Apparently, we were using it before but I don't see any reason to use them, as they are constructed in the same way as an Error. Are we veryfing the type of class somewhere above (to see if it's not unknown error or something)?

src/services/account.js
```diff
@@ -29,16 +29,11 @@ class AccountService {
 
         const company = await Company.create({ name: companyName });
 
-        const account = await Account.create({
+        await Account.create({
             email,
-            password,
+            password: (password),
```
DoStini: Why this change?

src/services/application.js
```diff
@@ -213,9 +211,47 @@ class CompanyApplicationService {
...
+            throw e;
         }
     }
+
+    buildConfirmationLink(id) {
+        const token = generateToken({ _id: id }, config.jwt_secret, VALIDATION_LINK_EXPIRATION);
+        return `${config.application_confirmation_link}${token}/validate`;
+    }
+
+    async applicationValidation(id) {
```
DoStini:
```suggestion
    async validateApplication(id) {
```

src/models/CompanyApplication.js
```diff
@@ -136,20 +152,35 @@ async function validateSingleActiveApplication(value) {
...
+
     if (application.state !== ApplicationStatus.PENDING)
-        throw new Error(CompanyApplicationRules.CANNOT_REVIEW_TWICE.msg);
+        throw new CompanyApplicationAlreadyReviewed(CompanyApplicationRules.CANNOT_REVIEW_TWICE.msg);
 
     return true;
 };
 
+
+CompanyApplicationSchema.methods.companyValidation = function() {
```
DoStini:
```suggestion
CompanyApplicationSchema.methods.verifyCompany= function() {
```

DoStini (changes requested):

src/api/middleware/application.js
FranciscoCardoso913: The function fails when the application exists and is exceeded.

src/lib/token.js
FranciscoCardoso913: yes

src/services/application.js
FranciscoCardoso913: Yes we use them to distinguish between errors

src/services/account.js
FranciscoCardoso913: It was an accident

src/api/middleware/auth.js
Naapperas: @FranciscoCardoso913 did you give an explanation regarding this?

src/api/routes/application.js
Naapperas, line 1: I just thought of something: although the token was generated as part of the process of a company applying to use NIJobs, the token itself is not part of the company. As such, I would suggest restructuring the code a bit, following the comments I'll be leaving next.

The general idea is that a token is part of an application but not a company, so having `/apply/company/:token` seemed a bit odd to me. What do you think?

Naapperas: Feel free to ignore the following comments if you don't agree with this opinion, given that you reason about it.

src/api/routes/application.js
```diff
@@ -11,15 +14,36 @@ export default (app) => {
     /**
      * Creates a new Company Application
      */
-    router.post("/", validators.create, async (req, res, next) => {
-
+    router.post("/", validators.create, applicationMiddleware.exceededCreationTimeLimit, async (req, res, next) => {
```
Naapperas, line 17:
```suggestion
    router.post("/company", validators.create, applicationMiddleware.exceededCreationTimeLimit, async (req, res, next) => {
```

I cannot edit unmodified lines but above, on line 12, it should become:
```js
app.use("/apply", router);
```

This change should _not_ break existing functionality/tests.

src/api/routes/application.js
```diff
@@ -11,15 +14,36 @@ export default (app) => {
...
         } catch (err) {
             console.error(err);
             return next(err);
         }
     });
+
+    /**
+     * Validates application
+     */
+    router.post("/:token/validate", validators.finishValidation, validToken, async (req, res, next) => {
```
Naapperas, line 31: All tests that hit this route should take into account the change made on the comment above.

src/api/routes/offer.js
```diff
@@ -98,14 +101,14 @@ export default (app) => {
         validators.offersDateSanitizers,
         async (req, res, next) => {
             try {
-
+                const account = await Account.findOne({ company: req.targetOwner });
+                const application = await CompanyApplication.findOne({ email: account.email });
```
Naapperas, line 105: On way for this to "look cleaner" is if the "pending" state could be computed from the account object itself. Take a look at [Mongoose Schema Virtuals](https://mongoosejs.com/docs/tutorials/virtuals.html) and see if you agree with this.

Naapperas: Also, maybe this could be hidden behind a service call? Don't we have an `AccountService` ?

src/api/routes/review.js
```diff
@@ -82,7 +83,9 @@ export default (app) => {
         async (req, res, next) => {
 
             try {
-                const { account } = await (new ApplicationService()).approve(req.params.id);
+                const account = await (new ApplicationService()).approve(req.params.id);
+                const company = await Company.findOne({ _id: account.company });
```
Naapperas, line 87: Same thing as last comment.

src/email-templates/approval_notification.handlebars
Naapperas: I believe that, although companies should already know this, we should warn them about our imposed limits on concurrent offers.

src/email-templates/new_company_application_company.handlebars
Naapperas: @FranciscoCardoso913 any comments on this?

src/lib/token.js
Naapperas, line 9: Why did you move the error handling logic out of this method? I recall you returning different errors based on the thrown exception. Couldn't we abstract those `jwt` exceptions into our custom, more semantic, ones? Food for thought.

src/services/account.js
Naapperas: @FranciscoCardoso913 this does not seem to have changed, was it a typo or not?

Naapperas (changes requested):
Most of the comments I made serve as introductions for discussion, not necessarily requested changes. There are some, however, hence this.

Besides this, I just remembered something: what is the behavior of offers belonging to a company that gains full access to NIJobs if there are more than the allowed amount of concurrent offers pending? (I might have forgotten, sorry :smiling_face_with_tear:)

Naapperas:
Some of the comments appear outdated and I don't know why: when I open the respective files nothing changed. Please resolve or answer the conversations as if they were not outdated.

src/lib/token.js
DoStini: I agree with this comment, I think the error handling error should be inside and outside a simpler null check

DoStini (approved):
As of my comments, everything seems to be fine. Before merging address the comments @Naapperas suggested

PR #308, opened 2023-02-14
Feature/edit company details

### LINKED ISSUE #58: Edit Company details
A Company Account should be able to edit its Company's details.

marco-vb:


marco-vb:
Please review the tests to see if they are good and complete or if I need to add more test cases.

src/api/middleware/company.js
```diff
@@ -110,15 +110,28 @@ export const isNotDisabled = (owner) => async (req, res, next) => {
...
+        const company = await (new CompanyService()).findById(companyId, true);
+
+        // company not found
+        if (!company) {
+            return next(new APIError(
+                HTTPStatus.NOT_FOUND,
+                ErrorTypes.VALIDATION_ERROR,
+                ValidationReasons.COMPANY_NOT_FOUND
+            ));
+        }
```
BrunoRosendo: Why do we add this in `canManageAccountSettings`? Other endpoints use the validator for this (might change in #309 ) so we should be consistent

src/api/middleware/validators/company.js
```diff
@@ -45,6 +47,10 @@ export const companyExists = async (companyId) => {
     return true;
 };
 
+export const companyEnabled = (companyId, req, res, next) => companyMiddleware.isNotDisabled(companyId)(req, res, next);
+
+export const companyNotBlocked = (companyId, req, res, next) => companyMiddleware.isNotBlocked(companyId)(req, res, next);
+
```
BrunoRosendo: Where is this being used? The route is calling the middleware directly

src/api/middleware/validators/company.js
```diff
@@ -71,6 +77,12 @@ export const enable = useExpressValidators([
     existingCompanyParamValidator,
 ]);
 
+export const canBeEditedBy = (req, res, next) => {
+    if (!req.user?.isAdmin && req.user?.companyId !== req.params.companyId)
+        return next(new APIError(HTTPStatus.FORBIDDEN, ErrorTypes.FORBIDDEN, ValidationReasons.UNAUTHORIZED));
+    return next();
+};
```
BrunoRosendo: Isn't this already covered by the middleware?

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,195 @@ describe("Company endpoint", () => {
...
+                    .expect(HTTPStatus.OK);
+
+                const res = await test_agent
+                    .put(`/company/${test_company._id}/edit`)
+                    .send({
+                        name: "Changing Company",
+                        bio: "Without permission",
+                        contacts: ["1"],
+                        logo: "http://awebsite.com/otherlogo.jpg",
+                    })
```
BrunoRosendo: In the same way we have for offers, a function to generate a company with default values might be useful. Just a suggestion

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,195 @@ describe("Company endpoint", () => {
...
+                    .put(`/company/${id}/edit`)
+                    .send()
+                    .expect(HTTPStatus.UNAUTHORIZED);
+
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.INSUFFICIENT_PERMISSIONS);
+            });
+        });
+
+        describe("Should fail if different user/company", () => {
+            test("Should fail if different user", async () => {
```
BrunoRosendo: Writing a describe with only one test with the same descriptions seems, redundant, I'd remove describe in cases like this one. Alternatively, wrap this test and the ones below into one describe

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,195 @@ describe("Company endpoint", () => {
...
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.INSUFFICIENT_PERMISSIONS_COMPANY_SETTINGS);
+            });
+        });
+
+        describe("Should pass if god", () => {
+            test("Should pass if god", async () => {
+                await test_agent
+                    .post("/auth/login")
+                    .send(test_user_admin)
+                    .expect(HTTPStatus.OK);
```
BrunoRosendo, line 2100: God doesn't need the login

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,195 @@ describe("Company endpoint", () => {
             expect(res.body).toHaveProperty("maxOffersReached", true);
         });
     });
+
+    describe("PUT /company/edit", () => {
```
BrunoRosendo, line 1946: Suggestions for new tests:
- No login or god token (fail)
- Blocked and disabled company (from admin and company perspective)
- Check if the company's offers update ownerName and ownerLogo. Since we don't use a relationship DB, data integration isn't as straight forward

BrunoRosendo (changes requested):
Good job so far! I left some suggestions

src/api/middleware/company.js
marco-vb: There was indeed no need for this, I removed it.

src/api/middleware/validators/company.js
marco-vb: This was also not being used, also removed.

src/api/middleware/validators/company.js
marco-vb: Also removed!

test/end-to-end/company.js
marco-vb: I did it like this because I don't always want to pass all parameters in the body (maybe the company just wants to edit contacts, for example).

test/end-to-end/company.js
marco-vb: Fixed this!

test/end-to-end/company.js
marco-vb: Added all of these, thank you!

src/api/middleware/company.js
```diff
@@ -110,15 +110,19 @@ export const isNotDisabled = (owner) => async (req, res, next) => {
...
-    const company = await (new CompanyService()).findById(companyId, true);
-
-    // only god or the same company can change account settings
-    if (!req.hasAdminPrivileges && company._id.toString() !== req.targetOwner) {
-        return next(new APIError(
-            HTTPStatus.FORBIDDEN,
-            ErrorTypes.FORBIDDEN,
-            ValidationReasons.INSUFFICIENT_PERMISSIONS_COMPANY_SETTINGS
-        ));
+    try {
```
BrunoRosendo, line 113: Why is this try catch only needed now?

src/services/offer.js
```diff
@@ -443,6 +443,15 @@ class OfferService {
         await Offer.deleteMany({ owner: companyId });
     }
 
+    async updateAllOffersByCompanyId(id, company) {
```
BrunoRosendo: Not a big deal, but `company` already has an `id` property, right? SO the `id` param is probably redundant. It's not really a problem, just a suggestion

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,257 @@ describe("Company endpoint", () => {
...
+                const res = await test_agent
+                    .put(`/company/${id}/edit`)
+                    .send()
+                    .expect(HTTPStatus.UNAUTHORIZED);
+
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.INSUFFICIENT_PERMISSIONS);
+            });
+        });
+
+        describe("Should fail if bad user", () => {
```
BrunoRosendo: Also a small detail, but we usually use "should" for the tests. I think a description like "Using a bad user" or "Using a good user" would make more sense

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,257 @@ describe("Company endpoint", () => {
...
+            });
+
+            test("Should fail if id is not a valid company", async () => {
+                const id = "111111111111111111111111";
+                const res = await test_agent
+                    .put(`/company/${id}/edit`)
+                    .send()
+                    .expect(HTTPStatus.UNAUTHORIZED);
+
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.INSUFFICIENT_PERMISSIONS);
```
BrunoRosendo: Should you be checking for unauthorized? I think this should be a bad request, right? Probably missing a login

src/api/middleware/company.js
```diff
@@ -110,15 +110,19 @@ export const isNotDisabled = (owner) => async (req, res, next) => {
...
+        const company = await (new CompanyService()).findById(companyId, true);
+        // only god or the same company can change account settings
```
Naapperas: Remove this comment since it no longer makes sense.

BrunoRosendo (changes requested):
Looks good, I left some suggestions. It'd probably be nice to get the company from db/service after editing and checking if everything is ok, at least in one test

src/services/offer.js
```diff
@@ -443,6 +443,15 @@ class OfferService {
...
+        const offer = {
+            owner: id,
```
Naapperas: I don't think this field is necessary, but correct me if I'm wrong

src/services/company.js
```diff
@@ -189,6 +190,26 @@ class CompanyService {
...
+                companyDetails,
+                { new: true }
+            );
+
+            return company;
+        } catch (err) {
+            console.error(err);
+            throw err;
+        }
+    }
```
Naapperas: We have a `changeAttributes` function so this behavior is duplicated: either use that or change the function's name.

Naapperas (changes requested):

src/api/middleware/company.js
BrunoRosendo: @marco-vb This was resolved but you didn't reply

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,294 @@ describe("Company endpoint", () => {
...
+        };
+
+        const test_random_user = {
+            email: "random@email.com",
+            password: "password123",
+        };
+
+        const test_agent = agent();
+
+        beforeEach(async () => {
```
Naapperas: Shouldn't this be a `beforeAll` ?

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,294 @@ describe("Company endpoint", () => {
...
+            await test_agent
+                .delete("/auth/login")
+                .expect(HTTPStatus.OK);
```
Naapperas: If you already have the logout call later this is redundant

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,294 @@ describe("Company endpoint", () => {
...
+        });
+
+        afterAll(async () => {
+            await Company.deleteMany({});
+            await Account.deleteMany({});
+        });
+
+        describe("ID Validation", () => {
```
Naapperas, line 2018: Since this is its own describe you could move the login call to a `beforeEach`: prevents code duplication and looks cleaner

Naapperas: And add an `afterEach` with a logout action

test/end-to-end/company.js
Naapperas: This explanation makes sense but in that sense maybe you could write 1 or 2 tests in which you change only a specific attribute, or assert that functionality multiple times in the same test.

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,294 @@ describe("Company endpoint", () => {
...
+                    .put(`/company/${test_company._id}/edit`)
+                    .send({
+                        name: "Changing Company",
+                        bio: "Without login",
+                        contacts: ["1"],
+                        logo: "http://awebsite.com/otherlogo.jpg",
+                    })
+                    .expect(HTTPStatus.UNAUTHORIZED);
+
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.INSUFFICIENT_PERMISSIONS);
```
Naapperas: We normally use a `toContainEqual`: just a suggestion.

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,294 @@ describe("Company endpoint", () => {
...
+            test("Should pass if god", async () => {
+                await test_agent
+                    .put(`/company/${test_company._id}/edit`)
+                    .send(withGodToken({
+                        name: "Changing Company",
+                        bio: "As god",
+                        contacts: ["1"],
+                        logo: "http://awebsite.com/otherlogo.jpg",
+                    }))
+                    .expect(HTTPStatus.OK);
```
Naapperas: Perhaps you could assert that the resulting company has the changed properties

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,294 @@ describe("Company endpoint", () => {
...
+                        bio: "As company itself",
+                        logo: "http://awebsite.com/otherlogo.jpg",
+                    })
+                    .expect(HTTPStatus.OK);
+
+                expect(res.body).toHaveProperty("name", "Changing Company");
+                expect(res.body).toHaveProperty("bio", "As company itself");
+                expect(res.body).toHaveProperty("logo", "http://awebsite.com/otherlogo.jpg");
+            });
+        });
```
Naapperas, line 2129: Missing tests for admins

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,294 @@ describe("Company endpoint", () => {
...
+            await test_agent
+                .post("/auth/login")
+                .send(test_user_company)
+                .expect(HTTPStatus.OK);
+
+            const res = await test_agent
+                .put(`/company/${test_company._id}/edit`)
+                .send({
+                    name: "Changing Company",
+                    contacts: ["122"],
```
Naapperas: Just a suggestion, but you could maybe factor these values out into another variable, since they are reused later in the test.

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,294 @@ describe("Company endpoint", () => {
...
+        });
+
+        describe("Should fail if company is blocked or disabled", () => {
+            test("Should fail if company is blocked (admin)", async () => {
+                await test_agent
+                    .post("/auth/login")
+                    .send(test_user_admin)
+                    .expect(HTTPStatus.OK);
+
+                await Company.findByIdAndUpdate(test_company._id, { isBlocked: true });
```
Naapperas: This is fine but it would be best, or at least consistent with other tests, if you created a separate company that starts out as blocked.

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,294 @@ describe("Company endpoint", () => {
...
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.COMPANY_BLOCKED);
+            });
+
+            test("Should fail if company is disabled (admin)", async () => {
+                await test_agent
+                    .post("/auth/login")
+                    .send(test_user_admin)
+                    .expect(HTTPStatus.OK);
+
+                await Company.findByIdAndUpdate(test_company._id, { isDisabled: true });
```
Naapperas: Same thing as above

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,294 @@ describe("Company endpoint", () => {
...
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.COMPANY_DISABLED);
+            });
+
+            test("Should fail if company is blocked (user)", async () => {
+                await test_agent
+                    .post("/auth/login")
+                    .send(test_user_company)
+                    .expect(HTTPStatus.OK);
+
+                await Company.findByIdAndUpdate(test_company._id, { isBlocked: true });
```
Naapperas: Same thing as above, since this is supposed to fail you can reuse the same company used in the other test: it is guaranteed to not change so there is no problem.

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,294 @@ describe("Company endpoint", () => {
...
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.COMPANY_BLOCKED);
+            });
+
+            test("Should fail if company is disabled (user)", async () => {
+                await test_agent
+                    .post("/auth/login")
+                    .send(test_user_company)
+                    .expect(HTTPStatus.OK);
+
+                await Company.findByIdAndUpdate(test_company._id, { isDisabled: true });
```
Naapperas: Same thing as above.

Naapperas (changes requested):

test/end-to-end/company.js
marco-vb: Ok, I will now only change a few attributes for each test.

src/api/middleware/company.js
marco-vb: Sorry forgor to reply, I wrapped this because of eventual errors in the database access, I don't know if it's necessary.

BrunoRosendo: ok sure

BrunoRosendo (dismissed):
lgtm, maybe @Naapperas wants to re-review

src/services/offer.js
```diff
@@ -468,6 +468,15 @@ class OfferService {
         await Offer.deleteMany({ owner: companyId });
     }
 
+    async updateAllOffersByCompanyId(company) {
+        const offer = {
+            ownerName: company.name,
+            ownerLogo: company.logo,
+            contacts: company.contacts,
+        };
+        const offers = await Offer.updateMany({ owner: company._id }, offer, { new: true });
```
Naapperas:
```suggestion
        const changes = {
            ownerName: company.name,
            ownerLogo: company.logo,
            contacts: company.contacts,
        };
        const offers = await Offer.updateMany({ owner: company._id }, changes, { new: true });
```

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,330 @@ describe("Company endpoint", () => {
...
+                isDisabled: true
+            });
+
+            test_random = await Company.create({
+                name: test_random_company_map.name,
+                contacts: test_random_company_map.contacts,
+                bio: test_random_company_map.bio,
+                logo: test_random_company_map.logo,
+                hasFinishedRegistration: true
+            });
```
Naapperas: Several things here:

- This is just my personal preference, but mongoose models support operations on arrays of objects, i.e., instead of creating a company on the DB and retrieving that company, you could pass in an array of company data and the return value of the function would be an array with the respective models. This can help avoid multiple calls to the same model method and makes the code look cleaner. Example:
```js
const [company1, company2] = await Company.create([company1Data, company2Data]);
```

- You should make use of the [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) to avoid rewriting the same object structure over and over. Another option is to write a helper method that returns a "default" company but could be overridden by the input arguments.

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,330 @@ describe("Company endpoint", () => {
...
+
+            const test_offer_data = generateTestOffer({
+                owner: test_company._id,
+                ownerName: test_company.name,
+                ownerLogo: test_company.logo,
+            });
+
+            test_offer = await Offer.create(test_offer_data);
```
Naapperas: Not important, but you could just put the `generateTestOffer` call directly in the model method, since the variable is not used elsewhere.

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,330 @@ describe("Company endpoint", () => {
...
+                company: test_disabled_company._id
+            });
+
+            await Account.create({
+                email: test_random_user.email,
+                password: await hash(test_random_user.password),
+                company: test_random._id
+            });
+        });
+
```
Naapperas: Same thing as above regarding batching multiple model creations

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,330 @@ describe("Company endpoint", () => {
...
+                    .put(`/company/${test_company._id}/edit`)
+                    .send({
+                        bio: "Without login",
+                        contacts: ["1"],
+                    })
+                    .expect(HTTPStatus.UNAUTHORIZED);
+
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.INSUFFICIENT_PERMISSIONS);
+                expect(test_company.bio).toBe(test_company_map.bio);
+                expect(test_company.contacts).toEqual(test_company_map.contacts);
```
Naapperas: What are you testing here?

Naapperas: I believe you wanted to check if the company hadn't changed but this is not the way to do so.

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,330 @@ describe("Company endpoint", () => {
...
+                    }))
+                    .expect(HTTPStatus.OK);
+
+                expect(res.body).toHaveProperty("name", "Changing Company");
+                expect(res.body).toHaveProperty("logo", "http://awebsite.com/otherlogo.jpg");
+
+                const changed_company = await Company.findById(test_company._id);
+
+                expect(changed_company.name).toBe("Changing Company");
+                expect(changed_company.logo).toBe("http://awebsite.com/otherlogo.jpg");
```
Naapperas: You do not need to query the DB again, the response body is the edited company.

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,330 @@ describe("Company endpoint", () => {
...
+                const res = await test_agent
+                    .put(`/company/${test_company._id}/edit`)
+                    .send({
+                        name: "Changing Company",
+                    })
+                    .expect(HTTPStatus.OK);
+
+                expect(res.body).toHaveProperty("name", "Changing Company");
+                const changed_company = await Company.findById(test_company._id);
+                expect(changed_company.name).toBe("Changing Company");
```
Naapperas: Same thing as above

test/end-to-end/company.js
Naapperas: @marco-vb please take a look at this

test/end-to-end/company.js
Naapperas: @marco-vb please look at this

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,330 @@ describe("Company endpoint", () => {
...
+        describe("Using disabled/blocked company (god)", () => {
+            test("Should fail if company is blocked (god)", async () => {
+                const res = await test_agent
+                    .put(`/company/${test_blocked_company._id}/edit`)
+                    .send(withGodToken({
+                        name: "Changing Blocked Company",
+                    }))
+                    .expect(HTTPStatus.FORBIDDEN);
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.COMPANY_BLOCKED);
+                expect(test_blocked_company.name).toBe(test_blocked_company_map.name);
```
Naapperas: Same thing

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,330 @@ describe("Company endpoint", () => {
...
+
+            test("Should fail if company is disabled (god)", async () => {
+                const res = await test_agent
+                    .put(`/company/${test_disabled_company._id}/edit`)
+                    .send(withGodToken({
+                        name: "Changing Disabled Company",
+                    }))
+                    .expect(HTTPStatus.FORBIDDEN);
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.COMPANY_DISABLED);
+                expect(test_disabled_company.name).toBe(test_disabled_company_map.name);
```
Naapperas: Same thing

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,330 @@ describe("Company endpoint", () => {
...
+
+                const res = await test_agent
+                    .put(`/company/${test_blocked_company._id}/edit`)
+                    .send({
+                        name: "Changing Blocked Company",
+                    })
+                    .expect(HTTPStatus.FORBIDDEN);
+
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.COMPANY_BLOCKED);
+                expect(test_blocked_company.name).toBe(test_blocked_company_map.name);
```
Naapperas: Same thing

test/end-to-end/company.js
```diff
@@ -1933,4 +1933,330 @@ describe("Company endpoint", () => {
...
+                    .expect(HTTPStatus.OK);
+
+                const res = await test_agent
+                    .put(`/company/${test_disabled_company._id}/edit`)
+                    .send({
+                        bio: "As user",
+                    })
+                    .expect(HTTPStatus.FORBIDDEN);
+                expect(res.body.errors[0]).toHaveProperty("msg", ValidationReasons.COMPANY_DISABLED);
+                expect(test_disabled_company.bio).toBe(test_disabled_company_map.bio);
```
Naapperas: Same thing

Naapperas (changes requested):

test/end-to-end/company.js
marco-vb: Fixed this!

test/end-to-end/company.js
marco-vb: That's right. I will remove this because since we have a error it is redundant to check if the company did not change.

test/end-to-end/company.js
marco-vb: Removed

test/end-to-end/company.js
marco-vb: Removed

test/end-to-end/company.js
marco-vb: Removed

test/end-to-end/company.js
marco-vb: Removed

src/services/offer.js
marco-vb: Done!

test/end-to-end/company.js
marco-vb: Fixed!

test/end-to-end/company.js
marco-vb: Fixed

test/end-to-end/company.js
marco-vb: Did a map of 'changing_values' and used that.

test/end-to-end/company.js
marco-vb: fixed this! sorry I missed it the first time

test/end-to-end/company.js
marco-vb: sure!

test/end-to-end/company.js
marco-vb: refactored this, see if it's ok

test/end-to-end/company.js
marco-vb: also refactored this

Naapperas (approved):

PR #309, opened 2023-02-15
Cleanup middlewares/validators

### LINKED ISSUE #282: Inconsistent errors when model is not found
As discussed in [this comment](https://github.com/NIAEFEUP/nijobs-be/pull/207#discussion_r815173860) and [this other comment](https://github.com/NIAEFEUP/nijobs-be/pull/207#discussion_r1029678637), when a model is not found a 404 Not Found error should be returned, however, this only happens for a few routes as we're currently using validators to check if a model exists, which return a 422 Unprocessable Entity instead.

Additionally, when a route does return a 404 this is implemented in the function body directly, leading to duplicated code. This could be abstracted out into a middleware.

ttoino:
Closes #282 and fixes other inconsistencies in the use/implementation of middlewares and validators.

### DIFF
--- a/src/api/middleware/company.js
+++ b/src/api/middleware/company.js
@@ -6,6 +6,7 @@ import CompanyConstants from "../../models/constants/Company.js";
 import Offer from "../../models/Offer.js";
 import CompanyService from "../../services/company.js";
 import OfferService from "../../services/offer.js";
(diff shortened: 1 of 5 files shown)


Naapperas:
@ttoino any updates on this?

PR #311, opened 2023-03-01
Refactor tests

### LINKED ISSUE #169: "Normalize" tests

Naapperas:
Closes #169 

This attempts to fix and "normalize" the many tests we have on NIJobs in order to make them more manageable.


Naapperas:
After refactoring work on this PR is finished, I am going to write documentation regarding how contributors should write tests.

Naapperas:
Looks like we will still need to wait for a new version of `fast-xml-parser`

PR #338, opened 2023-11-09
234 multiple locations in job offer

marco-vb:
Implemented the necessary changes to support multiple locations in a job offer instead of a single one.

### DIFF
--- a/src/models/Offer.js
+++ b/src/models/Offer.js
@@ -107,7 +107,7 @@ const OfferSchema = new Schema({
         type: [String],
         required: true,
         validate: [
-            (val) => val.length >= OfferConstants.locations.min_length && val.length <= OfferConstants.locations.max_length,
+            (val) => lengthBetweenValidator(val, OfferConstants.locations.min_length, OfferConstants.locations.max_length),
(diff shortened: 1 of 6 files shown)


src/models/Offer.js
```diff
@@ -103,7 +103,14 @@ const OfferSchema = new Schema({
...
     ownerLogo: { type: String, required: true, validate: (val) => validImageURL(val) },
-    location: { type: String, required: true },
+    location: {
+        type: [String],
+        required: true,
+        validate: [
+            (val) => val.length >= OfferConstants.locations.min_length && val.length <= OfferConstants.locations.max_length,
+            `There must be between ${OfferConstants.locations.min_length} and ${OfferConstants.locations.max_length} locations`,
+        ],
+    },
```
FranciscoCardoso913: Maybe pass this to a separate function like the validImageUrl

FranciscoCardoso913 (changes requested):
Seems good, I just have a small request

src/models/Offer.js
marco-vb: Used function `lengthBetweenValidator` that already existed.

PR #340, opened 2023-11-29
Added init script for database models

Naapperas:
This PR adds a script that takes a MongoDB dump and populates the local database with that data.

### DIFF
--- a/seed.sh
+++ b/seed.sh
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+
+mongorestore --uri mongodb://localhost:27017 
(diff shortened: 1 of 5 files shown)


seed.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/env bash
```
dsantosferreira: Had to replace the shebang to `#!/usr/bin/env bash` in macOS. Don't know if that should be considered an issue or not

dsantosferreira (dismissed):
Looks good to me! Left you with one small question

seed.sh
Naapperas: MacOS moment, will change that.

dsantosferreira (approved):
Looks good to me!

PR #341, opened 2023-11-29
299 custom reason for application rejection

### LINKED ISSUE #299: Custom reason for application rejection
Admins should be able to provide a custom reason for a company application to be rejected, instead of the default message that is in the email template.

odracir04:
the reason in the rejection email is now allowed to be specified by admins in the post request.

### DIFF
--- a/src/email-templates/rejection_notification.handlebars
+++ b/src/email-templates/rejection_notification.handlebars
@@ -1,8 +1,7 @@
 <p>Hello {{companyName}},</p>
 <br>  
-<p>We have reviewed your application for {{companyName}} and decided to reject your application.</p>    
-<p>Some common causes of rejection are spam, abuse or a motivation that does not align with our principles. 
-    If you believe this was a mistake, please get back to us at <a href="mailto:nijobs@aefeup.pt">nijobs@aefeup.pt</a></p>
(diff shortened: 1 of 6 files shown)


odracir04:
it's worth mentioning that we had to alter the tests to include the new rejection messages but the tests cover the new code

src/email-templates/rejection_notification.handlebars
```diff
@@ -1,8 +1,7 @@
 <p>Hello {{companyName}},</p>
 <br>  
-<p>We have reviewed your application for {{companyName}} and decided to reject your application.</p>    
-<p>Some common causes of rejection are spam, abuse or a motivation that does not align with our principles.
```
Naapperas, line 4: I don't think this should be removed, it provides more context as to why some applications might be removed in the future. Thoughts? @odracir04

Naapperas (changes requested):

src/email-templates/rejection_notification.handlebars
odracir04: I agree. It does make it clearer to the applicant. Will place it back

Naapperas:
@odracir04 please rebase and update the `package-lock.json` file so this PR can be reviewed.

PR #345, opened 2024-01-31
Bump mongoose version to accomodate MongoDB Atlas changes

Naapperas:
MongoDB announced that Atlas would be migrating all database instances to version 7 of MongoDB.

This PR updates `mongoose` and the code accordingly to reflect the changes made (since this is potentially bumping 2 major versions).

There is no need to change the `docker-compose.yaml` file since it pulls from the untagged `mongo` image, aka, the latest image available on Docker Hub. I already checked and, on my machine, the image, when running `mongosh`, runs server version 7.x, which is what we want.


Naapperas:
Audit fails due to [this](https://github.com/advisories/GHSA-9h6g-pr28-7cqp), which was only reported 11 hours ago, will need to wait for a fix :skull:

PR #347, opened 2024-02-13
346 add independent jobs

### LINKED ISSUE #346: Add independent jobs
In addition to traditional job offerings that require employment within a company, we should have offers for independent jobs that do not require employment, such as making a website for a company, making a database, etc...

jose-carlos-sousa:
I added the new type of jobs (Freelance) to the JobTypes I also made some params optional  for instance the min and max duration might not make a lot of sense for these types of jobs that are goal oriented

### DIFF
--- a/src/models/Offer.js
+++ b/src/models/Offer.js
@@ -32,15 +32,20 @@ const OfferSchema = new Schema({
     },
 
     jobMinDuration: {
-        type: Number
+        type: Number,
+        required: validateMinDuration,
     },
+
     jobMaxDuration: {
+        required: true,
         type: Number,
         validate: [
             validateJobMaxDuration,
             "`jobMaxDuration` must be larger than `jobMinDuration`",
+
         ],
     },
+
     jobStartDate: { type: Date },
     description: {
         type: String,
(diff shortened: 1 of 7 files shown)


src/models/Offer.js
```diff
@@ -135,9 +138,15 @@ export function validatePublishEndDateLimit(publishDate, publishEndDate) {
 
 // jobMaxDuration must be larger than jobMinDuration
 function validateJobMaxDuration(value) {
+    if (this.jobType === "FREELANCE" && this.jobMinDuration === undefined) return true;
     return value >= this.jobMinDuration;
```
dsantosferreira: This condition isn't consistent with the values used in the tests for the jobMinDuration field, as here the value "undefined" is used and "null" is used in the tests. Either change the value "undefined" to "null" or vice versa in the tests. I would recommend using "null". The tests only pass because of numeric type coersion (null is cast to 0 when using ">=").

dsantosferreira (changes requested):
Everything you did seems really good. Left a small recommendation and I'm left with a doubt. To my understanding of the issue, shouldn't the backend raise an error when freelance offers are defined with a non-null jobMinDuration? I am open to discuss that.

src/models/Offer.js
jose-carlos-sousa: Thanks for pointing that out, I followed your suggestion and changed "undefined" to "null"

jose-carlos-sousa: As you suggested and as was discussed in the meeting I decided to make it impossible for freelance offers to have a min duration

PR #349, opened 2024-03-13
Refactor Company Application Status

### LINKED ISSUE #312: Refactor Offer State
We should not have the offers' state be defined by a arbitrary amount of booleans: instead we should create a class/enum for this. After #306 Offers will have a pending pseudo-state, so that should be taken into account.

Minipoloalex:
Replaced the virtual state in company application with an enum.

Previously, the virtual state depended on whether the `rejectedAt` and `approvedAt` fields were defined. Now it is defined with an explicit enum instead of being a virtual field.

I also had to define a default function for the application state enum depending on those same fields. With a simple default `PENDING`, migrating the database would cause approved/rejected applications to be seen as pending. This function makes the default value depend on `approvedAt` and `rejectedAt`.

Also replaced the tests related with the virtual state.

Closes #312

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5882,9 +5882,9 @@
       "license": "ISC"
     },
     "node_modules/ip": {
-      "version": "2.0.0",
-      "resolved": "https://registry.npmjs.org/ip/-/ip-2.0.0.tgz",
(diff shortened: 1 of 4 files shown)


jose-carlos-sousa (approved):
Seems to work as expected :)

